PR #8, opened 2015-07-06
Added depend by tarantool python library

zloidemon:


Sulverus:
Tarantool-pyhton used as submodule in test-run

PR #45, opened 2017-03-03
Rename deprecated gevent.coros with gevent.lock

anatol:
Rename deprecated gevent.coros with gevent.lock

gevent.coros is deprecated and removed in the latest versions http://www.gevent.org/whatsnew_1_0.html

### DIFF
--- a/lib/connpool.py
+++ b/lib/connpool.py
@@ -1,7 +1,7 @@
 import logging
 
 import gevent
-from gevent.coros import BoundedSemaphore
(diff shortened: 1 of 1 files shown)


Totktonada:
This can be done in non-breaking (for those who using old gevent) manner:

```
-from gevent.coros import BoundedSemaphore
+try:
+    from gevent.lock import BoundedSemaphore
+except:
+    from gevent.coros import BoundedSemaphore  # before gevent-1.0
```

PR #64, opened 2017-10-16
Create function for environment injection into Tarantool servers

opomuc:
Required for integration tests

### DIFF
--- a/test-run.py
+++ b/test-run.py
@@ -100,7 +100,7 @@ def main_loop_parallel():
     lib.worker.print_greetings()
 
     color_stdout("\n", '=' * 86, "\n", schema='separator')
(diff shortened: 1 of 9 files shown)


test-run.py
```diff
@@ -100,7 +100,7 @@ def main_loop_parallel():
     lib.worker.print_greetings()
 
     color_stdout("\n", '=' * 86, "\n", schema='separator')
-    color_stdout("WORKR".ljust(6),     schema='t_name')
+    color_stdout("WORKER".ljust(7),    schema='t_name')
```
Totktonada: It was written abbreviated intentionally to look as a table header for output below. But to be honest, I don’t remember before or after a worker id becomes three-digit long.

Issue #65, opened 2017-10-17
test-run fails to kill cluster servers in case of an error


opomuc:
Say we have a Lua test like this:
```
#!/usr/bin/env tarantool
env = require('test_run')
test_run = env.new()

servers = {'stub1', 'stub2'}
test_run:create_cluster(servers)

local index_table = {}
local error = index_table.nil_value.error -- index nil field

test_run:drop_cluster(servers)
os.exit(0)
```
When the test crushes, test-run does not kill started in `create_cluster` servers

Totktonada:
I'll consider this issue as statement of the problem with 'core = app' tests.

PR #72, opened 2017-11-14
Kill hanging servers in case of errors

opomuc:
Solves issue #65.
Prints error messages from servers + last lines of tarantool log in case of an error

### DIFF
--- a/lib/preprocessor.py
+++ b/lib/preprocessor.py
@@ -152,13 +152,7 @@ class TestState(object):
             schema='test_var')
         if sname not in self.servers:
             raise LuaPreprocessorException('Can\'t start nonexistent server '+repr(sname))
-        wait = True
-        if 'wait' in opts and opts['wait'] == 'False':
-            wait = False
-        wait_load = True
-        if 'wait_load' in opts and opts['wait_load'] == 'False':
-            wait_load = False
-        self.servers[sname].start(silent=True, rais=True, wait=wait, wait_load=wait_load)
+        self.servers[sname].start(silent=True)
         self.connections[sname] = self.servers[sname].admin
         try:
             self.connections[sname]('return true', silent=True)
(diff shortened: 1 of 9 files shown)


lib/preprocessor.py
```diff
@@ -361,4 +360,7 @@ def cleanup_nondefault(self):
             v.cleanup()
 
     def kill_current_test(self):
-        self.servers['default'].kill_current_test()
+        for server in self.servers:
+            if server == 'default':
+                continue
+            self.servers[server].kill_current_test()
```
Totktonada: The idea was we have the one common 'current test' per cluster. Don’t get this change.

lib/tarantool_server.py
```diff
@@ -164,14 +163,21 @@ def execute(self, server):
 }
 
 
-class TarantoolStartError(OSError):
-    pass
+class TarantoolStartError(Exception):
```
Totktonada: Why do you inherit Exception instead of more specifi OSError?

lib/tarantool_server.py
```diff
@@ -569,37 +593,31 @@ def start(self, silent=True, wait=True, wait_load=True, rais=True,
...
 
         port = self.admin.port
         self.admin.disconnect()
         self.admin = CON_SWITCH[self.tests_type]('localhost', port)
         self.status = 'started'
 
+    def wait_load(self, wait_load):
+        try:
+            self.wait_until_started(wait_load)
+        except Exception:
```
Totktonada: I think we should avoid 'except Exception' (at least w/o proper comment what going on). There are few errors can be occured here. The code should expect only that errors.

lib/tarantool_server.py
```diff
@@ -739,15 +758,20 @@ def wait_until_started(self, wait_load=True):
...
+                raise
         while True:
             try:
                 temp = AdminConnection('localhost', self.admin.port)
                 if not wait_load:
                     ans = yaml.load(temp.execute("2 + 2"))
                     return True
                 ans = yaml.load(temp.execute('box.info.status'))[0]
+                if 'error' in ans:
+                    raise Exception(ans)
```
Totktonada: It is better to have a type for the exception.

PR #74, opened 2017-12-02
Handle non-default server crash

Totktonada:
TarantoolStartError is subject to propagate to a joining greenlet,
consider `lib.inspector.gevent_propagate_exc()`.


Totktonada:
Can be checked on current tarantool 1.7 (2db448b21d6a9ccd9ad1dd3c630d7ad9622c3793) with `./test-run.py engine/replica_join`. The vinyl configuration of the test fails, but test-run before the patch doesn’t catch that.

Issue #76, opened 2018-01-25
Add option --verbose


rosik:
Today `lib/test.py` removes test output when it is passed.
I'd like to have an option to see the output.

Proposal:

Add option `--verbose` which prints subtest output to test-run log.

Totktonada:
Implemented in b17fdb4d95b7a051610c4f24312bd9c876b22a51.

PR #77, opened 2018-01-25
Add option --verbose

rosik:
Print subtest output to test-run log
See also: #76

### DIFF
--- a/lib/options.py
+++ b/lib/options.py
@@ -58,6 +58,14 @@ class Options:
                 means find all available.""")
 
         parser.add_argument(
+                "--verbose",
+                dest = 'is_verbose',
+                action = "store_true",
+                default = False,
+                help = """Print TAP13 test output to log.
+                Default: false.""")
+
+        parser.add_argument(
                 "--force",
                 dest = "is_force",
                 action = "store_true",
--- a/lib/test.py
+++ b/lib/test.py
@@ -15,8 +15,8 @@ try:
 except ImportError:
     from StringIO import StringIO
 
+import lib
 from lib.utils import non_empty_valgrind_logs, print_tail_n
-
 from lib.colorer import color_stdout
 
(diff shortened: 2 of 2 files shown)


lib/test.py
```diff
@@ -188,6 +189,11 @@ def run(self, server):
             if self.is_executed_ok and os.path.isfile(self.result):
                 self.is_equal_result = filecmp.cmp(self.result, self.tmp_result)
             elif self.is_executed_ok:
+                if lib.Options().args.is_verbose:
+                    color_stdout('\n')
+                    with open(self.tmp_result, "r+") as logfile:
```
Totktonada: Why 'r+'? Why calling color_stdout() line by line?

Also I think it would be good to add newline in case the file ends w/o endline.

lib/options.py
```diff
@@ -57,6 +57,14 @@ def __init__(self):
                 help = """List of test suites to look for tests in. Default: "" -
                 means find all available.""")
 
+        parser.add_argument(
+                "--verbose",
```
Totktonada: I concerned a bit that the option name is a bit too general then we maybe want. But it is not clear for now, so leave it as is.

lib/options.py
```diff
@@ -57,6 +57,14 @@ def __init__(self):
...
+                dest = 'is_verbose',
+                action = "store_true",
+                default = False,
+                help = """Print test output to log.
```
Totktonada: It is about tests with TAP13 output. I guess it worth to mention in the help.

rosik: Fixed
`"Print TAP13 test output to log"`

lib/test.py
Totktonada: > Why calling color_stdout() line by line?

But from the other side of view it allows to process large files.

PR #93, opened 2018-06-09
Don't abort if 'start server' command fails

### LINKED ISSUE #92: Test is aborted if non-default server fails to start
I want to write a test in Lua that would check that a replica fails to start under certain conditions. So I write something like this:

```
env = require('test_run')
test_run = env.new()
box.error.injection.set(...) -- error injection that makes the replica fail to join
test_run:cmd("create server replica with rpl_master=default, script='replication/replica.lua'")
test_run:cmd("start server replica")
```

The test aborts right after "start server" command with "Result content mismatch" error. Following commands are not executed. I expect the "start server" command to fail gracefully in this case, printing a proper error message to the result file.

This issue blocks testing of https://github.com/tarantool/tarantool/issues/461

locker:
Sometimes, it is useful to check that a tarantool replica fails to start
under certain conditions, but currently it is impossible to do so in
Lua, because test_run:cmd('start server') aborts the test in this case.
This patch makes it print an error message to the result file instead.

Closes #92

### DIFF
--- a/lib/preprocessor.py
+++ b/lib/preprocessor.py
@@ -161,12 +161,12 @@ class TestState(object):
         args = []
         if 'args' in opts:
             args = opts['args'][1:-1].split(' ')
-        self.servers[sname].start(silent=True, rais=True, wait=wait, wait_load=wait_load, args=args)
-        self.connections[sname] = self.servers[sname].admin
         try:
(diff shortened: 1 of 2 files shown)


Totktonada:
Closed in favor of PR #95.

PR #95, opened 2018-06-15
Support expected fail of non-default server

### LINKED ISSUE #92: Test is aborted if non-default server fails to start

Totktonada:
Usage:

```
test_run:cmd("start server <name> with crash_expected=True")
```

Also fixed non-default server crash reporting, regression after
9d73ef1281c59177c7c23aae1371e71b4038e7f1

Fixes #92.

### DIFF
--- a/lib/preprocessor.py
+++ b/lib/preprocessor.py
@@ -161,12 +161,28 @@ class TestState(object):
         args = []
         if 'args' in opts:
             args = opts['args'][1:-1].split(' ')
-        self.servers[sname].start(silent=True, rais=True, wait=wait, wait_load=wait_load, args=args)
-        self.connections[sname] = self.servers[sname].admin
+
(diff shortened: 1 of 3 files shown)


Totktonada:
@locker says he ok.

Issue #102, opened 2018-07-21
Print whole reject file on tap test failure


locker:
Currently, only 10 last lines are printed (although test-run says 15 :-), which makes it impossible to investigate a failure in case the failed sub-test is not the last one. For instance, the output below doesn't give us a clue what happened here:
```
app-tap/tarantoolctl.test.lua                                   [ fail ]
Test failed! Last 15 lines of the result file:
    1..6
    ok - check 'start' command status for 'filler'
    ok - check lsn before
    ok - execution result
    ok - check lsn after
    ok - execution result
    ok - check lsn after
    # fill and test play output: end
ok - fill and test play output
# failed subtest: 1
```
That said, test-run must print the whole reject file produced by a failed tap test, because every its line may contain important information.

The issue blocks investigation of https://github.com/tarantool/tarantool/issues/3557

Totktonada:
Most usual case is (I think) when the last lines gives necessary diagnostics. I would implement #90 instead of this issue to help investigating failures that reproduces only in Travis-CI or very rarely.

locker:
> Most usual case is (I think) when the last lines gives necessary diagnostics.

That's simply not true. For instance, take app-tap/tarantoolctl.test.lua - unless it's one of the last 6 sub-tests that fails, the last 10 lines will always be the same.

> I would implement #90 instead of this issue to help investigating failures that reproduces only in Travis-CI or very rarely.

Why "instead"? Why not "along with"? Do you have any actual reason in mind why test-run shouldn't print the whole tap test output in case of failure? IMO the more information we have, the better. Besides, fixing this issue seems to be much easier than implementing #90.

Totktonada:
> That's simply not true.

If so, it is worth to do.

Totktonada:
@sergw Can you catch this issue?

sergw:
Currently 10 lines of code for client part, and 15 for server part. They are hardcoded and should be configurable. We can add the behaviour to output whole results by enabling two parameters `--verbose` plus non empty metavar test. Also `--verbose` will be added as env (for `make test` cmd).

locker:
Let's always print the whole tap result file. I don't see any point in making it configurable. We print the whole diff between .reject and .result for non-tap tests in case of failure. Why should we limit output of tap tests?

PR #103, opened 2018-07-27
Print whole reject file when test failed

sergw:
Closes: #102

### DIFF
--- a/lib/test.py
+++ b/lib/test.py
@@ -38,6 +38,7 @@ class TestRunGreenlet(gevent.Greenlet):
     def __repr__(self):
             return "<TestRunGreenlet at %s info='%s'>" % (hex(id(self)), getattr(self, "info", None))
 
+
 class FilteredStream:
     """Helper class to filter .result file output"""
     def __init__(self, filename):
@@ -224,7 +225,8 @@ class Test:
 
             where = ""
             if not self.is_crash_reported and not self.is_executed_ok:
-                self.print_diagnostics(self.reject, "Test failed! Last 15 lines of the result file:\n")
+                self.print_diagnostics(self.reject,
+                    "Test failed! Output from reject file {}:\n".format(self.reject))
                 server.print_log(15)
                 where = ": test execution aborted, reason '{0}'".format(diagnostics)
             elif not self.is_crash_reported and not self.is_equal_result:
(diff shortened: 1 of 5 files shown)


lib/tarantool_server.py
```diff
@@ -590,7 +590,7 @@ def start(self, silent=True, wait=True, wait_load=True, rais=True, args=[],
                         self.current_test.is_crash_reported = True
                     color_stdout('\n[Instance "{}"] Tarantool server failed to start\n'.format(
                         self.name), schema='error')
-                    self.print_log(15)
+                    self.print_log()
```
Totktonada: Don’t sure we should change log printing behaviour within the scope of this issue.

lib/tarantool_server.py
```diff
@@ -638,7 +637,7 @@ def crash_grep(self):
                     break
                 if assert_fail_re.match(line):
                     pos = len(lines) - rpos
-                    assert_lines = lines[max(0, pos - print_log_lines):pos]
+                    assert_lines = lines[:pos]
```
Totktonada: Don’t get why we need to perform any assert search if we going to show almost whole log file.

lib/tarantool_server.py
```diff
@@ -668,7 +667,7 @@ def crash_grep(self):
 
         # print log otherwise (if backtrace was not found)
         if not bt:
-            self.print_log(print_log_lines)
+            self.print_log()
```
Totktonada: The same: change of log printing in not what was requested.

lib/test.py
```diff
@@ -224,28 +224,28 @@ def run(self, server):
 
             where = ""
             if not self.is_crash_reported and not self.is_executed_ok:
-                self.print_diagnostics(self.reject, "Test failed! Last 15 lines of the result file:\n")
-                server.print_log(15)
+                self.print_diagnostics(self.reject, "Test failed! Output from reject file {}:\n".format(self.reject))
+                server.print_log()
```
Totktonada: The same: change of log printing in not what was requested.

lib/test.py
```diff
@@ -224,28 +224,28 @@ def run(self, server):
...
                 where = ": test execution aborted, reason '{0}'".format(diagnostics)
             elif not self.is_crash_reported and not self.is_equal_result:
                 self.print_unidiff()
-                server.print_log(15)
+                server.print_log()
```
Totktonada: The same: change of log printing in not what was requested.

lib/tarantool_server.py
sergw: We have logs per test case, so to investigate a failure (as described in the issue) it is good to see the whole log. And reject was just for instance.

locker: AFAIK we use one tarantool log per test suite (such as engine or replication) so it can get quite huge. And it will contain logs left from all test cases within the same suite, not just the failed one. I don't think it's reasonable to print it. The issue was solely about the tap test result file.

PR #105, opened 2018-08-01
test-run: do not reset grep_log on restart

kbelyavs:
Current 'grep_log' behaviour resets search result then server was
restarted.
Suggest to extend this behaviour with optional noreset parameter.
This option needed to test specific behaviout on server shutdown.

Needed for #3159

### DIFF
--- a/test_run.lua
+++ b/test_run.lua
@@ -215,7 +215,7 @@ local function get_cfg(self, name)
     return self.run_conf[name]
 end
 
-local function grep_log(self, node, what, bytes)
+local function grep_log(self, node, what, bytes, noreset)
     local filename = self:eval(node, "box.cfg.log")[1]
     local file = fio.open(filename, {'O_RDONLY', 'O_NONBLOCK'})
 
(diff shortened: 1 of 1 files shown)


test_run.lua
```diff
@@ -240,7 +240,8 @@ local function grep_log(self, node, what, bytes)
     if file:seek(-bytes, 'SEEK_END') == nil then
         fail("Failed to seek log file")
     end
-    local found, buf
+    local buf, found = nil
```
Totktonada: You are set buf to nil explicitly and found implicitly, it is inconsistent. As I see we mostly prefer to omit explicit `nil`.

test_run.lua
```diff
@@ -215,7 +215,7 @@ local function get_cfg(self, name)
     return self.run_conf[name]
 end
 
-local function grep_log(self, node, what, bytes)
+local function grep_log(self, node, what, bytes, noreset)
```
Totktonada: I would add `opts` parameter with `noreset` field to give a reader an idea what a parameter means in a place where the function invoked. It also allow to add more options in the future w/o large cout of parameters.

So, the code will lokk like so:

```
local function grep_log(self, node, what, bytes, opts)
    local opts = opts or {}
    local noreset = opts.noreset or false
    ...
```

Totktonada:
Added @sergw as reviewer.

test_run.lua
sergw: 👍 **+1**

PR #108, opened 2018-08-02
Print reproduce file when test failed

### LINKED ISSUE #104: Print reproduce file in case of a test fail
It will allow to investigate Travis CI fails.

In some cases it can be more easy to just show logs instead of download and unpack artefacts (#90).

sergw:
- When test-run runs in a consistent mode, there is no reproduce information.
This mode is used in CI. It is convenient to see the contents of the reproduce
file in the CI log.

- If test fails in parallel mode, print the reproduce file content

- Fix some code health by using `flake8`

### DIFF
--- a/test-run.py
+++ b/test-run.py
@@ -158,6 +158,13 @@ def main_loop_consistent(failed_test_ids):
         for task_id in task_ids:
             short_status = worker.run_task(task_id)
             if short_status == 'fail':
+                reproduce_file_path = \
+                    lib.worker.get_reproduce_file(worker.name)
+                color_stdout('Reproduce file %s\n' %
+                             reproduce_file_path, schema='error')
+                color_stdout("---\n", schema='separator')
+                lib.utils.print_tail_n(reproduce_file_path)
+                color_stdout("...\n", schema='separator')
                 failed_test_ids.append(task_id)
                 if not lib.Options().args.is_force:
                     worker.stop_server(cleanup=False)
--- a/test-run.py
+++ b/test-run.py
@@ -49,23 +49,17 @@
 #   tests passed. Otherwise non-zero.
 
 
+import multiprocessing
 import os
-import signal
 import sys
 import time
 
-import subprocess
-import collections
-import multiprocessing
-
 import lib
 import lib.worker
-from lib.colorer import color_stdout
-from lib.utils import signame, format_process
-from lib.tarantool_server import TarantoolServer
 
-from listeners import HangError
 from dispatcher import Dispatcher
+from lib.colorer import color_stdout
+from listeners import HangError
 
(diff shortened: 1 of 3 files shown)


test-run.py
```diff
@@ -88,16 +86,16 @@ def main_loop_parallel():
                      schema='tr_text')
     randomize = True
 
-    task_groups = lib.worker.get_task_groups()
+    task_groups = get_task_groups()
```
Totktonada: Maybe there is sense in visual differentiation of a 'thread / worker' code from the 'main process / dispatcher' code. I modestly vote for calling lib / worker code from outside of lib as 'lib.worker.<...>' or at least 'worker.<...>'.

test-run.py
```diff
@@ -163,12 +161,22 @@ def main_loop_consistent(failed_test_ids):
         for task_id in task_ids:
             short_status = worker.run_task(task_id)
             if short_status == 'fail':
+                reproduce_file_path = get_reproduce_file(worker.name)
+                if os.path.exists(reproduce_file_path):
```
Totktonada: I propose to wrap this code into a function and place it in 'Utils' section in worker.py in the future follow up commit (where parallel mode will gain support of the reproduce file printing).

test-run.py
```diff
@@ -163,12 +161,22 @@ def main_loop_consistent(failed_test_ids):
...
+                    with open(reproduce_file_path, 'r') as reproduce_file:
+                        reproduce_content = reproduce_file.read()
+                        color_stdout('Reproduce file %s\n' %
+                                     reproduce_file_path, schema='error')
+                        color_stdout(reproduce_content, schema='serv_text')
+                        color_stdout('\n', '-' * 75, "\n", schema='separator')
```
Totktonada: The first '\n' is redundant, because a test file always ends with the newline symbol (by the definition).

`"\n"` differs from any surrounding string by quote type (feels as inconsistency).

Don’t sure the separator (75 dashes) is needed here. Maybe it would be better to use yaml document syntax to visual highlight the file content. See as tarantool itself format a console output:

```
tarantool> 1, 2, 3
---
- 1
- 2
- 3
...
```

It starts with three dashes and ends with three periods.

Totktonada:
> I looked around briefly and I think [1] is appropriate place to add the reproduce file printing.

Or maybe [2] is more natural for that.

[1]: https://github.com/tarantool/test-run/blob/97f81e4c99c9d2a28026a4d6ee9d57cebd3f5fe0/lib/worker.py#L291
[2]: https://github.com/tarantool/test-run/blob/97f81e4c99c9d2a28026a4d6ee9d57cebd3f5fe0/listeners.py#L58

Issue #107, opened 2018-08-02
Print output of hung test after killing it


locker:
If a test hangs, the output of test-run will look like this:
```
No output during 10 seconds. List of workers not reporting the status: [1]; Will abort after 120 seconds without output.
No output during 20 seconds. List of workers not reporting the status: [1]; Will abort after 120 seconds without output.
No output during 30 seconds. List of workers not reporting the status: [1]; Will abort after 120 seconds without output.
No output during 40 seconds. List of workers not reporting the status: [1]; Will abort after 120 seconds without output.
No output during 50 seconds. List of workers not reporting the status: [1]; Will abort after 120 seconds without output.
No output during 60 seconds. List of workers not reporting the status: [1]; Will abort after 120 seconds without output.
No output during 70 seconds. List of workers not reporting the status: [1]; Will abort after 120 seconds without output.
No output during 80 seconds. List of workers not reporting the status: [1]; Will abort after 120 seconds without output.
No output during 90 seconds. List of workers not reporting the status: [1]; Will abort after 120 seconds without output.
No output during 100 seconds. List of workers not reporting the status: [1]; Will abort after 120 seconds without output.
No output during 110 seconds. List of workers not reporting the status: [1]; Will abort after 120 seconds without output.
No output during 120 seconds. List of workers not reporting the status: [1]; Will abort after 120 seconds without output.

[Main process] No output from workers. It seems that we hang. Send SIGKILL to workers; exiting...
---------------------------------------------------------------------------------
Count of didn't processed tasks: 1
```
That is, looking at the output it is impossible to say what happened. It would be great if test-run printed the output of hung tests (diff or tap result).

PR #110, opened 2018-08-10
List tasks information for hung workers

sergw:
If HungListener found some hung tests it will print information about:
- worker id
- test name
- test params
- output from .result file

Part of #107


sergw:
Checked on some hung tests / result empty. Looks like needs to flush.

dispatcher.py
```diff
@@ -78,10 +78,14 @@ def __init__(self, task_groups, max_workers_cnt, randomize):
         self.worker_next_id = 1
 
         tasks_cnt = 0
+        self.current_task_queue = SimpleQueue()
```
Totktonada: Why the new queue is needed?

Totktonada: Sergei (@sergw) helps me to remember the context: workers are separate processes, so we need a queue to pass information about a test that a worker receive.

I propose the following approach. A worker sends an object of specific type to the results queue (see 'Worker results' at worker.py:113) when receive a task. The HangWatcher listener (listeners.py) checks for this type of an object in `process_result` and saves necessary info. Then it can be used in `process_timeout` method.

So, we’ll avoid usage of extra queue and we’ll reuse our flexible approach to handle events depending of its type.

listeners.py
```diff
@@ -186,14 +190,30 @@ def process_result(self, obj):
...
+            "List of workers not reporting the status:\n" % (
+                self.inactivity, self.kill_timeout),
                 schema='test_var')
+        for worker_id in worker_ids:
+            task_name, task_param = self.get_task_by_worker_id(worker_id)
+            color_stdout("[{0:03d}] {1} {2}\n".format(worker_id,
+                                                  task_name,
+                                                  task_param or ''),
+                         schema='test_var')
+            task_path = "{0:03d}_{1}".format(worker_id,
```
Totktonada: I think it is better to get a path to result file from a worker and don’t try to reconstruct it here.

lib/worker.py
```diff
@@ -285,6 +286,8 @@ def run_loop(self, task_queue, result_queue):
                           schema='test_var')
                 self.stop_worker(task_queue, result_queue)
                 break
+
+            current_task_queue.put((self.id, task_id))
```
Totktonada: Please, create a subclass of BaseWorkerMessage to hold the data.

PR #111, opened 2018-08-10
Allow to pass arguments to create_cluster.

sergepetrenko:
Just like in case of `start server replica with args='1 2 3'`, it is
now possible to pass arguments to create_cluster, so that all instances
are started whith these arguments.

Also alter create_cluster a little bit: previous behaviour was to create
and start instances in a loop, change it to first creating all
instances and then starting them. This is needed because creating an
instance may take up a lot of time, if the old vardir isn't empty, and
the first started instance would have to wait for others for a
significant period of time, which makes it exit upon timeout if
replication_connect_timeout is relatively small.

Needed for tarantool/tarantool#3428

### DIFF
--- a/test_run.lua
+++ b/test_run.lua
@@ -94,6 +94,7 @@ local create_cluster_cmd1 = 'create server %s with script="%s/%s.lua"'
 local create_cluster_cmd1_return_listen_uri =
     'create server %s with script="%s/%s.lua", return_listen_uri=True'
 local create_cluster_cmd2 = 'start server %s with wait_load=False, wait=False'
+local create_cluster_cmd2_args = create_cluster_cmd2 .. ', args="%s"'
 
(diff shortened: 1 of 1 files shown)


test_run.lua
```diff
@@ -94,6 +94,7 @@ local create_cluster_cmd1 = 'create server %s with script="%s/%s.lua"'
 local create_cluster_cmd1_return_listen_uri =
     'create server %s with script="%s/%s.lua", return_listen_uri=True'
 local create_cluster_cmd2 = 'start server %s with wait_load=False, wait=False'
+local create_cluster_cmd2_args = create_cluster_cmd2..', args="%s"'
```
Totktonada: Please, follow our Lua style guide and add whitespaces around `..`.

PR #112, opened 2018-08-10
Support comments in config files

Totktonada:


lib/test_suite.py
```diff
@@ -48,9 +49,11 @@ def get_multirun_conf(self, suite_path):
         result = None
         with open(path) as cfg:
             try:
-                result = json.load(cfg)
+                content = cfg.read()
+                content = re.sub(r'^\s+//.*$', '', content, flags=re.M)
```
sergw: Looks like regex is not correct, it will only cut comments starting with speces in one line:

<pre>
// not this comment
{
    // this comment will be cleared
    "analyze9.test.lua": { // not this comment
        "memtx": {"engine": "memtx"}
    },
    "*": {
        "memtx": {"engine": "memtx"},
        "vinyl": {"engine": "vinyl"}
    }
}
</pre>

Totktonada: Ouch, thanks! Fixed.

PR #114, opened 2018-08-16
 Don't kill master server if replica crash was expected

sergw:


lib/tarantool_server.py
```diff
@@ -106,7 +106,14 @@ def killall_servers(self, server, ts, crash_occured):
 
         # check that all servers stopped correctly
         for server in check_list:
-            crash_occured = crash_occured or server.process.returncode not in (None, 0, -signal.SIGKILL, -signal.SIGTERM)
+            bad_returncode = server.process.returncode not in (None,
+                                                               0,
+                                                               -signal.SIGKILL,
+                                                               -signal.SIGTERM)
+            # if replica crashed but this crash is exptected
```
Totktonada: exptected → expected

Totktonada: From test-run point there are default and non-default servers. We can write it as 'non-default server (say, replica)'.

lib/tarantool_server.py
```diff
@@ -106,7 +106,14 @@ def killall_servers(self, server, ts, crash_occured):
...
+            # don't kill the master server and crash detectors
```
Totktonada: master server → default server (say, replication master)

sergepetrenko:
@Totktonada The fix works fine.

Issue #117, opened 2018-08-27
tarantool inherits file descriptors open by test-run


locker:
Tarantool instance started by test-run inherits all its file descriptors, which includes open logs left from dead instances. As a result, we can't temporarily lower ulimit in a test to check that there's no fd leak. To fix this, test-run should either close logs left from stopped instances or mark the corresponding file descriptors with O_CLOEXEC.

Blocks: https://github.com/tarantool/tarantool/issues/3642

How to reproduce:
1. Check out [g.kirichenko/gh-3642-fix-replication-socket-leak](https://github.com/tarantool/tarantool/tree/g.kirichenko/gh-3642-fix-replication-socket-leak)
2. Run replication tests sequentially:
```
./test-run -j -1 replication/
```
3. `replication/misc.test.lua` hangs. Here's what `/proc/pid/fd` shows:
```
total 0
lrwx------ 1 vlad vlad 64 Aug 27 18:26 0 -> /dev/pts/3
l-wx------ 1 vlad vlad 64 Aug 27 18:26 1 -> /home/vlad/src/tarantool/test/var/001_replication/master.log
lrwx------ 1 vlad vlad 64 Aug 27 18:26 10 -> socket:[18044038]
lrwx------ 1 vlad vlad 64 Aug 27 18:26 100 -> socket:[18038292]
l-wx------ 1 vlad vlad 64 Aug 27 18:26 101 -> /home/vlad/src/tarantool/test/var/001_replication/replica1.log
lrwx------ 1 vlad vlad 64 Aug 27 18:26 102 -> socket:[18043910]
l-wx------ 1 vlad vlad 64 Aug 27 18:26 103 -> /home/vlad/src/tarantool/test/var/001_replication/replica2.log
l-wx------ 1 vlad vlad 64 Aug 27 18:26 104 -> /home/vlad/src/tarantool/test/var/001_replication/proxy.log
l-wx------ 1 vlad vlad 64 Aug 27 18:26 105 -> /home/vlad/src/tarantool/test/var/001_replication/replica3.log
l-wx------ 1 vlad vlad 64 Aug 27 18:26 106 -> /home/vlad/src/tarantool/test/var/001_replication/replica3.log
lrwx------ 1 vlad vlad 64 Aug 27 18:26 107 -> socket:[18041864]
lrwx------ 1 vlad vlad 64 Aug 27 18:26 108 -> socket:[18041865]
[...]

PR #121, opened 2018-08-29
Don't inherit queues file descriptors

Totktonada:
Follows up #117 and PR #118.


Totktonada:
FYI: Force-pushed simplified patch.

PR #124, opened 2018-10-01
Perform cleanup before each test if param used

sergw:
By using param `pre_cleanup = True` test-run will copy
`pre_cleanup.lua` file then execute `clean()` from it before each
test in suite.

Part of https://github.com/tarantool/tarantool/issues/1961

### DIFF
--- a/lib/app_server.py
+++ b/lib/app_server.py
@@ -27,9 +27,7 @@ def run_server(execs, cwd, server, logfile, retval):
 
 class AppTest(Test):
     def execute(self, server):
-        server.current_test = self
-        if self.suite_ini['pre_cleanup']:
-            server.pre_cleanup()
+        super(AppTest, self).execute(server)
         ts = TestState(self.suite_ini, None, TarantoolServer,
                        self.run_params,
                        default_server_no_connect=server)
--- a/lib/app_server.py
+++ b/lib/app_server.py
@@ -28,6 +28,8 @@ def run_server(execs, cwd, server, logfile, retval):
 class AppTest(Test):
     def execute(self, server):
         server.current_test = self
+        if self.suite_ini['pre_cleanup']:
+            server.pre_cleanup()
         ts = TestState(self.suite_ini, None, TarantoolServer,
                        self.run_params,
                        default_server_no_connect=server)
@@ -57,11 +59,6 @@ class AppServer(Server):
         Server.__init__(self, ini, test_suite)
         self.testdir = os.path.abspath(os.curdir)
         self.vardir = ini['vardir']
-        self.re_vardir_cleanup += [
-            "*.snap", "*.xlog", "*.vylog", "*.inprogress",
-            "*.sup", "*.lua", "*.pid"
-        ]
-        self.cleanup()
         self.builddir = ini['builddir']
         self.debug = False
         self.lua_libs = ini['lua_libs']
--- a/lib/server.py
+++ b/lib/server.py
@@ -80,7 +80,7 @@ class Server(object):
             'inspector_port', self.DEFAULT_INSPECTOR
         ))
 
-        # filled in {Test,FuncTest,LuaTest,PythonTest}.execute()
+        # filled in {Test,AppTest,LuaTest,PythonTest}.execute()
         # or passed through execfile() for PythonTest (see
         # TarantoolServer.__init__).
         self.current_test = None
@@ -95,15 +95,9 @@ class Server(object):
     def pre_cleanup(self):
         self.cleanup()
 
-    def cleanup(self, full=False, dirname='.', patterns=DEFAULT_CHECKPOINT_PATTERNS):
+    def cleanup(self, dirname='.'):
         waldir = os.path.join(self.vardir, dirname)
-        if full:
-            try:
-                shutil.rmtree(waldir)
-            except OSError:
-                pass
-            return
-        for pattern in patterns:
+        for pattern in DEFAULT_CHECKPOINT_PATTERNS:
             for f in glob.glob(os.path.join(waldir, pattern)):
                 if os.path.isdir(f):
                     shutil.rmtree(f)
--- a/lib/server.py
+++ b/lib/server.py
@@ -11,6 +11,10 @@ from lib.colorer import color_stdout
 from lib.utils import print_tail_n
 
 
+DEFAULT_CHECKPOINT_PATTERNS = ["*.snap", "*.xlog", "*.vylog", "*.inprogress",
+                               "[0-9]*/"]
+
+
 class Server(object):
     """Server represents a single server instance. Normally, the
     program operates with only one server, but in future we may add
@@ -88,12 +92,19 @@ class Server(object):
     def prepare_args(self, args=[]):
         return args
 
-    def cleanup(self, full=False):
+    def pre_cleanup(self):
+        self.cleanup()
+
+    def cleanup(self, full=False, dirname='.', patterns=DEFAULT_CHECKPOINT_PATTERNS):
+        waldir = os.path.join(self.vardir, dirname)
         if full:
-            shutil.rmtree(self.vardir)
+            try:
+                shutil.rmtree(waldir)
+            except OSError:
+                pass
             return
-        for re in self.re_vardir_cleanup:
-            for f in glob.glob(os.path.join(self.vardir, re)):
+        for pattern in patterns:
+            for f in glob.glob(os.path.join(waldir, pattern)):
                 if os.path.isdir(f):
                     shutil.rmtree(f)
                 else:
--- a/lib/server.py
+++ b/lib/server.py
@@ -21,16 +21,18 @@ class Server(object):
     replication slaves. The server is started once at the beginning
     of each suite, and stopped at the end."""
     DEFAULT_INSPECTOR = 0
-    TEST_RUN_DIR = os.path.abspath(os.path.join(os.path.dirname(__file__), ".."))
+    TEST_RUN_DIR = os.path.abspath(os.path.join(os.path.dirname(__file__),
+                                                ".."))
 
     @property
     def vardir(self):
         if not hasattr(self, '_vardir'):
             raise ValueError("No vardir specified")
         return self._vardir
+
     @vardir.setter
     def vardir(self, path):
-        if path == None:
+        if path is None:
             return
         self._vardir = os.path.abspath(path)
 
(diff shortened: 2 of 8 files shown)


pre_cleanup.lua
```diff
@@ -0,0 +1,47 @@
...
+        return tuple[3]
+    end):filter(function(name)
+        return not allowed_roles[name]
+    end):each(function(name)
+        box.schema.role.drop(name)
+    end)
+
+    os.execute(string.format('rm -rf *.snap *.xlog'))
+end
+
```
Totktonada: I think we should clean _func and _cluster too.

Totktonada: It should be good to have asserts like so: https://gist.github.com/bigbes/47e9e5c5cfb9245dbda94fd0d6c3ed11#file-example-lua-L46 to check that the clean up really do its work.

pre_cleanup.lua
```diff
@@ -0,0 +1,47 @@
...
+return {
+    clean = clean_state;
```
Totktonada: Nit: why do you use two different names?

pre_cleanup.lua
```diff
@@ -0,0 +1,47 @@
...
+        return tuple[4] == 'user'
+    end):map(function(tuple)
+        return tuple[3]
+    end):filter(function(name)
+        return not allowed_users[name]
+    end):each(function(name)
+        box.schema.user.drop(name)
+    end)
+    local allowed_roles = {
+        public = true,
```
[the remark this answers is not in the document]
sergw: As I see we do not have `guest` **role**

pre_cleanup.lua
```diff
@@ -0,0 +1,47 @@
...
+        return tuple[4] == 'role'
+    end):map(function(tuple)
+        return tuple[3]
+    end):filter(function(name)
+        return not allowed_roles[name]
+    end):each(function(name)
+        box.schema.role.drop(name)
+    end)
+
+    os.execute(string.format('rm -rf *.snap *.xlog'))
```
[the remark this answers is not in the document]
sergw: > 2.  core = 'app' tests also need xlogs/snapshots deletion?

As I know `app` tests uses new tarantool instance per test, unlike `tarantool` tests uses tarantool instance per worker. So the `pre_cleanup` not needed?

pre_cleanup.lua
```diff
@@ -0,0 +1,47 @@
+local os = require('os')
```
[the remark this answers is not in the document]
sergw: `copy_files` executes only once per worker instalation process, put cleanup checkpoints in `execute` method

pre_cleanup.lua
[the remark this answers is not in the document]
Totktonada: Sorry, it is so.

pre_cleanup.lua
```diff
@@ -0,0 +1,47 @@
...
+
+local function clean_state(self)
+    box.space._space:pairs():map(function(tuple)
+        return tuple[3]
```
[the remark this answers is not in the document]
Totktonada: It is more or less clear, but I think the rule of thumb is to avoid such unnamed constants in any code (maybe except some bit manipulation involving constants 8, 16 and so on).

pre_cleanup.lua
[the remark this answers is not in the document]
Totktonada: But where the pre_cleanup.lua file copied into a working directory?

pre_cleanup.lua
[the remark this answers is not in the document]
Totktonada: xlogs/snapshots files remain on a disk after a test run. BTW, it can be convenient for debugging of a failing test.

pre_cleanup.lua
sergw: I guess adding asserts, will be enough:
```
assert(#box.space._func:select{} == 1)
assert(#box.space._cluster:select{} == 1)
```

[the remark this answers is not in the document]
Totktonada: Don't think the list legth is matter here. It is pretty easy to collect it: `for k, _ in pairs(_G) do print(k) end` on clean tarantool.

Totktonada: Will it fail the test in case on failing assert?

pre_cleanup.lua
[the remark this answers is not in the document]
Totktonada: My comment was about constant 3. I would define a 'constant' like `local _SPACE_NAME_FN = 3` above and use it here.

lib/tarantool_server.py
```diff
@@ -144,6 +147,12 @@ def execute(self, server):
             self.run_params
         )
         self.inspector.set_parser(ts)
+
+        if self.suite_ini.get('pre_cleanup', False):
```
Totktonada: So it will work only for 'core = tarantool' tests. What about 'core = app'?

pre_cleanup.lua
sergw: Yes, the test fails.

lib/tarantool_server.py
sergw: Yes it will works only for 'core = tarantool' test. As I mention above 'core = app' not needed cleanup.

pre_cleanup.lua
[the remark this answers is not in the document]
sergw: Also I dont think garbage globals can affect some tests.

pre_cleanup.lua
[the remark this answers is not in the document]
Totktonada: Don't sure where a line btw roles and users are, but it seems we have 'guest' record in the _user space. Reopened.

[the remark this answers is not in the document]
sergw: `guest` and `admin` already in `allowed_users`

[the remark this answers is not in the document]
Totktonada: Sorry. Got it.

lib/app_server.py
```diff
@@ -28,6 +28,10 @@ def run_server(execs, cwd, server, logfile, retval):
 class AppTest(Test):
     def execute(self, server):
         server.current_test = self
+        if self.suite_ini.get('pre_cleanup', False):
+            for regx in server.re_vardir_checkpoints:
+                for f in glob.glob(os.path.join(server.vardir, regx)):
+                    os.remove(f)
```
Totktonada: Please, move it to Server of Tarantool/AppServer fucncition(s) with or without the condition.

lib/app_server.py
```diff
@@ -28,6 +28,10 @@ def run_server(execs, cwd, server, logfile, retval):
 class AppTest(Test):
     def execute(self, server):
         server.current_test = self
+        if self.suite_ini.get('pre_cleanup', False):
```
Totktonada: It would be good to do the check just like `if self.suite_ini['pre_cleanup']:` if possible.

pre_cleanup.lua
```diff
@@ -0,0 +1,205 @@
...
+        title = true,
+        uri = true,
+        utf8 = true,
+        uuid = true,
+        xlog = true,
+        yaml = true,
+    }
+    cleanup_list(package.loaded, allowed_packages)
+
+    local user_count = #box.space._user:select{}
```
Totktonada: Nit: We can use `box.space._user:select:count()`.

Totktonada: And below too.

Totktonada:
Finally we found that cleanup should not be called during install / deploy (because now we always run a server on clean var directory) and it should not remove anything except wal / snap files when called explicitly ('restart server default with cleanup=1' and so on). It is going to be rewritten so.

lib/app_server.py
```diff
@@ -86,6 +83,7 @@ def prepare_args(self, args=[]):
 
     def deploy(self, vardir=None, silent=True, need_init=True):
         self.vardir = vardir
+        self.workdir = self.vardir
```
Totktonada: It is something unrelated to the PR?

lib/tarantool_server.py
```diff
@@ -551,6 +557,9 @@ def copy_files(self):
         shutil.copy('.tarantoolctl', self.vardir)
         shutil.copy(os.path.join(self.TEST_RUN_DIR, 'test_run.lua'),
                     self.vardir)
+        if self.ini.get('pre_cleanup', False):
```
Totktonada: Can be simplified to self.ini['pre_cleanup']?

lib/app_server.py
Totktonada: Or some renaming?

Totktonada: Ok, got it, that is split of vardir (scripts and so on) and workdir of tarantool. I suggest to mention it in the commit message, because it is unobvious refactoring.

lib/server.py
```diff
@@ -88,25 +93,30 @@ def __init__(self, ini, test_suite=None):
...
 
-    def cleanup(self, full=False):
-        if full:
-            shutil.rmtree(self.vardir)
-            return
-        for re in self.re_vardir_cleanup:
-            for f in glob.glob(os.path.join(self.vardir, re)):
+    def pre_cleanup(self):
+        if self.ini['pre_cleanup']:
+            self.cleanup()
```
Totktonada: Beside cleanup of that server we need to cleanup xlog/snap of non-default servers were run within previous test. We don't know non-default server names, so need to either save them somewhere or guess them from directory names inside vardir.

lib/tarantool_server.py
```diff
@@ -129,6 +136,7 @@ def killall_servers(self, server, ts, crash_occured):
 
     def execute(self, server):
         server.current_test = self
+        server.pre_cleanup()
```
Totktonada: Finally I understood we should not remove xlogs for working tarantool server. But we should do it for non-default servers and app server.

lib/server.py
Totktonada: Sergei found that we should not clean non-default servers in pre_cleanup, because it already cleaned at `create server`.

sergw:
Updated but something goes wrong. If enable on some tests (ex: engine/xlog), it fails on `server restart default`. May be it somehow connected with proxy.

lib/tarantool_server.py
```diff
@@ -551,10 +557,28 @@ def copy_files(self):
...
+        if self.ini.get('pre_cleanup', False):
+            shutil.copy(os.path.join(self.TEST_RUN_DIR, 'pre_cleanup.lua'),
+                        self.vardir)
 
     def prepare_args(self, args=[]):
         return [self.ctl_path, 'start', os.path.basename(self.script)] + args
 
+    def pre_cleanup(self):
+        # If current test make a restart 'default' server it can be affected
+        # by snapshots from previous test [wal_mode=off]. We have only one
```
Totktonada: Proposed to don't handle wal_off cases at all.

lib/server.py
```diff
@@ -80,7 +80,7 @@ def __init__(self, ini, test_suite=None):
             'inspector_port', self.DEFAULT_INSPECTOR
         ))
 
-        # filled in {Test,FuncTest,LuaTest,PythonTest}.execute()
+        # filled in {Test,AppTest,LuaTest,PythonTest}.execute()
```
Totktonada: In FuncTest too.

Totktonada: Got it, you removed it. Sorry.

lib/server.py
```diff
@@ -88,12 +92,19 @@ def __init__(self, ini, test_suite=None):
     def prepare_args(self, args=[]):
         return args
 
-    def cleanup(self, full=False):
+    def pre_cleanup(self):
+        self.cleanup()
+
+    def cleanup(self, full=False, dirname='.', patterns=DEFAULT_CHECKPOINT_PATTERNS):
```
Totktonada: `full` is not used, proposed to remove.

Totktonada: Other parameters are also don't used. Proposed to remove too.

I'm against introducing extra complexity until it is needed.

sergw: fixed

lib/tarantool_server.py
sergw: Need to use get here because of nondefault servers doesn't have ini

Totktonada: It seems that self.ini values used in *Server are mostly propagated to *Server fields, but I don't want to push you to rewrite it so.

PR #128, opened 2018-10-16
test: wait for logs

sergw:
Instead of sleep by constant time and then checking for some condition
use `wait_cond` to wait for condition in timeout period.

Using `wait_cond` to wrap `grep_log`.

### DIFF
--- a/test_run.lua
+++ b/test_run.lua
@@ -5,6 +5,7 @@ local log = require('log')
 local fiber = require('fiber')
 local fio = require('fio')
 local errno = require('errno')
+local clock = require('clock')
 
 local function cmd(self, msg)
     local sock = socket.tcp_connect(self.host, self.port)
@@ -284,6 +285,26 @@ local function grep_log(self, node, what, bytes, opts)
     return found
 end
 
+-- Checks fn with delay until timeout or fn returns true, otherwise returns false
+local function wait_cond(self, fn, timeout, delay)
+    assert(timeout ~= nil)
+    delay = delay or 0.001
+    local t1 = clock.monotonic()
+    while not fn() do
+        local work_time = clock.monotonic() - t1
+        if work_time > timeout then
+            return false
+        end
+        fiber.sleep(delay)
+    end
+    return true
+end
+
+-- Wrapper for grep_log, wait until expected log entry is appear in server log file
+local function wait_log(self, node, what, bytes, timeout, opts)
+    return wait_cond(self, function() return grep_log(self, node, what, bytes, opts) ~= nil end, timeout)
+end
+
 local inspector_methods = {
     cmd = cmd,
     eval = eval,
@@ -307,6 +328,8 @@ local inspector_methods = {
     wait_cluster_vclock = wait_cluster_vclock,
     --
     grep_log = grep_log,
+    wait_cond = wait_cond,
+    wait_log = wait_log,
 }
 
 local function inspector_new(host, port)
--- a/test_run.lua
+++ b/test_run.lua
@@ -285,24 +285,37 @@ local function grep_log(self, node, what, bytes, opts)
     return found
 end
 
--- Checks fn with delay until timeout or fn returns true, otherwise returns false
-local function wait_cond(self, fn, timeout, delay)
+-- Checks cmp(fn()) with delay until timeout or returns fn().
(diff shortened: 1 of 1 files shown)


test_run.lua
```diff
@@ -284,6 +285,26 @@ local function grep_log(self, node, what, bytes, opts)
     return found
 end
 
+-- Checks fn with delay until timeout or fn returns true, otherwise returns false
+local function wait_cond(self, fn, timeout, delay)
+    assert(timeout ~= nil)
+    delay = delay or 0.001
+    local t1 = clock.monotonic()
```
Totktonada: Don't sure about monotonic, but monotonic64 is fast enough, because does not do timezone-related work. Can you check it or change?

sergw: The `monotonic` and `monotonic64` uses same method to get time:
```
clock_gettime(CLOCK_MONOTONIC, &ts);
```

Totktonada: Thanks!

test_run.lua
```diff
@@ -284,6 +285,26 @@ local function grep_log(self, node, what, bytes, opts)
...
+    while not fn() do
+        local work_time = clock.monotonic() - t1
+        if work_time > timeout then
+            return false
+        end
+        fiber.sleep(delay)
+    end
+    return true
```
Totktonada: I think it would be good to return the value the function `fn` returns. It is worth to do as in the success case as well as in the fail case (it can be nil or false).

test_run.lua
```diff
@@ -284,6 +285,26 @@ local function grep_log(self, node, what, bytes, opts)
     return found
 end
 
+-- Checks fn with delay until timeout or fn returns true, otherwise returns false
```
Totktonada: Nit: Our code style suggests to keep comments within 66 symbols, start it with an uppercase letter and end with a period.

test_run.lua
```diff
@@ -284,6 +285,26 @@ local function grep_log(self, node, what, bytes, opts)
...
+        local work_time = clock.monotonic() - t1
+        if work_time > timeout then
+            return false
+        end
+        fiber.sleep(delay)
+    end
+    return true
+end
+
+-- Wrapper for grep_log, wait until expected log entry is appear in server log file
```
Totktonada: Nit: code style (see above).

test_run.lua
```diff
@@ -284,6 +285,26 @@ local function grep_log(self, node, what, bytes, opts)
...
+local function wait_log(self, node, what, bytes, timeout, opts)
+    return wait_cond(self, function() return grep_log(self, node, what, bytes, opts) ~= nil end, timeout)
```
Totktonada: I think we should not discard the value returned from grep_log, just bind needed arguments (perform carrying of the function). Also we can allow a user to set delay. Proposed (didn't tested):

```
local delay = opts.delay
opts.delay = nil
local cond = function()
    return grep_log(self, node, what, bytes, opts)
end
```

test_run.lua
```diff
@@ -284,6 +285,39 @@ local function grep_log(self, node, what, bytes, opts)
     return found
 end
 
+-- Checks cmp(fn()) with delay until timeout or returns fn().
+-- Comparator cmp should return true if it what we waiting for.
+local function wait_cond(self, fn, timeout, delay, cmp)
```
Totktonada: I think `cmp` is redundant here, we can just return what `fn` gives us and implement some complex behavious with custom `fn`. This way we leave the common code easy to read, while do not lost its powerfulness.

test_run.lua
```diff
@@ -284,6 +285,39 @@ local function grep_log(self, node, what, bytes, opts)
...
+end
+
+-- Wrapper for grep_log, wait until expected log entry is appear
+-- in server log file.
+local function wait_log(self, node, what, bytes, timeout, opts)
+    local opts = opts or {}
+    local cond = function()
+        return grep_log(self, node, what, bytes, opts)
+    end
+    local cmp = opts.cmp or function(val) return val ~= nil end
```
Totktonada: As said above I'm against `cmp`, because it introduces extra complexity into the common code.

Totktonada:
Pushed in 670f330aacaf44bc8b1f969fa0cd5f811c5ceb1b with minor changes.

PR #132, opened 2018-11-21
Set cloexec before connect

sergw:
When using use_unix_sockets, the file descriptors are not closed after
the default server was restart, and tests that limit number of
open files fail.

Fixes: https://github.com/tarantool/tarantool/issues/3806


Totktonada:
The reason is that greenlet with Popen can wakeup while connection greenlet waiting for connect(). Don't know why it is not observed with tcp sockets.

It is does not matter for sockets itself whether we sef FD_CLOEXEC before or after connect(). Consider the following gist: https://gist.github.com/Totktonada/d11eb15cdf589e5b45f477bc270c5da2

Totktonada:
Pushed to master as b45fbfabeb975d637284bdc642d17c00e3082cc2 (added comments).

PR #138, opened 2019-01-21
Fixes around --long option and results reporting

Totktonada:


lib/utils.py
```diff
@@ -40,12 +40,15 @@ def non_empty_valgrind_logs(paths_to_log):
...
+def print_tail_n(filename, num_lines=None, numerate=True):
     """Print N last lines of a file. If num_lines is not set, prints the whole file"""
     with open(filename, "r") as logfile:
-        tail_n = collections.deque(logfile, num_lines)
-        for line in tail_n:
-            color_stdout(line, schema='tail')
+        tail_n = collections.deque(enumerate(logfile, 1), num_lines)
+        for num, line in tail_n:
+            if numerate:
+                color_stdout('{0:6d} {1:s}'.format(num, line), schema='tail')
```
locker: So you print line numbers when you show result/log file of a hung test. I don't think it makes much sense - I just don't see why anybody would need this information at all. For a result file I could explain this as such: one wants to quickly jump to the line in the result file committed to the repository where the test hung so that the line number might come in handy, but you already print the line number in warning messages. For log files there's no point in printing line numbers at all as logs have timestamps.

Come to think of it, why don't you print the diff when a test hangs, just like when the test fails? IMO it would make it much easier to figure out what went wrong by looking at the test output.

Totktonada:
@locker I agree. Implemented your suggestions.

PR #142, opened 2019-02-27
tests: improved pre_cleanup script

avtikhon:
Improved pre cleanup script and made it switched on by default

### DIFF
--- a/lib/test_suite.py
+++ b/lib/test_suite.py
@@ -110,7 +110,7 @@ class TestSuite:
                 dict.fromkeys(self.ini[i].split())
                 if i in self.ini else dict())
 
-        self.parse_bool_opt('pre_cleanup', False)
+        self.parse_bool_opt('pre_cleanup', True)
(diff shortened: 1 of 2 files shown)


lib/test_suite.py
```diff
@@ -110,7 +110,7 @@ def __init__(self, suite_path, args):
                 dict.fromkeys(self.ini[i].split())
                 if i in self.ini else dict())
 
-        self.parse_bool_opt('pre_cleanup', False)
+        self.parse_bool_opt('pre_cleanup', True)
```
Totktonada: To be honest, I don't sure we should enable the option by default in test-run. Say, ok, we decided to use this way in tarantool testing. But test-run is the common harness for testing tarantool-oriented project and I tentative about pushing users to understand and use this extra logic.

Totktonada:
Pushed as 007c6a0c4905f1e91ff19a2734a4582b4e153ff5.

PR #146, opened 2019-03-01
Fix test_run:cmd('set variable ...') for strings

avtikhon:
Before this commit the following command fails when
use_unix_sockets_iproto is set to True:

test_run:cmd("set variable r_uri to 'replica.listen'")

The reason of the fail is that the following Lua statement is formed to
execute the command:

r_uri=/path/to/unix/socket

Found on replication/catch.test.lua from tarantool.

### DIFF
--- a/lib/preprocessor.py
+++ b/lib/preprocessor.py
@@ -5,6 +5,7 @@ from collections import deque
 
 import yaml
 from gevent import socket
+import six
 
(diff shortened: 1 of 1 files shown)


Totktonada:
@avtikhon Updated. Now review me :)

avtikhon:
Seems all changes are fine, please pull.

PR #154, opened 2019-04-05
Update pyyaml version

Totktonada:
GitHub warns about pyyaml < 4.2b1, because of CVE-2017-18342:
yaml.load() could cause arbitrary code execution. I think it is not very
important in the case of test-run, because a test code is controlled by
a user, but anyway it is good to be up to date.

Eliminated new pyyaml warnings about using of unsafe yaml.load():
replaced it with yaml.safe_load(). We don't construct Python classes
directly according to yaml tags, so safe_load() fit our needs.

### DIFF
--- a/lib/admin_connection.py
+++ b/lib/admin_connection.py
@@ -69,7 +69,7 @@ class ExecMixIn(object):
                 break
 
         try:
-            yaml.load(res)
+            yaml.safe_load(res)
         finally:
             if not silent:
                 sys.stdout.write(res.replace("\r\n", "\n"))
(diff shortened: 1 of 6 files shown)


Totktonada:
The main reason for this update for me is described in https://github.com/tarantool/tarantool/commit/d5fdc533990620a8af520dd68f697b26d9513f7c:

> The primary reason why this change is needed is that yaml.load() w/o an
> explicit loader was banned in Gentoo Linux for recent pyyaml versions;
> see [1].
>
> [1]: https://gitweb.gentoo.org/repo/gentoo.git/commit/?id=79ba924d94cb0cf8559565178414c2a1d687b90c

PR #163, opened 2019-04-18
Add wait_upstream() and wait_downstream()

### LINKED ISSUE #158: Add wait_upstream() and wait_downstream() functions
We already have wait_log() that is implemented via wait_cond(). I propose to add the following functions:

* test_run:wait_upstream(id, [status[, message]])
* test_run:wait_downstream(id, [status[, message]])

No-op when both status and message are omitted.

Totktonada:
These functions are needed to wait for a certain state of an upstream /
downstream as tracked in box.info.replication.

Fixes #158.

### DIFF
--- a/test_run.lua
+++ b/test_run.lua
@@ -1,3 +1,5 @@
+--- Utils provided by test-run.
+
 local socket = require('socket')
 local json = require('json')
 local yaml = require('yaml')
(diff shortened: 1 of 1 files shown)


Totktonada:
Now I think about expanding the API to support waiting vclocks. See https://github.com/tarantool/test-run/issues/158#issuecomment-484792009

PR #164, opened 2019-04-24
Add the signal option into stop server routine

avtikhon:
Added server kill routine based on server stop routine
except the SIGKILL is used instead of SIGTERM and server
is killing before connections stopping.

### DIFF
--- a/lib/tarantool_server.py
+++ b/lib/tarantool_server.py
@@ -735,7 +735,11 @@ class TarantoolServer(Server):
     def wait_stop(self):
         self.process.wait()
 
-    def stop(self, silent=True):
+    def stop(self, silent=True, signal=signal.SIGTERM):
+        """ Kill tarantool server using specified signal (SIGTERM by default)
+
+            signal - a number of a signal
+        """
         if self._start_against_running:
             color_log('Server [%s] start against running ...\n',
                       schema='test_var')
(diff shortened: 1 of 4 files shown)


avtikhon:
Stand alone test gc_4162 for replication suite prepared, unzip from the attachments.
[gc_4162.zip](https://github.com/tarantool/test-run/files/3116947/gc_4162.zip)

lib/tarantool_server.py
```diff
@@ -763,7 +763,13 @@ def stop(self, silent=True):
                 schema='test_var'
             )
             try:
-                self.process.terminate()
+                if signal:
+                    color_log('Sending signal {0} to process {1}\n'.format(
+                        signal, self.process.pid))
+                    os.system('/bin/kill --signal {0} {1}'.format(
```
Totktonada: It is better to do it w/o fork+exec using process.send_signal().

lib/tarantool_server.py
```diff
@@ -735,7 +735,7 @@ def kill_current_test(self):
     def wait_stop(self):
         self.process.wait()
 
-    def stop(self, silent=True):
+    def stop(self, silent=True, signal=None):
```
Totktonada: It would be better to receive signal.SIG* constants here instead of strings. We can also set signal.SIGTERM as default. process.send_signal(signal.SIGTERM) is the alias for process.terminate() (see [here]), so this will not change the behaviour, but will shrink the code.

[1]: https://docs.python.org/2/library/subprocess.html#subprocess.Popen.send_signal

Totktonada:
BTW, I would add a usage example (just how to send the command) into a commit message. It would also good to update the README with this option: it lack options at all now, but at least start to list new ones would improve the page.

Totktonada:
Please also ask informal OK from @locker re changing the test case from https://github.com/tarantool/tarantool/issues/4162 . Otherwise this feature can become just dead code.

Totktonada:
Rebased on the current master.

lib/tarantool_server.py
Totktonada: I should have to clarify that it is good to provide ability to give a signal as a string for the test-run's command to provide better support for different platforms. However stop() is the internal function and so should be called from test-run itself with signal.SIG* as an argument value.

Changed the commit accordingly.

Totktonada:
Added support for giving a signal by a name: 'stop server foo with signal=KILL' (see the comment above).

Shrunk the comment to describe only those things that are relevant for test-run.

Fixed the README formatting in a separate commit.

Please, stop name everything 'routine'. There are test-run *commands*, C *functions*, Python *methods*. More concrete words make a sentence easier to understand.

Totktonada:
Updated in tarantool: master, 2.1 and 1.10.

PR #166, opened 2019-05-03
Fix core name (unit-testing) from 'unit' to 'unittest' in README

SudoBobo:
Actually there is 'unittest' core name, not 'unit'. Commit fixes README.


Totktonada:
Changed the commit message (a header should be within 50 symbols) and pushed in e48b0a5d7335ea97bda9c2dabeedc28bb9333021.

PR #173, opened 2019-05-28
Allow to start multiple tarantool instances from single script.

rtokarev:
This patch is allowed to start multiple tarantool instances from a single script. Each instance will have a unique work directory, log, pidfile, control socket and etc, named with the supplied instance name.

The instance name will not be overridden with an initial script name. For each instance, the initial script will be copied to the var directory and will be named with the instance name.

### DIFF
--- a/lib/tarantool_server.py
+++ b/lib/tarantool_server.py
@@ -294,7 +294,7 @@ class TarantoolServer(Server):
 
     @property
     def script_dst(self):
-        return os.path.join(self.vardir, os.path.basename(self.script))
+        return os.path.join(self.vardir, self.name + '.lua')
 
     @property
     def logfile_pos(self):
@@ -319,7 +319,8 @@ class TarantoolServer(Server):
                 delattr(self, '_script')
             return
         self._script = os.path.abspath(val)
-        self.name = os.path.basename(self._script).split('.')[0]
+	if self.name == 'default' :
+	    self.name = os.path.basename(self._script).split('.')[0]
 
     @property
     def _admin(self):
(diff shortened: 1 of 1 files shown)


Totktonada:
Sorry, I had those thoughts for a long time in the mind, but something constantly take me away from formulating them.

What disquiets me here. We have three things:

1. An instance name in test-run.
2. A script name.
3. An instance name in tarantoolctl (xlogs, snaps, text logs are named similarly).

Now 2 == 3. It is not quite obvious, how everything works: where to find logs and so on. However most developers work with that for a long time and more or less know where to find necessary things.

After the change, 1 == 3. It is not bad per se, if we would do this from scratch. However I afraid that the change may confuse developers more than how it is implemented now. Just because one non-quite-obvious behaviour will be replaced with another one that is of this kind too.

My objections are not quite strict, though.

I have an idea in the mind that likely will allow us to get rid of all those extra files. Something like:

```lua
test_run:create_server(<name>, {
    <..other options are here..>
    script = [[
        box.cfg{
            listen = os.getenv('LISTEN'),
        }
    ]],
})
test_run:start_server(<name>, {<startup options>})
```

The one `name` for test-run and tarantoolctl.

We can pass instance name as well as user defined data using environment variables.

Is not that nice?

Totktonada:
However after the change the script is renamed according to test-run's instance name in the working directory. So, maybe, everything is actually good: one may look into the working directory and understand how instance files are relevant to logs, xlogs, snaps and so on (as well as now). In fact, we discuss four things, not three ones:

1. An instance name in test-run.
2. A script name in the source directory (`test/foo/`).
3. A script name in the working directory (`test/var/001_foo/`).
4. An instance name in tarantoolctl (xlogs, snaps, text logs are named similarly).

Now 2 == 3 == 4.

After the change it'll be 1 == 3 == 4.

PR #177, opened 2019-07-01
Implement SQL driver

Totktonada:
Actually three features are introduced with this commit.

## SQL driver

Now test-run recognizes *.test.sql files.

The content of those files are SQL statements written line-by-line. A
long statement may be splitted, see the next section.

The implementation leans on '\set language [lua|sql]' command of
tarantool console: appropriate language is set depending of a file name:
*.test.lua or *.test.sql before executing commands for a test file.

A value of 'engine' option when it is provided by a configuration has a
special meaning for *.test.sql tests (a configuration is a file set by
'config' option in suite.ini) . If it is 'memtx' or 'vinyl', then
corresponding default engine will be set for SQL subsystem before
executing commands for a test file. The implementation uses "pragma
sql_default_engine='memtx|vinyl'" command.

It is possible to mix Lua and SQL code in *.test.lua and *.test.sql
tests. Consider an example:

```
-- Verify output of large integers.
CREATE TABLE test (id INTEGER PRIMARY KEY)
INSERT INTO test VALUES (9223372036854775807)
SELECT * FROM test
\set language lua
box.space.TEST:select()
\set language sql
DROP TABLE test
```

## Line carrying with backslash

Consider the following code:

```
test_run:cmd("setopt delimiter ';'")
function echo(...)
    return ...
end
test_run:cmd("setopt delimiter ''");
```

Now it may be rewritten in the following way:

```
function echo(...) \
    return ...     \
end
```

This ability also works in SQL test files: a long statement may be
written in this way:

```
CREATE TABLE t1 (s1 VARCHAR(10) PRIMARY KEY)
CREATE TABLE t2 (s1 VARCHAR(10) PRIMARY KEY, s2 VARCHAR(10) \
    REFERENCES t1 ON DELETE CASCADE)
```

## New result file format

A new result file format is introduced. The differences are following.

1. A result of a Lua command or SQL statement is indented to easier
   distinguish from a command / statement itself.

Before:

```
CREATE TABLE t1 (s1 VARCHAR(10) PRIMARY KEY)
---
- row_count: 1
...
```

After:

```
CREATE TABLE t1 (s1 VARCHAR(10) PRIMARY KEY)
 | ---
 | - row_count: 1
 | ...
```

2. Empty lines from a test file are preserved in a result file.

3. A small bug when a comment in a setopt-delimiter block is written
   before a previous command was fixed.

4. The first line of a result file in the new format is a version line.

```
-- test-run result file version 2
```

A result file for a new test will be written in the new format, but when
a result file already exists (and has the old format), test-run performs
formatting of a test output in the old way. So the new test-run version
is backward compatible with previous ones and no changes are needed for
existing test and result files.

If you want to update a result file to the new format, remove it and run
a test.

If you want to produce a result file in the old format for a new test,
create an empty result file, run a test and move *.reject to *.result.

Fixes https://github.com/tarantool/tarantool/issues/4123

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,9 +30,9 @@ Field `core` must be one of:
 
 ### Test
 
-Each test consists of files `*.test(.lua|.py)?`, `*.result`, and may have skip
-condition file `*.skipcond`.  On first run (without `.result`) `.result` is
-generated from output.  Each run, in the beggining, `.skipcond` file is
+Each test consists of files `*.test(.lua|.sql|.py)?`, `*.result`, and may have
+skip condition file `*.skipcond`.  On first run (without `.result`) `.result`
+is generated from output.  Each run, in the beggining, `.skipcond` file is
 executed. In the local env there's object `self`, that's `Test` object. If test
 must be skipped - you must put `self.skip = 1` in this file. Next,
 `.test(.lua|.py)?` is executed and file `.reject` is created, then `.reject` is
@@ -67,6 +67,11 @@ engine = test_run:get_cfg('engine')
 -- first run engine is 'memtx'
 -- second run engine is 'sophia'
 ```
+
+"engine" value has a special meaning for *.test.sql files: if it is "memtx" or
+"vinyl", then the corresponding engine will be set using "pragma
+sql_default_engine='memtx|vinyl'" command before executing commands from a test
+file.
 
 #### Python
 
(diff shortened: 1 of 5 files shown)


Totktonada:
I have discussed this patch with my team and decided to perform several follow-ups afterwards.

Backslash is convenient for carrying a long command, but is less convenient for code blocks. We have \setopt delimiter in test-run, which is lua-specific (works via test_run lua module) and is not compatible with \set delimiter syntax from tarantool console. Let's support console syntax for this: https://github.com/tarantool/test-run/issues/178

It is often convenient to copy-paste a code hunk from a test to tarantool console, so we need to support backslashes in console too: https://github.com/tarantool/tarantool/issues/4317

Issue #157, opened 2019-04-10
Kill hung tarantool with SIGKILL if SIGTERM is not sufficient


Totktonada:
Maybe it worth to try prlimit --core=unlimited and send SIGQUIT to try to collect a core file.

Totktonada:
A partial solution is implemented in the PR #186. See also https://github.com/tarantool/tarantool/commit/24f57a35304a31bfbdf7519f3a304e04daddd4ee.

PR #186, opened 2019-08-19
Add the 'timeout' option into 'stop server' command and kill a server…

rtokarev:
… with

 'SIGKILL' if it doesn't finished before timeout is expired..

By default drop_cluster() routine uses SIGTERM signal to stop the
replications. Found that in some situations SIGTERM couldn't kill
all instances on OSX and some processes left. To avoid of such
situations need additionally to send SIGKILL signal to all instances
that were not finished before timeout was expired to be able to stop
them all.

### DIFF
--- a/lib/colorer.py
+++ b/lib/colorer.py
@@ -20,6 +20,33 @@ def color_log(*args, **kwargs):
     color_stdout(*args, **kwargs)
 
 
+def qa_notice(*args, **kwargs):
+    """ Print a notice for an QA engineer at the terminal.
+
+        Example::
+
+            * [QA Notice]
+            *
+            * Attempt to stop already stopped server 'foo'
+            *
+    """
+    # Import from the function to avoid recursive import.
+    from lib.utils import prefix_each_line
+
+    # Use 'info' color by default (yellow).
+    if 'schema' not in kwargs:
+        kwargs = dict(kwargs, schema='info')
+
+    # Join all positional arguments (like color_stdout() do) and
+    # decorate with a header and asterisks.
+    data = ''.join([str(msg) for msg in args])
+    data = prefix_each_line('* ', data)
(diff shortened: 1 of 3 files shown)


avtikhon:
PR successfully checked on multi runs in CI:
https://gitlab.com/tarantool/tarantool/-/pipelines?page=1&scope=all&ref=avtikhon%2Fflaky_stable_min_prs
LGTM
Also it better resolves issue with dropping cluster than PR #232 
and fixes for OSX the the same issue that PR # 242 fixes for Linux only.

Totktonada:
Part of #157 (solves the problem for 'core = tarantool' test suites).

Totktonada:
This PR would be helpful for us to overcome https://github.com/tarantool/tarantool/issues/5573, that hit us in testing from time to time.

Totktonada:
Updated the patch:

* Rebased on the latest master.
* Added a patch with the `qa_notice()` function that I want to use here.
* Issue a QA notice in case of a stuck tarantool.
* Define the `kill()` function within the `stop()` one, because the warning becomes specific for the situation in `stop()`.
* Removed `timeout` parameter from `stop()`, because it is not used anywhere for now.
* Written more about motivation in the commit message and added myself to co-authors.

Totktonada (approved):

avtikhon:
The patch successfully checked at [1], LGTM:

[1] - https://gitlab.com/tarantool/tarantool/-/pipelines/222917874

avtikhon (approved):

LeonidVas (approved):
LGTM.

Totktonada:
Updated the test-run submodule in tarantool in the following commits: 2.7.0-83-g24f57a353, 2.6.1-70-g5e660f7cd, 2.5.2-44-g85eb5eea6, 1.10.8-31-g1e01031e9.

PR #191, opened 2019-10-27
Add --exclude option

### LINKED ISSUE #54: Add option --exclude
It is useful if the test-run would be able to skip some tests specified via option `exclude`

Totktonada:
The option allows to avoid execution of a certain subset of tests.

The option sets a pattern for exclusion and all tests are filtered using
it: if a full test name (say, app-tap/string.test.lua) contains the
pattern as a substring, then the test will be excluded from execution.

The option can be passed several times to set several exclusions
patterns.

A list of such exclusion patterns also can be provided using
TEST_RUN_EXCLUDE environment variable (patterns are separated by
whitespaces).

If both the environment variable and the option are provided, then the
environment variable value will be ignored.

Trick: `--exclude foo/` allows to exclude a test suite.

Fixes #54.

### DIFF
--- a/lib/app_server.py
+++ b/lib/app_server.py
@@ -140,6 +140,7 @@ class AppServer(Server):
         test_suite.ini['suite'] = suite_path
 
         test_names = sorted(glob.glob(os.path.join(suite_path, "*.test.lua")))
+        test_names = Server.exclude_tests(test_names, test_suite.args.exclude)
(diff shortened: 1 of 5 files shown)


Totktonada:
One idea just appear in my head: we can print the `[ excluded ]` status for excluded tests rather then filter them out during tests collecting. Anyway, it is quite minor thing.

PR #210, opened 2020-05-13
Update 'use_unix_sockets_iproto' option for core = app tests

avtikhon:
Update 'use_unix_sockets_iproto' for 'core = app'
    
Updated 'use_unix_sockets_iproto' option to support 'core = app'
tests, which enabled unix sockets for iproto connections. It helped
to handle the problem with 'Address already in use' error.
    
Part of issue https://github.com/tarantool/tarantool/issues/4459

### DIFF
--- a/lib/app_server.py
+++ b/lib/app_server.py
@@ -14,6 +14,7 @@ from lib.tarantool_server import TarantoolServer
 from lib.tarantool_server import TarantoolStartError
 from lib.utils import find_port
 from lib.utils import format_process
+from lib.utils import warn_unix_socket
 from test import TestRunGreenlet, TestExecutionError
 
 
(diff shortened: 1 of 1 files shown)


lib/app_server.py
```diff
@@ -103,7 +105,12 @@ def deploy(self, vardir=None, silent=True, need_init=True):
                     if (e.errno == errno.ENOENT):
                         continue
                     raise
-        os.putenv("LISTEN", str(find_port()))
+        if self.use_unix_sockets_iproto:
+            path = os.path.join(self.vardir, self.name + ".socket-iproto")
+            warn_unix_socket(path)
+            os.putenv("LISTEN", str(path))
```
Totktonada, line 111: Nit: It is already string, `str()` is not needed.

avtikhon: All corrected as you suggested.

Issue #141, opened 2019-02-18
Fix 'Address already in use' error with TCP sockets


Totktonada:
TcpPortDispatcher was introduced to eliminate the race condition when two workers trying to use the same port: the idea is that each worker use its own ports range. Really these ports could race with client ports (from, say, net.box or replication), which typically don't use bind() and so binds to a random available port (despite any dispatched ranges). So this approach is broken by design.

There is the workaround with `use_unix_sockets` and `use_unix_sockets_iproto` options: setting them to `True` should eliminate the problem. However it is possible that some tests are designed to test tcp connections, so I think we should fix the problem.

The way I see is no-bind / bind to zero port and grep logs for a LISTEN / CONSOLE port (or write it from a script to a predefined location). Remove TcpPortDispatcher and find_port() at all.

Hmm, `box.cfg{listen = 0}` shows `binary: bound to 0.0.0.0:0`. Need to elaborate how to get the port.

The discussion had initially arisen in #115.

Totktonada:
See https://github.com/tarantool/tarantool/issues/4620

Totktonada:
Related: PR #210.

sergos:
Hmm.. strange, for me it looks like `tarantool> box.cfg{listen=0}` yields `2022-02-16 01:07:32.744 [19216] main/103/interactive I> tx_binary: bound to [::]:53061` on MacOS and `2022-02-16 01:06:33.753 [10090] main/103/interactive I> tx_binary: bound to 0.0.0.0:36267` on Linux (Ubuntu 18.04)
The problem might be in the difference:
```
tarantool> box.cfg.listen
---
- 0
...

tarantool> box.info.listen
---
- '[::]:53061'
...

```

Anyways, letting the Tarantool instance to pick a port by itself looks way more robust, since no delay between assigning a port and binding it.

Totktonada:
@sergos It is because https://github.com/tarantool/tarantool/issues/4620 was fixed. I filed this issue against tarantool specifically to have a change to use it here, in test-run.

Totktonada:
Regarding 1.10, where we have no `box.info.listen`: https://github.com/tarantool/tt/pull/112#issuecomment-1209661575.

NickZay:
I have some problems with ports when I use luatest with tarantool

PR #217, opened 2020-07-15
Enable test reruns on failed fragiled tests

### LINKED ISSUE #189: Retry fragile tests or allow them to fail
Manual retrying of testing to understand whether a fail is persistent spends my time, it is undesirable.

I have two ways to overcome this, both are on 'raw idea' rights.

### Provide a way to set a retrying count for fragile tests

It could be an option in suite.ini. If one of those attempts is successful, then the test should be considered as passed.

We also need an ability to retry a hang test.

I was against retrying for a long time and had hope that we'll fix all flaky tests in some future. It seems we really unable to achieve this goal.

### Allow fragile tests to fail

Maybe it worth to do that under command line option and use it in CI, but not locally.

avtikhon:
Added ability to use fragile list in JSON format from 'suite.ini'
files.
```
   fragile = {
        "retries": 10,
        "tests": {
            "bitset.test.lua": {
                "issues": [ "gh-4095" ],
                "checksums": [ "050af3a99561a724013995668a4bc71c", "f34be60193cfe9221d3fe50df657e9d3" ]
            }
        }}
```

Added ability to set per suite in suite.ini configuration file
'retries' option, which sets the number of accepted reruns of
the tests failed from 'fragile' list.

Added ability to check failed tests w/ fragile list to be sure that
the current fail equal to the issue mentioned in the fragile list.

Closes #189.

### DIFF
--- a/lib/test.py
+++ b/lib/test.py
@@ -8,6 +8,7 @@ import shutil
 import sys
 import traceback
 from functools import partial
+from hashlib import md5
 
 try:
     from cStringIO import StringIO
@@ -152,9 +153,9 @@ class Test(object):
             it to stdout.
 
             Returns short status of the test as a string: 'skip', 'pass',
-            'new', 'updated' or 'fail'. There is also one possible value for
-            short_status, 'disabled', but it returned in the caller,
-            TestSuite.run_test().
+            'new', 'updated' or 'fail' and results file checksum on fail.
+            There is also one possible value for short_status, 'disabled',
+            but it returned in the caller, TestSuite.run_test().
         """
 
         # Note: test was created before certain worker become known, so we need
@@ -219,6 +220,7 @@ class Test(object):
             self.is_valgrind_clean = not bool(non_empty_logs)
 
         short_status = None
+        result_checksum = None
 
         if self.skip:
             short_status = 'skip'
@@ -252,6 +254,8 @@ class Test(object):
             has_result = os.path.exists(self.tmp_result)
             if has_result:
                 shutil.copy(self.tmp_result, self.reject)
+                with open(self.tmp_result, mode='rb') as result_file:
+                    result_checksum = md5(result_file.read()).hexdigest()
             short_status = 'fail'
             color_stdout("[ fail ]\n", schema='test_fail')
 
@@ -277,7 +281,7 @@ class Test(object):
                                            "Test failed! Output from log file "
                                            "{0}:\n".format(log_file))
                 where = ": there were warnings in the valgrind log file(s)"
-        return short_status
+        return short_status, result_checksum
 
     def print_diagnostics(self, log_file, message):
         """Print whole lines of client program output leading to test
--- a/lib/test_suite.py
+++ b/lib/test_suite.py
@@ -81,7 +81,7 @@ class TestSuite:
         self.args = args
         self.tests = []
         self.ini = {}
-        self.fragile = {'tests': {}}
+        self.fragile = {'retries': 0, 'tests': {}}
         self.suite_path = suite_path
         self.ini["core"] = "tarantool"
 
@@ -264,6 +264,9 @@ class TestSuite:
 
     def is_parallel(self):
         return self.ini['is_parallel']
+
+    def fragile_retries(self):
+        return self.fragile.get('retries', 0)
 
     def show_reproduce_content(self):
         return self.ini['show_reproduce_content']
--- a/lib/test_suite.py
+++ b/lib/test_suite.py
@@ -185,6 +185,12 @@ class TestSuite:
                 res.append(test)
         return res
 
+    def get_test_fragile_checksums(self, test):
+        try:
+            return self.fragile['tests'][test]['checksums']
+        except Exception:
+            return []
+
     def gen_server(self):
         try:
             return Server(self.ini, test_suite=self)
@@ -237,7 +243,7 @@ class TestSuite:
 
     def run_test(self, test, server, inspector):
         """ Returns short status of the test as a string: 'skip', 'pass',
-            'new', 'fail', or 'disabled'.
+            'new', 'fail', or 'disabled' and results file checksum on fail.
         """
         test.inspector = inspector
         test_name = os.path.basename(test.name)
@@ -251,16 +257,17 @@ class TestSuite:
         color_stdout(conf.ljust(16), schema='test_var')
 
         if self.is_test_enabled(test, conf, server):
-            short_status = test.run(server)
+            short_status, result_checksum = test.run(server)
         else:
             color_stdout("[ disabled ]\n", schema='t_name')
             short_status = 'disabled'
+            result_checksum = None
 
         # cleanup only if test passed or if --force mode enabled
         if lib.Options().args.is_force or short_status == 'pass':
             inspector.cleanup_nondefault()
 
-        return short_status
+        return short_status, result_checksum
 
     def is_parallel(self):
         return self.ini['is_parallel']
--- a/lib/test_suite.py
+++ b/lib/test_suite.py
@@ -81,6 +81,7 @@ class TestSuite:
         self.args = args
         self.tests = []
         self.ini = {}
+        self.fragile = {'tests': {}}
         self.suite_path = suite_path
         self.ini["core"] = "tarantool"
 
@@ -110,6 +111,17 @@ class TestSuite:
                 lambda x: os.path.join(suite_path, x),
                 dict.fromkeys(self.ini[i].split())
                 if i in self.ini else dict())
+        if config.has_option("default", "fragile"):
+            fragiles = config.get("default", "fragile")
+            try:
+                self.fragile = json.loads(fragiles)
+                if 'tests' not in self.fragile:
+                    raise RuntimeError(
+                        "Key 'tests' absent in 'fragile' json: {}"
+                        . format(self.fragile))
+            except ValueError:
+                # use old format dictionary
+                self.fragile['tests'] = self.ini['fragile']
 
         self.parse_bool_opt('pretest_clean', False)
         self.parse_bool_opt('use_unix_sockets', False)
@@ -154,11 +166,14 @@ class TestSuite:
         self.tests_are_collected = True
         return self.tests
 
+    def get_fragile_list(self):
+        return self.fragile['tests'].keys()
+
     def stable_tests(self):
         self.collect_tests()
         res = []
         for test in self.tests:
-            if os.path.basename(test.name) not in self.ini['fragile']:
+            if os.path.basename(test.name) not in self.get_fragile_list():
                 res.append(test)
         return res
 
@@ -166,7 +181,7 @@ class TestSuite:
         self.collect_tests()
         res = []
         for test in self.tests:
-            if os.path.basename(test.name) in self.ini['fragile']:
+            if os.path.basename(test.name) in self.get_fragile_list():
                 res.append(test)
         return res
 
--- a/lib/worker.py
+++ b/lib/worker.py
@@ -142,13 +142,15 @@ class WorkerTaskResult(BaseWorkerMessage):
     """ Passed into the result queue when a task processed (done) by the
     worker. The short_status (string) field intended to give short note whether
     the task processed successfully or not, but with little more flexibility
-    than binary True/False. The task_id (any hashable object) field hold ID of
+    than binary True/False. The result_checksum (string) field saves the results
+    file checksum on test fail. The task_id (any hashable object) field hold ID of
     the processed task. The show_reproduce_content configuration form suite.ini
     """
     def __init__(self, worker_id, worker_name, task_id,
-                 short_status, show_reproduce_content):
+                 short_status, result_checksum, show_reproduce_content):
         super(WorkerTaskResult, self).__init__(worker_id, worker_name)
         self.short_status = short_status
+        self.result_checksum = result_checksum
         self.task_id = task_id
         self.show_reproduce_content = show_reproduce_content
 
@@ -214,8 +216,9 @@ class Worker:
         return WorkerCurrentTask(self.id, self.name, task_name, task_param,
                                  task_result, task_tmp_result)
 
-    def wrap_result(self, task_id, short_status):
+    def wrap_result(self, task_id, short_status, result_checksum):
         return WorkerTaskResult(self.id, self.name, task_id, short_status,
+                                result_checksum,
                                 self.suite.show_reproduce_content())
 
     def sigterm_handler(self, signum, frame):
@@ -302,7 +305,7 @@ class Worker:
             with open(self.reproduce_file, 'a') as f:
                 task_id_str = yaml.safe_dump(task.id, default_flow_style=True)
                 f.write('- ' + task_id_str)
-            short_status = self.suite.run_test(
+            short_status, result_checksum = self.suite.run_test(
                 task, self.server, self.inspector)
         except KeyboardInterrupt:
             self.report_keyboard_interrupt()
@@ -312,7 +315,7 @@ class Worker:
                 '\nWorker "%s" received the following error; stopping...\n'
                 % self.name + traceback.format_exc() + '\n', schema='error')
             raise
-        return short_status
+        return short_status, result_checksum
 
     def run_loop(self, task_queue, result_queue):
         """ called from 'run_all' """
@@ -327,7 +330,10 @@ class Worker:
                 break
 
             short_status = None
+            result_checksum = None
             result_queue.put(self.current_task(task_id))
+            testname = os.path.basename(task_id[0])
+            fragile_checksums = self.suite.get_test_fragile_checksums(testname)
             retries_left = self.suite.fragile_retries()
             # let's run till short_status became 'pass'
             while short_status != 'pass' and retries_left >= 0:
@@ -335,13 +341,18 @@ class Worker:
                 if short_status == 'fail':
                     color_stdout(
                         'Test "%s", conf: "%s"\n'
-                        '\tfrom "fragile" list failed, rerunning ...\n'
-                        % (task_id[0], task_id[1]), schema='error')
+                        '\tfrom "fragile" list failed with results'
+                        ' file checksum: "%s", rerunning ...\n'
+                        % (task_id[0], task_id[1], result_checksum), schema='error')
                 # run task and save the result to short_status
-                short_status = self.run_task(task_id)
+                short_status, result_checksum = self.run_task(task_id)
+                # check if the results file checksum set on fail and if
+                # the newly created results file is known by checksum
+                if not result_checksum or (result_checksum not in fragile_checksums):
+                    break
                 retries_left = retries_left - 1
 
-            result_queue.put(self.wrap_result(task_id, short_status))
+            result_queue.put(self.wrap_result(task_id, short_status, result_checksum))
             if not lib.Options().args.is_force and short_status == 'fail':
                 color_stdout(
                     'Worker "%s" got failed test; stopping the server...\n'
--- a/lib/worker.py
+++ b/lib/worker.py
@@ -326,8 +326,21 @@ class Worker:
                 self.stop_worker(task_queue, result_queue)
                 break
 
+            short_status = None
             result_queue.put(self.current_task(task_id))
-            short_status = self.run_task(task_id)
+            retries_left = self.suite.fragile_retries()
+            # let's run till short_status became 'pass'
+            while short_status != 'pass' and retries_left >= 0:
+                # print message only after some fails occurred
+                if short_status == 'fail':
+                    color_stdout(
+                        'Test "%s", conf: "%s"\n'
+                        '\tfrom "fragile" list failed, rerunning ...\n'
+                        % (task_id[0], task_id[1]), schema='error')
+                # run task and save the result to short_status
+                short_status = self.run_task(task_id)
+                retries_left = retries_left - 1
+
             result_queue.put(self.wrap_result(task_id, short_status))
             if not lib.Options().args.is_force and short_status == 'fail':
                 color_stdout(
--- a/listeners.py
+++ b/listeners.py
@@ -45,6 +45,7 @@ class StatisticsWatcher(BaseWatcher):
         if obj.short_status == 'fail':
             self.failed_tasks.append((obj.task_id,
                                       obj.worker_name,
+                                      obj.result_checksum,
                                       obj.show_reproduce_content))
 
     def print_statistics(self):
@@ -58,10 +59,11 @@ class StatisticsWatcher(BaseWatcher):
             return False
 
         color_stdout('Failed tasks:\n', schema='test_var')
-        for task_id, worker_name, show_reproduce_content in self.failed_tasks:
+        for task_id, worker_name, result_checksum, show_reproduce_content in self.failed_tasks:
             logfile = self.get_logfile(worker_name)
             task_id_str = yaml.safe_dump(task_id, default_flow_style=True)
             color_stdout('- %s' % task_id_str, schema='test_var')
+            color_stdout('# results file checksum: %s\n' % result_checksum)
             color_stdout('# logfile:        %s\n' % logfile)
             reproduce_file_path = get_reproduce_file(worker_name)
             color_stdout('# reproduce file: %s\n' % reproduce_file_path)


lib/worker.py
```diff
@@ -313,7 +315,7 @@ def run_task(self, task_id):
             raise
         return short_status
 
-    def run_loop(self, task_queue, result_queue):
+    def run_loop(self, task_queue, result_queue, fragile_retries):
```
LeonidVas: fragile_retries is saved in self.suite.ini, may be forward all fragile stuff through arguments is not a good idea, may be just pass the is_fragile flag?

lib/worker.py
```diff
@@ -326,7 +328,19 @@ def run_loop(self, task_queue, result_queue):
                 break
 
             result_queue.put(self.current_task(task_id))
-            short_status = self.run_task(task_id)
+            short_status = ''
+            # 'fragile_retries' value comes from the suite configuration
+            retries_left = fragile_retries
+            while short_status != 'pass' and retries_left >= 0:
```
LeonidVas, line 337: Must be used only for "fragile" tests. IMHO, retries_left > 0 looks more correct.
May be something like:
run_test()
if no pass and fragile then retry...

lib/worker.py
```diff
@@ -340,14 +354,14 @@ def run_loop(self, task_queue, result_queue):
                 raise VoluntaryStopException()
             self.task_done(task_queue)
 
-    def run_all(self, task_queue, result_queue):
+    def run_all(self, task_queue, result_queue, fragile_retries):
```
LeonidVas: The same as for run_loop() method.

lib/test_suite.py
```diff
@@ -174,6 +174,14 @@ def fragile_tests(self):
                 res.append(test)
         return res
 
+    def fragile_checksums(self):
```
LeonidVas, line 177: Maybe fragile_checksums()  should return a dict {test_name_1 -> [checksum1, checksum2],
test_name_2 -> [checksum1]}

lib/worker.py
```diff
@@ -83,6 +84,7 @@ def get_task_groups():
                 'task_ids': stable_task_ids,
                 'is_parallel': suite.is_parallel(),
                 'fragile_retries': 0,
+                'fragile_checksums': suite.fragile_checksums(),
```
LeonidVas, line 87: As I proposed, just add the "is_fragile" flag.

LeonidVas:
It looks like this closes # 189. Add this information to the commit message instead of "Part of tarantool / tarantool problem # 5050".

dispatcher.py
```diff
@@ -355,6 +355,7 @@ def __init__(self, key, task_group, randomize):
         self.gen_worker = task_group['gen_worker']
         self.task_ids = task_group['task_ids']
         self.is_parallel = task_group['is_parallel']
+        self.fragile_retries = int(task_group['fragile_retries'])
```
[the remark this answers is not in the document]
avtikhon: Right, check is not needed here.

lib/worker.py
avtikhon: Right, removed fragile_retries and used self.suite.fragile_retries() where was needed. Set is_fragile here.

lib/test_suite.py
```diff
@@ -117,6 +117,10 @@ def __init__(self, suite_path, args):
         self.parse_bool_opt('is_parallel', False)
         self.parse_bool_opt('show_reproduce_content', True)
 
+        fragile_retries = self.ini.get('fragile_retries')
```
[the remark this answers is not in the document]
avtikhon: Created self.parse_int_opt() method.

lib/worker.py
avtikhon: Split into 2 parts for fragile and not fragile.

lib/worker.py
avtikhon: Right, corrected.

lib/worker.py
```diff
@@ -315,7 +318,7 @@ def run_task(self, task_id):
             raise
         return short_status
 
-    def run_loop(self, task_queue, result_queue, fragile_retries):
+    def run_loop(self, task_queue, result_queue, fragile_retries, fragile_checksums):
```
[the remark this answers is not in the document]
avtikhon: Right, removed.

lib/worker.py
avtikhon: Right, added flag is_fragile.

avtikhon:
> It looks like this closes # 189. Add this information to the commit message instead of "Part of tarantool / tarantool problem # 5050".

Ok, changed.

LeonidVas:
After some discussion with @Totktonada and @avtikhon, it was decided to try using JSON for configuration a "fragile".
Example:
```
fragile =
 {
   "tests": {
     "test_name": {
       "checksums": [
         ch1,
         ch2
       ]
     }
   },
   "retries": 10
 }
```
Useful links:
https://docs.python.org/3/library/configparser.html#supported-ini-file-structure
https://tools.ietf.org/html/rfc7159

lib/test_suite.py
avtikhon: Sure, added ability to use JSON format.

avtikhon:
> After some discussion with @Totktonada and @avtikhon, it was decided to try using JSON for configuration a "fragile".
> Example:
> 
> ```
> fragile =
>  {
>    "tests": {
>      "test_name": {
>        "checksums": [
>          ch1,
>          ch2
>        ]
>      }
>    },
>    "retries": 10
>  }
> ```
> 
> Useful links:
> https://docs.python.org/3/library/configparser.html#supported-ini-file-structure
> https://tools.ietf.org/html/rfc7159

Implemented.

Totktonada:
I didn't look thoroughly, but I would highlight several points.

1. The approach with checksums is not universal. It does not work if time-related information appears in a result file. It requires periodical checksums updating if a line number is reported as part of an error. I guess it is okay to operate on a test level and either mark it as stable or as fragile.
2. We can either implement reruns within a worker or on the level upward: within the task management system (dispathers-listeners-workers). Since there is nothing tarantool-specific I would look into the latter way. I pushed a variant to `Totktonada/rerun-fragile-test` branch (0292fe2ad901f33c903387f5498ade4f13acef94).

----

If you want my opinion about extracting checksums from comments — it is disgusting. Sorry.

There are more points I would highlight about the patchset, but it does not have much sense if we'll get rid of checksums at all. The patchset modifies many different parts of test-run: too much for this small feature, IMHO.

lib/test_suite.py
```diff
@@ -110,13 +122,23 @@ def __init__(self, suite_path, args):
...
+                    print "WARNING: suite '{}' uses old fragile format" . format(suite_path)
+                    pass
 
         self.parse_bool_opt('pretest_clean', False)
         self.parse_bool_opt('use_unix_sockets', False)
         self.parse_bool_opt('use_unix_sockets_iproto', False)
         self.parse_bool_opt('is_parallel', False)
         self.parse_bool_opt('show_reproduce_content', True)
 
+        self.parse_int_opt('fragile_retries', 0)
```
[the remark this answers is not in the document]
avtikhon: Removed.

lib/test_suite.py
```diff
@@ -154,22 +176,44 @@ def collect_tests(self):
         self.tests_are_collected = True
         return self.tests
 
+    def get_fragile_list(self):
+        if self.ini['fragile_is_json']:
```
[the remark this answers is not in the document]
avtikhon: Removed.

lib/test_suite.py
```diff
@@ -154,22 +176,44 @@ def collect_tests(self):
...
         res = []
         for test in self.tests:
-            if os.path.basename(test.name) in self.ini['fragile']:
+            if os.path.basename(test.name) in self.get_fragile_list():
                 res.append(test)
         return res
 
+    def get_fragile_checksums(self):
+        checksums = []
+        if self.ini['fragile_is_json']:
```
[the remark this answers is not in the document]
avtikhon: Removed.

lib/test_suite.py
```diff
@@ -250,5 +294,14 @@ def run_test(self, test, server, inspector):
     def is_parallel(self):
         return self.ini['is_parallel']
 
+    def fragile_retries(self):
+        if self.ini['fragile_is_json']:
```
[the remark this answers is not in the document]
avtikhon: Removed.

lib/worker.py
```diff
@@ -326,9 +330,30 @@ def run_loop(self, task_queue, result_queue):
                 break
 
             result_queue.put(self.current_task(task_id))
-            short_status = self.run_task(task_id)
+            if is_fragile:
```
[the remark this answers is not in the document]
avtikhon: No, this check only to choose the way of running the tests - from fragile list or not.

lib/test.py
```diff
@@ -250,9 +251,12 @@ def run(self, server):
             color_stdout("[ updated ]\n", schema='test_new')
         else:
             has_result = os.path.exists(self.tmp_result)
+            result_checksum = ''
             if has_result:
                 shutil.copy(self.tmp_result, self.reject)
-            short_status = 'fail'
+                with open(self.tmp_result, mode='rb') as result_file:
+                    result_checksum = md5(result_file.read()).hexdigest()
+            short_status = 'fail:' + result_checksum
```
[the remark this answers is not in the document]
avtikhon: Added.

lib/worker.py
```diff
@@ -82,13 +83,15 @@ def get_task_groups():
                 'gen_worker': gen_worker,
                 'task_ids': stable_task_ids,
                 'is_parallel': suite.is_parallel(),
+                'is_fragile': False,
```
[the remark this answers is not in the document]
avtikhon: Seems not needed for now to use additionally enum, code w/o it looks more readable as I think.

lib/worker.py
```diff
@@ -82,13 +83,15 @@ def get_task_groups():
...
                 'show_reproduce_content': suite.show_reproduce_content(),
             }
         if fragile_task_ids:
             res[key + '_fragile'] = {
                 'gen_worker': gen_worker,
                 'task_ids': fragile_task_ids,
                 'is_parallel': False,
+                'is_fragile': True,
```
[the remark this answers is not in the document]
avtikhon: See comment above.

dispatcher.py
```diff
@@ -361,6 +361,7 @@ def __init__(self, key, task_group, randomize):
                 random.shuffle(self.task_ids)
         else:
             self.randomize = False
+        self.is_fragile = task_group['is_fragile']
```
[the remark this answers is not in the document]
avtikhon: See comment above.

lib/test_suite.py
```diff
@@ -154,22 +176,44 @@ def collect_tests(self):
...
 
+    def get_fragile_checksums(self):
+        checksums = []
+        if self.ini['fragile_is_json']:
+            for test in self.get_fragile_list():
+                try:
+                    for checksum in self.ini['fragile']['tests'][test]['checksums']:
+                        checksums.append(checksum)
+                except Exception:
+                    pass
```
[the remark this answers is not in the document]
avtikhon: Reworked a bit this part, please check if the comment still needed.

avtikhon:
Hi, Alexander, please check my comments below.

> 1. The approach with checksums is not universal. It does not work if time-related information appears in a result file. It requires periodical checksums updating if a line number is reported as part of an error. I guess it is okay to operate on a test level and either mark it as stable or as fragile.

Right, some tests that have commands in format like "<failing command> or <print overall information>" and etc. would not pass using this feature, but in all other ways it must help, due to our results files should be the same between runs. It means that this feature is not absolute medicine for flaky tests, but it can help to avoid of it in some places.

> 2. We can either implement reruns within a worker or on the level upward: within the task management system (dispathers-listeners-workers). Since there is nothing tarantool-specific I would look into the latter way. I pushed a variant to `Totktonada/rerun-fragile-test` branch ([0292fe2](https://github.com/tarantool/test-run/commit/0292fe2ad901f33c903387f5498ade4f13acef94)).
> 

It seems too general change, while in my commit for retries it looks much more simple to rerun the test in loop.

> If you want my opinion about extracting checksums from comments — it is disgusting. Sorry.
> 

Ok, sure, I've removed it completely and updated the commits based only on JSON format.

> There are more points I would highlight about the patchset, but it does not have much sense if we'll get rid of checksums at all. The patchset modifies many different parts of test-run: too much for this small feature, IMHO.

In opposite we can't rerun w/o check of results files checksum will hide the new issues.

lib/test_suite.py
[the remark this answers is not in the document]
avtikhon: Removed.

kyukhin:
> I didn't look thoroughly, but I would highlight several points.
> 
> 1. The approach with checksums is not universal. It does not work if time-related information appears in a result file. It requires periodical checksums updating if a line number is reported as part of an error. I guess it is okay to operate on a test level and either mark it as stable or as fragile.

Using time related information in result file is disgusting. However I doubt such things exists to the moment. AFAIU, number of flaky tests supposed to be about 2-3 dozens and we do not update those tests heavily.

> 2. We can either implement reruns within a worker or on the level upward: within the task management system (dispathers-listeners-workers). Since there is nothing tarantool-specific I would look into the latter way. I pushed a variant to `Totktonada/rerun-fragile-test` branch ([0292fe2](https://github.com/tarantool/test-run/commit/0292fe2ad901f33c903387f5498ade4f13acef94)).

I think this is useless over complication. As you justly mentioned, this is simple feature. You approach doesn't look simple at all. I'd stick to Sasha's solution. Moreover your proposal will complicate subsequent changes in part of md5 search and match.
 
> If you want my opinion about extracting checksums from comments — it is disgusting. Sorry.

Agree.

> There are more points I would highlight about the patchset, but it does not have much sense if we'll get rid of checksums at all. The patchset modifies many different parts of test-run: too much for this small feature, IMHO.

I don't agree here. Patchset itself is not that big, it is like 90 lines of code and most changes looks really straightforward.

LeonidVas (commented):
I agree with A. Turenko that the use of hashsum for fragile tests is a slippery slope. But after some discussion, the Big Boss says:"On monday I'll push it to origin master as is." So, all the proposed changes - up to you. WBR , L. Vasilyev.

lib/test_suite.py
```diff
@@ -110,6 +111,14 @@ def __init__(self, suite_path, args):
                 lambda x: os.path.join(suite_path, x),
                 dict.fromkeys(self.ini[i].split())
                 if i in self.ini else dict())
+        for item in config.items("default"):
+            if item[0] == "fragile":
+                try:
+                    self.fragile = json.loads(item[1])
+                except ValueError:
+                    print "WARNING: suite '{}' uses old fragile format" . format(suite_path)
```
Totktonada, line 119: test-run redirects standard input / output / error streams and so `color_stdout()` is used across the whole codebase. This print will likely excuted before the redirection, but I would stay on the safe side and use `color_stdout()` as usual. BTW, it prints everything line-by-line and so prevent mixing output of different processes in a strange way.

I would accept the old format without deprecation warnings, but it is up to you.

lib/test_suite.py
```diff
@@ -110,6 +111,14 @@ def __init__(self, suite_path, args):
                 lambda x: os.path.join(suite_path, x),
                 dict.fromkeys(self.ini[i].split())
                 if i in self.ini else dict())
+        for item in config.items("default"):
```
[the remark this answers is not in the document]
Totktonada: I agree, there is the `self.ini` dictionary.

lib/worker.py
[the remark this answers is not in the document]
Totktonada: Nice discussion :)

lib/worker.py
```diff
@@ -327,7 +329,23 @@ def run_loop(self, task_queue, result_queue):
...
 
             result_queue.put(self.current_task(task_id))
-            short_status = self.run_task(task_id)
+            if is_fragile:
+                short_status = ''
+                retries_left = self.suite.fragile_retries()
+                # let's run till short_status became 'pass'
+                while short_status != 'pass' and retries_left >= 0:
+                    # print message only after some fails occurred
+                    if re.match(r'^fail:.*', short_status):
```
Totktonada, line 341: This commit does not change `short_status` from just `fail` to `fail:<checksum>`, so the regexp looks more as part of the next commit.

lib/test.py
Totktonada, line 259: I would pass it in some other way. Say, as a field in `WorkerTaskResult`. Because `short_status` now becomes a structure with a short status and a checksum, but it is not a structure, but a string… it is strange, at least. Not something that a reader of the code would expect.

lib/test.py
```diff
@@ -8,6 +8,7 @@
 import sys
 import traceback
 from functools import partial
+from hashlib import md5
```
Totktonada, line 11: I don't want to nitpick to much, but the commit header 'Add ability to check failed tests w/ fragile list' does not give the idea what the commit offers. It is like 'everything as before, but with *check*'. 'check' may be almost anything for one, who doesn't read the patch itself.

Totktonada:
> Using time related information in result file is disgusting.

It is about places, where an attempt to give more information in case of fail is performed ([example](https://github.com/tarantool/tarantool/blob/f7bcdf4c0405112d390b8ebb6d296717153020ef/test/replication/status.test.lua#L69-L74), consider `or box.info`).

> > We can either implement reruns within a worker or on the level upward: within the task management system (dispathers-listeners-workers).

> I think this is useless over complication.

The queue management is a bit complex part of test-run, yep. It is more appropriate place for this feature, but it is not something critical. I'm okay with the simpler approach.

> > The patchset modifies many different parts of test-run: too much for this small feature, IMHO.
>
> I don't agree here. Patchset itself is not that big, it is like 90 lines of code and most changes looks really straightforward.

I looked a big more into the patchset: it is mostly due to changing of `short_status` in a strange way: it becomes a structure with the status and checksum encoded as a string. So several places need to be updated to extract the actual status from this value.

Totktonada:
NB: Don't forget to verify that artifacts from an reexecuted test will be collected.

LeonidVas:
> NB: Don't forget to verify that artifacts from an reexecuted test will be collected.

Maybe I don’t understand you, but this is not necessary (with this approach to solving the problem of reruns fragile tests), because re-execution will be run in case of a known failure. If you want to get artifacts for some fragile test - remove the checksum of this test.

Totktonada:
> > NB: Don't forget to verify that artifacts from an reexecuted test will be collected.
> 
> Maybe I don’t understand you, but this is not necessary (with this approach to solving the problem of reruns fragile tests), because re-execution will be run in case of a known failure. If you want to get artifacts for some fragile test - remove the checksum of this test.

That's right, but I would not break ability to collect as much statistics as possible if we'll want to do it later.

kyukhin:
> > Using time related information in result file is disgusting.
> 
> It is about places, where an attempt to give more information in case of fail is performed ([example](https://github.com/tarantool/tarantool/blob/f7bcdf4c0405112d390b8ebb6d296717153020ef/test/replication/status.test.lua#L69-L74), consider `or box.info`).

So, this things always arise on error path. What's the problem then? We will never see that in .result.

Totktonada:
> > > Using time related information in result file is disgusting.
> > 
> > 
> > It is about places, where an attempt to give more information in case of fail is performed ([example](https://github.com/tarantool/tarantool/blob/f7bcdf4c0405112d390b8ebb6d296717153020ef/test/replication/status.test.lua#L69-L74), consider `or box.info`).
> 
> So, this things always arise on error path. What's the problem then? We will never see that in .result.

All those checksums are for output of a failed test.

lib/test_suite.py
```diff
@@ -154,19 +163,27 @@ def collect_tests(self):
         self.tests_are_collected = True
         return self.tests
 
+    def get_fragile_list(self):
+        res = []
+        try:
+            res = self.fragile['tests'].keys()
```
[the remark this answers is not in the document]
avtikhon: Ok, in this way changed the list to dictionary even for the old 'fragile' list.

lib/test_suite.py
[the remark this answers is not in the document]
avtikhon: Both comments implemented.

lib/test_suite.py
```diff
@@ -187,6 +187,15 @@ def fragile_tests(self):
                 res.append(test)
         return res
 
+    def get_test_fragile_checksums(self, test):
+        checksums = []
+        try:
+            for checksum in self.fragile['tests'][test]['checksums']:
+                checksums.append(checksum)
+        except Exception:
+            pass
```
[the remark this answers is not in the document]
avtikhon: No, some tests without checksums will not be rerun, but its entry will be used to set test in non parallel testing.

lib/worker.py
```diff
@@ -329,7 +330,9 @@ def run_loop(self, task_queue, result_queue, is_fragile):
                 break
 
             result_queue.put(self.current_task(task_id))
-            if is_fragile:
+            testname = re.split(r'/', task_id[0])[1]
```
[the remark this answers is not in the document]
avtikhon: Changed.

lib/test_suite.py
avtikhon: Right, decided to remove.

lib/worker.py
avtikhon: Right, corrected.

avtikhon:
> > > > Using time related information in result file is disgusting.
> > > 
> > > 
> > > It is about places, where an attempt to give more information in case of fail is performed ([example](https://github.com/tarantool/tarantool/blob/f7bcdf4c0405112d390b8ebb6d296717153020ef/test/replication/status.test.lua#L69-L74), consider `or box.info`).
> > 
> > 
> > So, this things always arise on error path. What's the problem then? We will never see that in .result.
> 
> All those checksums are for output of a failed test.

Right, but we have only a few of it:
```
grep -RI "or box.info" 2>/dev/null | grep -v ".result:" | grep -v ".reject:" | sed 's#:.*##g' | sort -u | wc -l
9
```

lib/test.py
avtikhon: Corrected as you suggested.

lib/test.py
avtikhon: Corrected.

avtikhon:
> > NB: Don't forget to verify that artifacts from an reexecuted test will be collected.
> 
> Maybe I don’t understand you, but this is not necessary (with this approach to solving the problem of reruns fragile tests), because re-execution will be run in case of a known failure. If you want to get artifacts for some fragile test - remove the checksum of this test.

Checked.

LeonidVas (approved):
Looks syntactically correct to me.

Totktonada:
I pushed two branches into tarantool today (and rerun failed jobs several times):

- Got several 'address already in use' fails (http_client.test.lua). It is the problem of test-run design (#141), not the test per se.
- Got several fails on net.box_reconnect_after_gh-3164.test.lua with the whole box.info.replication output (time related information, UUIDs).
- Got several fails on net.box_reconnect_after_gh-3164.test.lua with peer UUID, Unix socket path, schema version and so on (will differ from run to run).

There are other fails, but I didn't look whether the output was predictable (at least vinyl.ddl.test.lua, don't remember others).

I know, you're sure you can mark almost everything with checksums, but I see the opposite picture now. Maybe I'm just lucky.

At least, please, left ability to rerun a test, which is not marked with any checksum (I didn't look at the last patchset, maybe it is there already).

Aside of this, the 'transient fail' statistics was nice: no need to roll a screen to find whether reruns occur after full local testing.

avtikhon:
> I pushed two branches into tarantool today (and rerun failed jobs several times):
> 
I've run the same 3 days twice in hour and I got really lot of results. Checked some of them I've got the same results as Alexander below, but it was expected. I've updated the table [1] with the new column 'Checksum' and set there some of results, green - checksums that available, red - that have changing data. As I've checked before the same issue with checksum we'll have just in some tests - found only 9 with printing box info and some other tests printing other changing data.

[1] - https://docs.google.com/spreadsheets/d/1LGfHM_6tn84lzlz3XIXQnOqNY_KtzRWuZg33_yuv4YI/edit#gid=1890598391

> * Got several 'address already in use' fails (http_client.test.lua). It is the problem of test-run design (#141), not the test per se.

Right, we have some these fails, but it passes when tests run inline using fragile list for it.

> * Got several fails on net.box_reconnect_after_gh-3164.test.lua with the whole box.info.replication output (time related information, UUIDs).
> * Got several fails on net.box_reconnect_after_gh-3164.test.lua with peer UUID, Unix socket path, schema version and so on (will differ from run to run).
> 

Right this test I've in the table marked as red.

> There are other fails, but I didn't look whether the output was predictable (at least vinyl.ddl.test.lua, don't remember others).
> 
> I know, you're sure you can mark almost everything with checksums, but I see the opposite picture now. Maybe I'm just lucky.
> 

Right, we have some tests like this, but we have 2 ways to resolve it: fix the test either remove box information printing if it is not really needed. I've prepared the initial patch shown as green in the table mentioned above and the results really good - only 1 new fail found [2] ! 

[2] - https://gitlab.com/tarantool/tarantool/-/pipelines/193120097/builds

> At least, please, left ability to rerun a test, which is not marked with any checksum (I didn't look at the last patchset, maybe it is there already).
> 

That was the main cause why this feature with tests reruns I didn't want to have without checksums.

> Aside of this, the 'transient fail' statistics was nice: no need to roll a screen to find whether reruns occur after full local testing.

Actually after checksums will be written to the suites.ini files these fails would be really rare.

Totktonada:
> > Got several 'address already in use' fails (http_client.test.lua). It is the problem of test-run design (#141), not the test per se.
>
> Right, we have some these fails, but it passes when tests run inline using fragile list for it.

It is maybe okay as a temporary workaround, but it is not so complex to fix the cause: bind httpd.py on zero port (to let kernel to choose a random one) and read a real port from its output (we already read heartbeats from here).

> > At least, please, left ability to rerun a test, which is not marked with any checksum (I didn't look at the last patchset, maybe it is there already).
>
> That was the main cause why this feature with tests reruns I didn't want to have without checksums.

To be honest, I don't get the answer.

> > Aside of this, the 'transient fail' statistics was nice: no need to roll a screen to find whether reruns occur after full local testing.
>
> Actually after checksums will be written to the suites.ini files these fails would be really rare.

But I still want to know about this, when it occurs, to verify whether it is introduced by a patch I work on. It is okay to postpone this feature, but it would be usable.

lib/worker.py
```diff
@@ -329,24 +332,32 @@ def run_loop(self, task_queue, result_queue, is_fragile):
                 break
 
             short_status = None
+            result_checksum = None
             result_queue.put(self.current_task(task_id))
-            if is_fragile:
+            testname = os.path.basename(task_id[0])
+            fragile_checksums = self.suite.get_test_fragile_checksums(testname)
+            if is_fragile and fragile_checksums:
```
Totktonada, line 339: Here you block ability to rerun a test without a checksum. It is possible that a test cannot be verified just using a result file, but we want to allow reruns for it.

I remember the reason why you don't want to enable reruns by default when no checksums are provided: because it means that existing configs will let tests that are currently in a fragile list being retried.

However this is not so until we'll set `retries` explicitly, so it seems we can do this without subtle behaviour changes. I mean, if we'll consider empty (or non-existing) checksums list as 'rerun it anyway', everything will be good. But if you prefer explicit '*' in checksum, I don't mind.

You can also consider this as out of scope (not needed right now), but for me it looks quite natural that…

```json
fragile = {
    "retries": 2,
    "tests": {
        "bitset.test.lua": {
            "issues": ["gh-4095"]
        }
    }
}
```

…means 'rerun bitset.test.lua two times'.

Anyway, it is up to you. I just noted that the problem you highlighted looks as not applicable to our situation.

lib/test_suite.py
```diff
@@ -265,5 +265,11 @@ def run_test(self, test, server, inspector):
     def is_parallel(self):
         return self.ini['is_parallel']
 
+    def fragile_retries(self):
+        try:
+            return self.fragile['retries']
+        except Exception:
+            return 0
```
Totktonada, line 279: Nit: It is the same as `self.fragile.get('retries', 0)`.

Totktonada: And catching a general exception is considered as the bad practice.

lib/test_suite.py
```diff
@@ -110,6 +111,17 @@ def __init__(self, suite_path, args):
                 lambda x: os.path.join(suite_path, x),
                 dict.fromkeys(self.ini[i].split())
                 if i in self.ini else dict())
+        if config.has_option("default", "fragile"):
+            fragiles = config.get("default", "fragile")
```
Totktonada, line 115: Nit: I still don't get why not to use `self.ini.get('fragile', None)`. Since this dictionary is filled, it is used everywhere instead of `config` directly. Mix of `self.ini` and `config` accesses is a bit harder to read, IMHO.

dispatcher.py
```diff
@@ -388,7 +389,7 @@ def _run_worker(self, worker_id, tcp_port_range):
         os.environ['TEST_RUN_TCP_PORT_END'] = str(tcp_port_range[1])
         color_stdout.queue = self.result_queue
         worker = self.gen_worker(worker_id)
-        worker.run_all(self.task_queue, self.result_queue)
+        worker.run_all(self.task_queue, self.result_queue, self.is_fragile)
```
Totktonada, line 392: Several points (not critical, more kind of stylistic thoughts):

1. It seems more natural to pass it to gen_worker() function (because it is property of a worker in fact). If you'll do, don't forget about 'get _id as an arg' comment in worker.py.
2. For worker it is more like 'whether task retries are allowed' property, because it is quite abstract from tests. Maybe we can name it better: say, `enable_retries`.
3. However we can just lean on `retries` suite.ini value, it is available in `run_loop()`. There are two cons. First, we don't need to pass the flag over several calls deep into `run_loop()`. Second, it is quite natural to just follow `retries`.

Totktonada:
Status: It looks almost ready (considering given decisions), I just want to discuss several moments before we'll land the patchset.

avtikhon:
> > > Got several 'address already in use' fails (http_client.test.lua). It is the problem of test-run design (#141), not the test per se.
> > 
> > 
> > Right, we have some these fails, but it passes when tests run inline using fragile list for it.
> 
> It is maybe okay as a temporary workaround, but it is not so complex to fix the cause: bind httpd.py on zero port (to let kernel to choose a random one) and read a real port from its output (we already read heartbeats from here).
> 

Absolutely agree that run the tests like this is temporary workaround and it need to be fixed. And we'll continue work on its fixing in already exists standalone issues like:

https://github.com/tarantool/tarantool/issues/4949

> > > At least, please, left ability to rerun a test, which is not marked with any checksum (I didn't look at the last patchset, maybe it is there already).
> > 
> > 
> > That was the main cause why this feature with tests reruns I didn't want to have without checksums.
> 
> To be honest, I don't get the answer.
> 

I mean that blind rerun of the flaky issues is more dangerous than it's manual rerun that we currently have. Check of 'checksums' is a way to make these reruns partly automate to help the testing infrastructure to avoid of manual reruns. Standalone rerun we'll hide the new flaky issues.
Right, we can add the new feature to rerun the needed tests. It can be provided with special checksum/mark like:
   "checksum": [ "temporary_mark_to_rerun_on_any_fail_NOT_TO_COMMIT" ]
and add the special checker in test-run scripts on this mark to pass the tests on rerun on any failed checksum. It will help to check manually some flaky tests. But I think this feature must be committed in separate patch from the current patchset with standalone issue number.
Also in this way I would suggest to have new feature as Sergey Bronnikov suggested to have in some testing mode mandatory reruns for all tests to be able to find the flaky tests at the very start of its appearing, but this feature I would like to implement as standalone commit too.

> > > Aside of this, the 'transient fail' statistics was nice: no need to roll a screen to find whether reruns occur after full local testing.
> > 
> > 
> > Actually after checksums will be written to the suites.ini files these fails would be really rare.
> 
> But I still want to know about this, when it occurs, to verify whether it is introduced by a patch I work on. It is okay to postpone this feature, but it would be usable.

lib/worker.py
avtikhon: I've commented it in the same time when you wrote the questions in here:
https://github.com/tarantool/test-run/pull/217#issuecomment-698128881
in short: I agree with it, but as standalone commit for this feature.

lib/test_suite.py
avtikhon: Ok, sure, corrected.

lib/test_suite.py
avtikhon: Not possible, because self.ini['fragile'] consists of dictionary instead of string needed for json (because of

self.ini.update(dict(config.items("default")))

below), and convert it to string mess it, check the following change:
```
diff --git a/lib/test_suite.py b/lib/test_suite.py
index d1faa46..cac377a 100644
--- a/lib/test_suite.py
+++ b/lib/test_suite.py
@@ -112,7 +112,8 @@ class TestSuite:
                 dict.fromkeys(self.ini[i].split())
                 if i in self.ini else dict())
         if config.has_option("default", "fragile"):
-            fragiles = config.get("default", "fragile")
+            fragiles = self.ini.get('fragile', None)
+            print("==================================== {}\n" . format(type(fragiles)))
             try:
                 self.fragile = json.loads(fragiles)
                 if 'tests' not in self.fragile:
```
got:
```
==================================== <type 'dict'>

  File "./test-run.py", line 219, in <module>
    status = main_parallel()
  File "./test-run.py", line 134, in main_parallel
    res = main_loop_parallel()
  File "./test-run.py", line 86, in main_loop_parallel
    task_groups = lib.worker.get_task_groups()
  File "/home/avtikhon/Workspaces/tarantool/test-run/lib/worker.py", line 70, in get_task_groups
    suites = find_suites()
  File "/home/avtikhon/Workspaces/tarantool/test-run/lib/worker.py", line 29, in find_suites
    for suite_name in sorted(suite_names)]
  File "/home/avtikhon/Workspaces/tarantool/test-run/lib/test_suite.py", line 118, in __init__
    self.fragile = json.loads(fragiles)
  File "/usr/lib/python2.7/json/__init__.py", line 339, in loads
    return _default_decoder.decode(s)
  File "/usr/lib/python2.7/json/decoder.py", line 364, in decode
    obj, end = self.raw_decode(s, idx=_w(s, 0).end())
TypeError: expected string or buffer
```

dispatcher.py
avtikhon: Right, this flag is not needed and I completely removed it.

lib/test_suite.py
Totktonada: And this dict is strange. Okay.

Totktonada (approved):
LGTM.

Totktonada:
The test-run submodule is updated in tarantool in the following commits: [2.6.0-104-g43482eedc](https://github.com/tarantool/tarantool/commit/43482eedc3e139e24e9f12d55761f88ea7f9926c), [2.5.1-86-gc5bb549f6](https://github.com/tarantool/tarantool/commit/c5bb549f6d303c17cb63e4fd6021b22bd09d93b9), [2.4.2-70-gef330c3b0](https://github.com/tarantool/tarantool/commit/ef330c3b06a9fbab7e686347ba94856e614600b7), [1.10.7-35-g91260069c](https://github.com/tarantool/tarantool/commit/91260069c1dc53d86a0c756c2e821e1a9f9e0e95).

PR #219, opened 2020-08-18
Pass on space drop fail

avtikhon:
Found issue:
```
      [026] --- box/update.result   Thu Aug 13 01:06:17 2020
      [026] +++ box/update.reject   Tue Aug 18 01:31:24 2020
      [026] @@ -1,2130 +1,3 @@
      [026] -s = box.schema.space.create('tweedledum')
      [026]  ---
      [026] +- error: '/private/tmp/tnt/026_box/pretest_clean.lua:30: attempt to index a nil value'
      [026]  ...
      [026] -index = s:create_index('pk')
```
Issue happened in additional tool pretest_clean.lua which is used for
cleanning up the environment before running the test. Fail occurred
when space to be deleted was not found, like:
```
      tarantool> name = "BAD_SPACE_NAME" box.space[name]:drop()
      ---
      - error: '[string "name = "BAD_SPACE_NAME" box.space[name]..."]:1: attempt to
          index a nil value'
      ...
```
It could happen because their is some delay between getting the list of
spaces and the moment when these spaces are removing, check the failing
function from pretest_clean.sh file:
```
      box.space._space:pairs():map(function(tuple)
            local name = tuple[3]
            return name
        end):filter(function(name)
            -- skip internal spaces
            local first_char = string.sub(name, 1, 1)
            return first_char ~= '_'
        end):each(function(name)
            box.space[name]:drop()
        end)
```
In this delay the space could be already removed with information about
it. In this way getting the space with wrong name returned nil and drop
routine failed on it. To avoid of it the existence of the space should
be rechecked just before the call to its drop:

```
      box.space._space:pairs():map(function(tuple)
            local name = tuple[3]
            return name
        end):filter(function(name)
            -- skip internal spaces
            local first_char = string.sub(name, 1, 1)
            return first_char ~= '_'
        end):each(function(name)
            if box.space[name] ~= nil then box.space[name]:drop() end
        end)
```
Closes tarantool/tarantool#5247

### DIFF
--- a/pretest_clean.lua
+++ b/pretest_clean.lua
@@ -27,7 +27,9 @@ local function clean()
         local first_char = string.sub(name, 1, 1)
         return first_char ~= '_'
     end):each(function(name)
-        box.space[name]:drop()
+        if box.space[name] ~= nil then
+            box.space[name]:drop()
+        end
     end)
 
     local _USER_TYPE = 4


pretest_clean.lua
```diff
@@ -25,9 +25,12 @@ local function clean()
     end):filter(function(name)
         -- skip internal spaces
         local first_char = string.sub(name, 1, 1)
-        return first_char ~= '_'
+        return (name ~= nil and first_char ~= '_')
```
LeonidVas, line 28: Can the name be nil? The same effect will be if name = {} (the filter will be passed), but this is also not a valid case.

pretest_clean.lua
```diff
@@ -25,9 +25,12 @@ local function clean()
...
     end):each(function(name)
-        box.space[name]:drop()
+        local status, err = pcall(function(name) box.space[name]:drop() end, name)
+        if not status then
```
LeonidVas, line 31: If I understand correctly, you are trying to filter all invalid values in the previous step, and a WARNING should not be printed at all.

LeonidVas (changes requested):
It look like a bug in tarantool, not a test system. As I can see in https://gitlab.com/tarantool/tarantool/-/jobs/694714589 similar behavior starts after box/net.box_count_inconsistent_gh-3262.test.lua , is this test always the root of the problem? It look like we can use  artifacts for reproduce the problem. Or not?

alyapunov (changes requested):
Actually I don't understand what's going on. The main problem in the issue is _why_ `name` becomes `nil` but not how to handle `nil` value. I don't like (and don't undestand aclually) that `:pairs:map:filter:each` approach but when I tried to add logging I saw that names of spaces were repeated twice each. I would rewrite the cleanup code using normal plain `for` loops and sure that the problem will be no more.

pretest_clean.lua
LeonidVas: yep

avtikhon: Right, changed.

pretest_clean.lua
avtikhon: Right, not needed.

avtikhon:
> Actually I don't understand what's going on. The main problem in the issue is _why_ `name` becomes `nil` but not how to handle `nil` value. I don't like (and don't undestand aclually) that `:pairs:map:filter:each` approach but when I tried to add logging I saw that names of spaces were repeated twice each. I would rewrite the cleanup code using normal plain `for` loops and sure that the problem will be no more.

Right, changed.

LeonidVas (changes requested):
Hi! I think the patch is a bad idea. AFAIU you collect all the space names from the`box.space._space`, filter out all "internal" spaces, and drop all others. And problem look like you have a tuple in '_space' that points to a non-existent space.
Discussed the problem with Alexander. He said that he has a branch with this patch and that he uses the branch every day for testing. We decided to add `print` to this place to understand what space is causing the problem.

LeonidVas:
> I saw that names of spaces were repeated twice each.
Strange. I don't observe this behavior. Can you provide a reproducer?

Totktonada:
It would be good to have a reproducer, however I can imagine a situation, when the old code fails, but the new one will succeed. `:drop()` yields and so another fiber have a chance to drop a space. See:

```
$ rm 0000*
$ tarantool
Tarantool 2.7.0-72-g33b0dd582
type 'help' for interactive help
tarantool> box.cfg{} fiber = require('fiber') for i = 1, 1000 do box.schema.create_space('s' .. tostring(i)) end for i = 1, 10 do fiber.new(function() for i = 1, 10 do fiber.yield() end box.space['s' .. tostring(i)]:drop() end) end box.space._space:pairs():map(function(tuple) return tuple[3] end):filter(function(name) local first_char = string.sub(name, 1, 1) return first_char ~= '_' end):each(function(name) box.space[name]:drop() end)
<...>
2020-11-30 15:02:50.311 [5825] main/110/lua utils.c:1014 E> LuajitError: [string "box.cfg{} fiber = require('fiber') for i = 1,..."]:1: attempt to index a nil value
2020-11-30 15:02:50.311 [5825] main/111/lua utils.c:1014 E> LuajitError: [string "box.cfg{} fiber = require('fiber') for i = 1,..."]:1: attempt to index a nil value
2020-11-30 15:02:50.312 [5825] main/112/lua schema.lua:529 E> ER_NO_SUCH_SPACE: Space 's3' does not exist
---
...
```

Totktonada (approved):
The change itself is okay. I would add some explanation and push if others don't object.

LeonidVas:
> The change itself is okay. I would add some explanation and push if others don't object.

I would like to see the explanations before pushing. Please don't push without approval.

Totktonada:
> It would be good to have a reproducer, however I can imagine a situation, when the old code fails, but the new one will succeed. `:drop()` yields and so another fiber have a chance to drop a space. See:
> 
> ```
> $ rm 0000*
> $ tarantool
> Tarantool 2.7.0-72-g33b0dd582
> type 'help' for interactive help
> tarantool> box.cfg{} fiber = require('fiber') for i = 1, 1000 do box.schema.create_space('s' .. tostring(i)) end for i = 1, 10 do fiber.new(function() for i = 1, 10 do fiber.yield() end box.space['s' .. tostring(i)]:drop() end) end box.space._space:pairs():map(function(tuple) return tuple[3] end):filter(function(name) local first_char = string.sub(name, 1, 1) return first_char ~= '_' end):each(function(name) box.space[name]:drop() end)
> <...>
> 2020-11-30 15:02:50.311 [5825] main/110/lua utils.c:1014 E> LuajitError: [string "box.cfg{} fiber = require('fiber') for i = 1,..."]:1: attempt to index a nil value
> 2020-11-30 15:02:50.311 [5825] main/111/lua utils.c:1014 E> LuajitError: [string "box.cfg{} fiber = require('fiber') for i = 1,..."]:1: attempt to index a nil value
> 2020-11-30 15:02:50.312 [5825] main/112/lua schema.lua:529 E> ER_NO_SUCH_SPACE: Space 's3' does not exist
> ---
> ...
> ```

My bad, the 'attempt to index a nil value' error is from the spawned fibers. Only ER_NO_SUCH_SPACE is from the luafunish code.

LeonidVas:
> > The change itself is okay. I would add some explanation and push if others don't object.
> 
> I would like to see the explanations before pushing. Please don't push without approval.

I had a discussion with @Totktonada and @avtikhon and I don't mind such push. But please add a commit message describing how this can happen.

Totktonada:
Raw idea: restart a server if pretest_clean fails. This cleaner is not a silver bullet anyway.

Totktonada:
I [failed](https://github.com/tarantool/tarantool/issues/5247#issuecomment-750723176) to prove my initial guess that the problem is somewhere around dropping a space from a background fiber. If it would be a reason, your fix would not help: when `box.space` table has a space, which is actually removed, `:drop()` raises an error.

However I see that an effect of this kind may be achieved with `box.space.some_existing_space = nil`. What we can do in this situation?

1. Your way: left the space without an entry in the `box.space` Lua table. This way the next attempt to create a space with the same name will raise "Duplicate key exists in unique index 'name' in space '_space'". It seems, it'll only make things harder to investigate.
2. When we found tarantool in some unexpected state (say, when `require('pretest_clean').clean()` raises an error) just restart it.
3. If fails of this kind are rare, we can wait until GC64 related patch from LuaJIT upstream will be backported and looks whether the situation will be better.

I know, I approved this patch before as something that look harmless. But now I think a bit more and I see that it may be not so.

The problem still in the responsibility of QA team, so if you'll agree with @okoshovetc and @kyukhin to push it anyway, I'll not block.

okoshovetc:
Why don't we just use for loop as @alyapunov proposed? The iterator seems to be working strangely on _space space when there are ddl operations. Still it would be nice to have a reproducer for the issue on MacOS, but as long as the reasoning behind the patch is about test-run stability, I think the patch is a bit dirty, but OK.

Totktonada:
> Why don't we just use for loop as @alyapunov proposed?

Create a space, call `box.space.<space_name> = nil` and try to create it again.

Totktonada:
No updates from the author of the PR for more than three months. We still need a reproducer, otherwise everything is a kind of guess work. I'll close the PR as stale. Feel free to reopen if you'll get a reproducer.

PR #220, opened 2020-09-14
Fix reproduce file formatting (use flow style)

Totktonada:
Set flow style (JSON-like) explicitly instead of block style to don't
break indentation.

Before the patch (incorrect):

```yaml
- - app-tap/iconv.test.lua
- null
```

After the patch:

```yaml
- [app-tap/iconv.test.lua, null]
```

The default was changed in [1].

[1]: https://github.com/yaml/pyyaml/pull/256

### DIFF
--- a/dispatcher.py
+++ b/dispatcher.py
@@ -255,7 +255,8 @@ class Dispatcher:
                 'queue, but were not reported as done (does not matters '
                 'success or fail):\n', schema='test_var')
             for task_id in undone:
-                color_stdout('- %s' % yaml.safe_dump(task_id))
+                task_id_str = yaml.safe_dump(task_id, default_flow_style=True)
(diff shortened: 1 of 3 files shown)


avtikhon (approved):
Hi Alexander, thanks for the patch. I've checked the syntax, number of need entries to fix and patch in real run - all passed. Patch LGTM.

LeonidVas (approved):

Totktonada:
Thanks!

Totktonada:
Updated the test-run submodule in tarantool in 2.6.0-73-ga33cd1bd9, 2.5.1-72-g78b68f97c, 2.4.2-55-ged81ba6ef, 1.10.7-25-g83272871a.

PR #222, opened 2020-09-25
Justify columns in the output

Totktonada:


lib/utils.py
```diff
@@ -258,3 +258,10 @@ def prefix_each_line(prefix, data):
     data = data.rstrip('\n')
     lines = [(line + '\n') for line in data.split('\n')]
     return prefix + prefix.join(lines)
+
+
+def just_and_trim(s, width):
```
LeonidVas, line 263: pylint doesn't like such names http://pylint-messages.wikidot.com/messages:c0103 . I and Mons too)
PEP8 is on your side. So it's up to you.

LeonidVas (approved):
LGTM. See a comment.

lib/utils.py
Totktonada: I'm completely okay with such names when its scope is short (I would use `str`, but it is a type name in python). I don't see any reason to change `with open(<…>) as f` (or `fh`) too.

Anyway, what name you propose?

Totktonada: > lib/utils.py:264:4: R1705: Unnecessary "else" after "return" (no-else-return)

BTW, fixed this warning.

Totktonada: Renamed to `src`.

avtikhon (approved):
As I see this is a style change for output. I've checked and it seems looks nice. Patch itself LGTM.

Totktonada:
Updated the test-run submodule in tarantool in 2.6.0-112-g59dca1898, 2.5.1-92-ge2f2daea2, 2.4.2-76-gc71a5b4aa, 1.10.7-39-g821f490e3.

PR #227, opened 2020-10-23
Add box_snapshot wrapper

avtikhon:
Found the issue with collecting results file checksum failing on
box.snapshot calls. In results file it returned changing data, like:

[076] --- vinyl/snapshot.result	Thu Oct 22 09:46:53 2020
[076] +++ vinyl/snapshot.reject	Thu Oct 22 19:35:41 2020
[076] @@ -128,7 +128,7 @@
[076]  ...
[076]  box.snapshot()
[076]  ---
[076] -- ok
[076] +- error: 'Invalid VYLOG file: Slice 1664 deleted but not registered'
[076]  ...
[076]  s:select({99}, {iterator = box.index.LE, limit = 10})
[076]  ---
[076]

Where slice number always different. To be able to collect results file
checksum for rerun ability added helper box_snapshot for box.snapshot
to save changing data in log file instead of saving it in results file.

### DIFF
--- a/test_run.lua
+++ b/test_run.lua
@@ -437,6 +437,17 @@ local function wait_log(self, node, what, bytes, timeout, opts)
     return wait_cond(self, cond, timeout, delay)
 end
 
+-- Helper for box.snapshot to save changing data in log file instead of
+-- saving in results file, to be able to collect results file checksum
+-- for rerun ability.
+local function box_snapshot()
+    local ok, res = pcall(box.snapshot)
+    if not ok then
+        log.error("box.snapshot() failed with error: " .. res)
+    end
+    return ok
+end
+
 local inspector_methods = {
     cmd = cmd,
     eval = eval,
@@ -464,6 +475,7 @@ local inspector_methods = {
     grep_log = grep_log,
     wait_cond = wait_cond,
     wait_log = wait_log,
+    box_snapshot = box_snapshot,
 }
 
 local function inspector_new(host, port)


test_run.lua
```diff
@@ -437,6 +437,17 @@ local function wait_log(self, node, what, bytes, timeout, opts)
     return wait_cond(self, cond, timeout, delay)
 end
 
+-- Helper for box.snapshot to save changing data in log file instead of
+-- saving in results file, to be able to collect results file checksum
+-- for rerun ability.
+local function box_snapshot()
+    local ok, err = pcall(box.snapshot)
+    if ok == false then
```
LeonidVas, line 445: You can use `not ok`  because `pcall` always returns a bool.

LeonidVas (approved):
Hi! Thank you for the patch.
This is a first kludge for use the checksums)
I suggest following https://github.com/tarantool/tarantool/wiki/Code-review-procedure#commit-message : "...started from lowercase letter..."

LeonidVas:
All comments up to you.

test_run.lua
avtikhon: Ok, changed.

Totktonada:
There is the special mechanism for such cases: `test_run:cmd("push filter <...>")`. [Example](https://github.com/tarantool/tarantool/blob/1ad696c09d5735afb0888e9351660b963f432f12/test/wal_off/oom.test.lua#L4).

Totktonada:
Alexander (@avtikhon) said that using filters fit his needs. Since it is more general solution, I would stick with it. Closing so.

Alternatives (in context of a 'core = tarantool' test):

```lua
-- The parentheses are to discard return values except first one.
-- We need it here to discard an error message.
(pcall(box.snapshot))
```

Or:

```lua
-- Discard error message.
ok = pcall(box.snapshot)
ok
```

Since the similar need may occur for another function, I would not introduce the specific helper for `box.snapshot()`.

We can list some tricks in the README instead.

PR #232, opened 2020-11-04
Use SIGKILL in drop_cluster() routine

avtikhon:
By default drop_cluster() routine uses SIGTERM signal to stop the
replications. Found that in some situations SIGTERM couldn't kill
all replicas and some processes left. To avoid of such situations
better to use SIGKILL signal to be sure that all the replications
processes will be stopped.


Totktonada:
Superseded by PR #186.

PR #236, opened 2020-11-11
Add signal argument to server restart routine

avtikhon:
Found that stop and restart routines should have the ability to set
signal using restart routine to fix the issues, like:
```
Test hung! Result content mismatch:
--- replication-py/init_storage.result	Wed Aug 26 06:06:15 2020
+++ /tmp/tnt/101_replication-py/init_storage.result	Wed Nov 11
10:17:50 2020
@@ -130,5 +130,3 @@
 -------------------------------------------------------------
 waiting reconnect on JOIN...
 ok
-waiting reconnect on SUBSCRIBE...
-ok
```

### DIFF
--- a/.tarantoolctl
+++ b/.tarantoolctl
@@ -1,0 +1,15 @@
+-- Options for test-run tarantoolctl
+
+local workdir = os.getenv('TEST_WORKDIR')
+default_cfg = {
+    pid_file   = workdir,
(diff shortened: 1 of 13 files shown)


Totktonada:
I guess it was resolved without changing of Python tests by PR #186. Right?

avtikhon:
Right, fixed.

PR #237, opened 2020-11-18
Improve observability of the testing system

Totktonada:
Sometimes it is hard to understand what exactly the testing system doing. In this patchset I made attempt to improve its observability: give meaningful error messages, log more events and so on.

I picked up [one of the tests][1], which fails from time to time, and use it as a test for the testing system: run it, look at output and logs, change test-run to give me useful details, repeat.

[1]: https://github.com/tarantool/tarantool/issues/5395

### DIFF
--- a/lib/options.py
+++ b/lib/options.py
@@ -76,6 +76,14 @@ class Options:
                 Default: false.""")
 
         parser.add_argument(
+                '--debug',
+                dest='debug',
+                action='store_true',
+                default=False,
+                help="""Print test-run logs to the terminal.
+                Default: false.""")
+
+        parser.add_argument(
                 "--force",
                 dest="is_force",
                 action="store_true",
--- a/listeners.py
+++ b/listeners.py
@@ -258,11 +258,20 @@ class HangWatcher(BaseWatcher):
         self.worker_current_task = dict()
 
     def process_result(self, obj):
-        self.warned_seconds_ago = 0.0
-        self.inactivity = 0.0
-
+        # Track tasks in progress.
         if isinstance(obj, WorkerCurrentTask):
             self.worker_current_task[obj.worker_id] = obj
+
+        # Skip irrelevant events.
+        if not isinstance(obj, WorkerOutput):
+            return
+
+        # Skip color_log() events if --debug is not passed.
+        if obj.log_only and not lib.Options().args.debug:
+            return
+
+        self.warned_seconds_ago = 0.0
+        self.inactivity = 0.0
 
     def process_timeout(self, delta_seconds):
         self.warned_seconds_ago += delta_seconds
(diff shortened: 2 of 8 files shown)


Totktonada:
TODO list:

- Add timeout for execution of a particular tarantool command (Lua or SQL) to don't hit the common `--no-output-timeout` limit.
- Add timeout for starting of a tarantool instance.
- Add timeout for stopping of a tarantool instance.
- Move .tarantoolctl configuration file to test-run repository and provide lower `replication_sync_timeout` default value to fit into the common `--no-output-timeout` limit (120 seconds by default). Default tarantool's `replication_sync_timeout` is 300 seconds.

Totktonada:
It seems, I'll not able to continue and resolve my TODO items soon. So I'm going to mark the PR as ready for review.

avtikhon:
PR successfully checked on multi runs in CI:
https://gitlab.com/tarantool/tarantool/-/pipelines?page=1&scope=all&ref=avtikhon%2Fflaky_stable_min_prs
LGTM

avtikhon (approved):

lib/options.py
```diff
@@ -75,6 +75,14 @@ def __init__(self):
                 help="""Print TAP13 test output to log.
                 Default: false.""")
 
+        parser.add_argument(
+                "--debug",
+                dest='debug',
```
LeonidVas, line 80: Why single quotes?

LeonidVas (approved):
LGTM.

lib/options.py
Totktonada: I generally prefer single quotes in Python and Lua (without any reasons, just habit). I have to be consistent with the surrounding code, but there is no consistency within the file. I'll use single quotes within the code block so (if there are no objections).

Totktonada: Done.

listeners.py
```diff
@@ -147,7 +147,25 @@ def process_result(self, obj):
...
+        # In OutputWatcher() such chunks are bufferized and
+        # written to the terminal when a newline arrives (to
+        # don't mix partial lines from different workers).
+        #
+        # Here we want to dump output without any buffering,
+        # because it may help with debugging a problem. So
+        # we just mark such chunks and write them out.
+        output = obj.output
+        if not decolor(output).endswith('\n'):
+            output = output + ' <no line carry>\n'
```
Totktonada, line 162: @Mons proposed to change it to something with a 'newline' word (referring diff's `\ No newline at end of file`). I wrote it as 'line carry' to don't make confusion around a 'new line' and a 'newline symbol', but it seems I introduced an association with the 'line carriage' symbol this way (while we're about the 'line feed' symbol in fact).

The ideal solution (restricting ourself with ASCII) would be something like `<no newline at end of chunk>`, but since it is not a primary information, it should be short. I'll pick up `<no newline>` and will hope that it is more intuitive than `<no line carry>`.

Totktonada: Done.

Totktonada:
Thank you all for scrutiny, testing and comments!

Totktonada:
Updated the test-run submodule in tarantool in 2.7.0-71-ga53760477, 2.6.1-61-gd58c82766, 2.5.2-34-g39702e742, 1.10.8-24-g3b55af5d2.

Issue #78, opened 2018-01-25
Simplify configuration needed for setup tap / tarantool tests


Totktonada:
core = tarantool tests:

- [ ] use default .tarantoolctl when there is no .tarantoolctl
- [ ] use default `script` (e.g. master.lua or app.lua) when there is no one

all tests:

- [ ] add recipe how to integrate test-run into the new project, like so:

```
In project dir:

git submodule add https://github.com/tarantool/test-run.git
mkdir test && cd test
ln -s ../test-run/test-run.py ./
cd -

In Makefile:

.PHONY: test
test:
    virtualenv -p python2.7 ./.env-2.7
    source ./.env-2.7/bin/activate && \
        pip install -r ./test-run/requirements.txt && \
        pip install tarantool && \
        cd test && ./test-run.py
```

Totktonada:
The first bullet is implemented in PR #242.

PR #242, opened 2020-11-24
Move tarantoolctl to test-run tool submodule and add replication_sync_timeout to it

avtikhon:
1. Move tarantoolctl to test-run tool submodule
    
    Moved tarantoolctl to test-run tool submodule repository as:
    
      <tarantool repository>/test-run/.tarantoolctl
    
    Also set backward compability to use old path location:
    
      <tarantool repository>/test/.tarantoolctl
    
    as its primary place.

    To set PWD environment for tarantoolctl for popen() call tried to use
    os.environ. It was needed for the later commits where tarantoolctl
    would be moved from to test-run repository and its configuration file
    would be read right from the working directory of the test. In this
    case all os.putenv() calls changed to environment set with os.environ,
    to be able to modify environment directly, check [1] ("Note: Calling
    putenv() directly does not change os.environ"). But in this case all
    the tests had to be changed to use os.environ instead of os.putenv(),
    it will block the use of os.putenv() in the future tests. Decided to
    avoid of this way and temporary change 'PWD' around the Popen() call.
    
    Needed for tarantool/tarantool#5504
    Part of #78
    
    [1] - https://docs.python.org/2/library/os.html#process-parameters

2. Setup replication_sync_timeout at .tarantoolctl
    
    Found that tests may fail due to hang in seek_wait() loop on starting
    and stopping instances. It happened, because instances were not synced
    within default output timeout which is by default 120 seconds, while
    replication sync happens only each 300 seconds by default. To fix it
    replication_sync_timeout should be decreased to some value lower than
    'no-output-timeout', so decided to set it to 100 seconds.

    The issue looked like in tests:
    ```
      --- replication/gh-5140-qsync-casc-rollback.result    Mon Oct 19
      17:29:46 2020
      +++ /rw_bins/test/var/070_replication/gh-5140-qsync-casc-rollback.result
      Mon Oct 19 17:31:35 2020
      @@ -169,56 +169,3 @@
       -- all the records are replayed one be one without yields for WAL writes, and
       -- nothing should change.
       test_run:cmd('restart server default')
      - |
      -test_run:cmd('restart server replica')
      - | ---
      - | - true
      - | ...
      -
    ```
    Part of tarantool/tarantool#5504

### DIFF
--- a/.tarantoolctl
+++ b/.tarantoolctl
@@ -1,0 +1,15 @@
+-- Options for test-run tarantoolctl
+
+local workdir = os.getenv('TEST_WORKDIR')
+default_cfg = {
+    pid_file   = workdir,
+    wal_dir    = workdir,
+    memtx_dir  = workdir,
+    vinyl_dir  = workdir,
+    log        = workdir,
+    background = false,
+}
+
+instance_dir = workdir
+
+-- vim: set ft=lua :
--- a/.tarantoolctl
+++ b/.tarantoolctl
@@ -1,6 +1,9 @@
 -- Options for test-run tarantoolctl
 
+-- Note: tonumber(nil) is nil.
 local workdir = os.getenv('TEST_WORKDIR')
+local replication_sync_timeout = tonumber(os.getenv('REPLICATION_SYNC_TIMEOUT'))
+
 default_cfg = {
     pid_file   = workdir,
     wal_dir    = workdir,
@@ -8,6 +11,7 @@ default_cfg = {
     vinyl_dir  = workdir,
     log        = workdir,
     background = false,
+    replication_sync_timeout = replication_sync_timeout,
 }
 
 instance_dir = workdir
--- a/lib/tarantool_server.py
+++ b/lib/tarantool_server.py
@@ -848,6 +848,12 @@ class TarantoolServer(Server):
             os.putenv("MASTER", self.rpl_master.iproto.uri)
         self.logfile_pos = self.logfile
 
+        # This is strange, but tarantooctl leans on the PWD
+        # environment variable, not a real current working
+        # directory, when it performs search for the
+        # .tarantoolctl configuration file.
+        os.environ['PWD'] = self.vardir
+
         # redirect stdout from tarantoolctl and tarantool
         os.putenv("TEST_WORKDIR", self.vardir)
         self.process = subprocess.Popen(args,
@@ -855,6 +861,9 @@ class TarantoolServer(Server):
                                         stdout=self.log_des,
                                         stderr=self.log_des)
         del(self.log_des)
+
+        # Restore the actual PWD value.
+        os.environ['PWD'] = os.getcwd()
 
         # gh-19 crash detection
         self.crash_detector = TestRunGreenlet(self.crash_detect)
--- a/lib/tarantool_server.py
+++ b/lib/tarantool_server.py
@@ -776,7 +776,15 @@ class TarantoolServer(Server):
                     if (e.errno == errno.ENOENT):
                         continue
                     raise
-        shutil.copy('.tarantoolctl', self.vardir)
+        # Previously tarantoolctl configuration file located in tarantool
+        # repository at test/ directory. Currently it is located in root
+        # path of test-run/ submodule repository. For backward compatibility
+        # this file should be checked at the old place and only after at
+        # the current.
+        tntctl_file = '.tarantoolctl'
+        if not os.path.exists(tntctl_file):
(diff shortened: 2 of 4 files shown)


Totktonada:
Please, mark the first commit as part of #78.

.tarantoolctl
```diff
@@ -0,0 +1,16 @@
...
+
+local workdir = os.getenv('TEST_WORKDIR')
+default_cfg = {
+    pid_file   = workdir,
+    wal_dir    = workdir,
+    memtx_dir  = workdir,
+    vinyl_dir  = workdir,
+    log        = workdir,
+    background = false,
+    replication_sync_timeout = 100,
```
Totktonada, line 11: Should be passed from the option (via the environment variable).

lib/tarantool_server.py
```diff
@@ -767,7 +767,10 @@ def copy_files(self):
                     if (e.errno == errno.ENOENT):
                         continue
                     raise
-        shutil.copy('.tarantoolctl', self.vardir)
+        tntctl_file = '.tarantoolctl'
+        if not os.path.exists(tntctl_file):
+            tntctl_file = os.path.join(self.TEST_RUN_DIR, '.tarantoolctl')
```
Totktonada, line 772: Please, clarify it with a short comment: it is not clear that the reason to find the file in two different places in backward compatibility. Future readers of the code will not have any advice, whether it may be simplified to just copy the file from test-run.

Totktonada:
Please, also mention the issue in tarantool that motivates those changes.

lib/tarantool_server.py
avtikhon: Comment added.

.tarantoolctl
avtikhon: Set to use environment variable.

avtikhon:
> Please, also mention the issue in tarantool that motivates those changes.

Made all the changes as you suggested, please review.

lib/tarantool_server.py
```diff
@@ -825,10 +828,13 @@ def start(self, silent=True, wait=True, wait_load=True, rais=True, args=[],
 
         # redirect stdout from tarantoolctl and tarantool
         os.putenv("TEST_WORKDIR", self.vardir)
+        saved_cwd = os.getcwd()
+        os.putenv("PWD", self.vardir)
```
Totktonada, line 837: This hell really deserves its own description, so I'll extract this change into its own commit. Hope you don't mind.

Totktonada:
Made several changes:

1. Moved the comment re `.tarantoolctl` file search into the appropriate commit.
2. Renamed `repl_sync_timeout` to `replication_sync_timeout`. There is no need to abbreviate it.
3. Extracted `PWD` tweaking into its own commit, because it deserves its own explanation and may be considered as the bug fix on its own. Used `os.environ` here, because why not. Always set PWD to actual cwd with `os.getcwd()`, because, again, why not.
4. Fixed misusage of 'tarantoolctl' word, where '.tarantoolctl' configuration file was meant.
5. Rebased on top of the current master.

lib/tarantool_server.py
Totktonada: Extracted into its own commit.

Totktonada (approved):
I made some tweaks and now everything looks okay for me. I think we both spin around here enough time and the change can be pushed. I verified in-source and out-of-source builds before and after the corresponding tarantool [patch](https://lists.tarantool.org/pipermail/tarantool-patches/2020-December/021041.html) and everything looks okay too.

Totktonada:
Found a nice side effect: now we can run `./test/test-run.py` and `cd test && ./test-run.py` is not mandatory anymore. I'll add a note about this into the commit message.

Totktonada:
(Regarding the last force-push.) Fixed a typo in a commit message.

Totktonada:
Updated the test-run submodule in tarantool in the following commits: [2.7.0-97-g6c8efa849](https://github.com/tarantool/tarantool/commit/6c8efa8491ed3fc2834938ec65d91dc1f7ec7e34), 2.6.1-84-g0d738037b, 2.5.2-48-gfa36857a3, 1.10.8-33-g64e2fb849.

Totktonada:
Removed the `test/.tarantoolctl` file from the tarantool repository in the following commits: [2.7.0-98-g54667f6fa](https://github.com/tarantool/tarantool/commit/54667f6fa86ff1ab0030501abc9a59d0d7aea82f), 2.6.1-85-g8d4623652, 2.5.2-49-gabd4660df, 1.10.8-34-gb5d31b60a. After this the new --replication-sync-timeout option (and its default) will be in effect.

Issue #256, opened 2020-12-15
test-run does not kill server that hangs in an instance script


Totktonada:
### How to reproduce

Mangle a supplementary (non-default) instance like so:

```diff
diff --git a/test/app-tap/debug/server.lua b/test/app-tap/debug/server.lua
index 3362d46cc..8dbd74b66 100644
--- a/test/app-tap/debug/server.lua
+++ b/test/app-tap/debug/server.lua
@@ -9,3 +9,5 @@ box.schema.func.create('debug.sourcefile')
 box.schema.func.create('debug.sourcedir')
 box.schema.user.grant('guest','execute','function','debug.sourcefile')
 box.schema.user.grant('guest','execute','function','debug.sourcedir')
+
+while true do end
```

And run a test (remove `--test-timeout 5` for test-run prior PR #244):

```sh
./test/test-run.py --test-timeout 5 -j1 app-tap/debug.test.lua
```

Expected behaviour: the test will fail at reaching the test timeout and all relevant instances will be terminated / killed.

Observed behaviour: the test timeout works, but the instance is not stopped. If the test is marked as fragile and we make attempt to rerun it, we'll get the 'The daemon is already running' error.

### Investigation

It seems, the reason is the following check in the `TarantoolServer.stop()` method:

https://github.com/tarantool/test-run/blob/1f6d7ba65938f2feefeeab158625fb2ca820887e/lib/tarantool_server.py#L845-L849

Given server is not considered as 'started', so test-run does not stop it. However the process exists and should be killed.

The attempt to just remove the check fails:

<details>
<summary>The entire output (without this check)</summary>

```
$ ./test/test-run.py --test-timeout 5 -j1 --force app-tap/debug.test.lua
Started ./test/test-run.py --test-timeout 5 -j1 --force app-tap/debug.test.lua
Running in parallel with 1 workers

Collecting tests in 'app'            (Found 0   tests): application server tests.
Collecting tests in 'app-tap'        (Found 1   tests): application server tests (TAP).
Collecting tests in 'box'            (Found 0   tests): Database tests.
Collecting tests in 'box-py'         (Found 0   tests): legacy python tests.
Collecting tests in 'box-tap'        (Found 0   tests): Database tests with #! using TAP.
Collecting tests in 'engine'         (Found 0   tests): tarantool multiengine tests.
Collecting tests in 'engine_long'    (Found 0   tests): tarantool engine stress tests.
[...]

Issue #255, opened 2020-12-15
The same Test object may be used several times


Totktonada:
See:

https://github.com/tarantool/test-run/blob/1f6d7ba65938f2feefeeab158625fb2ca820887e/lib/worker.py#L300-L304

When we run something like `./test/test-run.py -j1 --force foo/bar.test.lua foo/bar.test.lua`, the same Test object will be used, because it has the same ID ('foo/bar.test.lua', None) — (test_name, conf_name). Usually it does not lead to any problem, but there are corner cases.

Say, if we report a failure (last log lines and so on), we store the fact that the fail was reported in `is_crash_reported` in the Test instance. If both runs fails, the second will not show logs.

Not sure, whether it may lead to other problems.

Issue #254, opened 2020-12-15
Take a closer look at crash detector


Totktonada:
There are several points around crash detector that looks dubious for me:

1. Crash detector is started before `wait_until_started()`. So there is a race condition and we can miss `TarantoolStartError`?
2. Crash detector is not killed / joined after `TarantoolStartError`. Not sure whether it affect anything, but it is strange.

Issue #253, opened 2020-12-15
Take a look at non-default server failure detection


Totktonada:
The implementation kills a currently executed greenlet (`<Test instance>.current_test_greenlet`) in the case of fail of a non-default instance. A greenlet that is stored in the `<Test instance>.current_test_greenlet` field may be killed at any point by the `kill_current_test()` method.

So we can assume some predicates after `<greenlet>.join()` as true, however it may be not so. And the greenlet will not perform any shutdown actions: they should be done outside, somewhere after `<greenlet>.join()` / `<greenlet>.get()`. That's all make the code less obvious.

We already store all servers that are related to a particular test in the TestState object. So, maybe, we can reverse the dependency and periodically check whether a failure of any other server occurs from a code that executes a test.

PR #257, opened 2020-12-15
Stop instances properly

### LINKED ISSUE #65: test-run fails to kill cluster servers in case of an error

### LINKED ISSUE #157: Kill hung tarantool with SIGKILL if SIGTERM is not sufficient

Totktonada:
Attempt to get rid of 'The daemon is already running' error and hang instances after testing.

- Wait for a process to terminate after a signal to get rid of the 'The daemon is already running' error.
- Stop supplementary (non-default) instances after an app test.
- Use SIGKILL, when SIGTERM is not sufficient for non-default instance of an app test (follows up #186).
- Kill an app server if it hangs and a non-default instance fails.

I use the test cases from the PR #244 and tweak them to hang or fail instances at different points: before startup or after. Say, hang a default app server and fail a supplementary one after startup.

Follows up PR #244.
Fixes #65.
Fixes #157.

----

While looking into this, found several doubtful situations or code: #252, #253, #254, #255, #256.

### DIFF
--- a/lib/app_server.py
+++ b/lib/app_server.py
@@ -9,6 +9,7 @@ from gevent.subprocess import Popen, PIPE
 
 from lib.colorer import color_stdout
 from lib.colorer import color_log
+from lib.colorer import qa_notice
 from lib.options import Options
 from lib.preprocessor import TestState
 from lib.server import Server
@@ -18,6 +19,7 @@ from lib.tarantool_server import TarantoolServer
 from lib.tarantool_server import TarantoolStartError
 from lib.utils import find_port
 from lib.utils import format_process
+from lib.utils import signame
 from lib.utils import warn_unix_socket
 from test import TestRunGreenlet, TestExecutionError
 from threading import Timer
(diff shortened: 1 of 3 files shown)


avtikhon (approved):

LeonidVas (approved):
LGTM.

Totktonada:
Updated the test-run submodule in tarantool in [2.7.0-116-g6e6e7b29b](https://github.com/tarantool/tarantool/commit/6e6e7b29b913c9ccd27e731c3a781d77f27af79c), 2.6.1-101-g13ff2ff2b, 2.5.2-65-gcdc10c888, 1.10.8-44-g0e95810dd.

PR #259, opened 2020-12-22
Add timeouts setup from environment

### LINKED ISSUE #258: Implement ability to get test-run timeouts from the hosts local configurations
Trying to fix #251 found that on old hosts vinyl/ suite tests failed to finish in default timeout '--test-timeout' == 110 seconds. Also it happened because of slow IO with slow hard drives setup on hosts. In major it happened on box.snapshot() calls, but could happen on any other commands.
To fix it was decided to collect statistics and setup on hosts local timeout configurations.
Found that running in complete parallel mode for all the tests even from fragile lists then test-timeout value had results:
  110 (default) -> a lot of fails
  210 -> fails rare, but still exists in different places, a lot of checksums needed to pass
  310 -> fails very rare, exists in a few places only, small number of checksums may help to pass it

avtikhon:
Added ability to set using environment the following timeouts based on
appropriate run options:
```
  TEST_TIMEOUT			--test-timeout
  NO_OUTPUT_TIMEOUT		--no-output-timeout
  REPLICATION_SYNC_TIMEOUT	--replication-sync-timeout
```
Closes #258

### DIFF
--- a/dispatcher.py
+++ b/dispatcher.py
@@ -143,10 +143,9 @@ class Dispatcher:
             self.listeners.append(self.fail_watcher)
         if watch_hang:
             warn_timeout = 60.0 if args.long else 10.0
-            no_output_timeout = float(args.no_output_timeout or 120)
             hang_watcher = listeners.HangWatcher(
                 output_watcher.not_done_worker_ids, self.kill_all_workers,
-                warn_timeout, no_output_timeout)
+                warn_timeout, float(args.no_output_timeout))
             self.listeners.append(hang_watcher)
 
     def run_max_workers(self):
--- a/lib/options.py
+++ b/lib/options.py
@@ -193,7 +193,7 @@ class Options:
         parser.add_argument(
                 "--test-timeout",
                 dest="test_timeout",
-                default=110,
+                default=env_int('TEST_TIMEOUT', 110),
                 type=int,
                 help="""Break the test process with kill signal if the test runs
                 longer than this amount of seconds. Default: 110 [seconds].""")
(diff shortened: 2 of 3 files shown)


okoshovetc (approved):
LGTM

dispatcher.py
```diff
@@ -143,10 +143,9 @@ def init_listeners(self):
             self.listeners.append(self.fail_watcher)
         if watch_hang:
             warn_timeout = 60.0 if args.long else 10.0
-            no_output_timeout = float(args.no_output_timeout or 120)
             hang_watcher = listeners.HangWatcher(
                 output_watcher.not_done_worker_ids, self.kill_all_workers,
-                warn_timeout, no_output_timeout)
+                warn_timeout, args.no_output_timeout)
```
Totktonada, line 148: It seems, it does not break anything, but I would keep the `float` type. At least, it is what HangWatcher expects.

lib/options.py
```diff
@@ -193,15 +193,15 @@ def __init__(self):
         parser.add_argument(
                 "--test-timeout",
                 dest="test_timeout",
-                default=110,
+                default=env_int('TEST_TIMEOUT', 110),
```
Totktonada, line 196: Since those values may be passed implicitly now, I would show them on the terminal before start testing.

Totktonada (approved):
LGTM aside of two non-critical comments.

lib/options.py
avtikhon: Sure, added print block, looks like:
```
Running in parallel with 16 workers

Timeout variables:
-------------------
REPLICATION_SYNC_TIMEOUT: 100
TEST_TIMEOUT:             110
NO_OUTPUT_TIMEOUT:        120

Collecting tests in 'app'            (Found 0   tests): application server tests.
```

dispatcher.py
avtikhon: Ok, sure, added float().

Issue #15, opened 2015-10-12
There is no python2 on OSX (it has python2.7)


mejedi:
Test-run.py fails to run osx, because it doesn't have python2 (specified in test-run via schebang).

I suggest changing schebang to python2.7 instead of python2.

Totktonada:
I hope it'll be resolved by #20.

ligurio:
Duplicate of #20.

Issue #39, opened 2016-11-22
Replace test_run:get_cfg('user') with environment variables


rtsisyk:
Environment variables is well-known and well-defined way to pass arguments to processes. Please stop re-inventing the wheel and use setenv() to pass arguments.

bigbes:
What's for and what's the problem with that way?

rtsisyk:
I meant that test-run can just set environment variables instead of implementing it own incompatible way (`inspector:get_param()`) to pass arguments to tarantool.

Issue #20, opened 2016-01-07
[10sp] Make test (and other python code) Python3 compatible


kostja:
Python3 is the next major version and some Linux distros (e.g. Linux Arch) are already use it as a default. It makes sense to make the code python3 compatible. In theory conversion should be simple especially if you use tools like py2to3.

Moved from https://github.com/tarantool/tarantool/issues/56

anatol:
Python2 is EOL. Python3 is a default version at major OS distributions. Could you please prioritize this effort and make tarantool project python2-free?

ligurio:
I would try to decompose ticket to subtasks:
- [ ] remove relative module imports
- [ ] fix module imports for modules changed in Python 3 (for example ConfigParser)
- [ ] fix printing to make it compatible with 2 and 3 versions
- [ ] support Python 3 in Tarantool tests

ligurio:
[Tarantool-patches] [PATCH v1] Support to run tests with Python 3
https://lists.tarantool.org/pipermail/tarantool-patches/2020-November/020872.html

PR #263, opened 2021-01-11
Support Python 3

ligurio:
### Uncommitted changes:
- patch for Tarantool tests (branch: https://github.com/tarantool/tarantool/compare/ligurio/python3-support)
- ~~bump tarantool-python version in test-run to include patches (https://github.com/tarantool/test-run/pull/264)~~
    - ~~https://github.com/tarantool/tarantool-python/pull/186~~
    - ~~https://github.com/tarantool/tarantool-python/pull/187~~

#### vshard, Python 2

PR to bump test-run version in vshard - https://github.com/tarantool/vshard/pull/268

#### Tarantool, Python 3

CI (enabled Python 3 in `test-run.py`, branch `ligurio/python3-support` in tarantool): everything is passed except job with FreeBSD, it requires manual update of golden image https://github.com/tarantool/tarantool/pull/5751

#### Tarantool, Python 2

We want to keep compatibility with Python 2, so I have prepared a special branches in tarantool (`ligurio/python3-support-py2`) and test-run (`ligurio/gh-15-support-python-3-wip-py2`) repositories to check that tests passed with Python 2. See CI status in [branch `ligurio/python3-support-py2`](https://github.com/tarantool/tarantool/tree/ligurio/python3-support-py2) in tarantool).

- tarantool `ligurio/python3-support-py2` contains all commits from `ligurio/python3-support` except those that force using python3:
  - "Bump test-run (Python 3 support)"
  - "readme: update package names in installation steps"
  - "rpm: use Python 3 in testing on RPM packaging"
  - "travis: switch to Python 3 version"
  - "test: use Python 3 in httpd.py shebang"
- test-run `ligurio/gh-15-support-python-3-wip-py2` contains all commits from `ligurio/gh-15-support-python-3-wip-py2` with removed commit that changes shebang to Python 3 in `test-run.py`.


Totktonada:
My thoughts regarding the shebang: https://github.com/packpack/packpack-docker-images/pull/68#issuecomment-763953885.

Totktonada (commented):
Nice to see that the Python 3 support is closer to the reality!

I'll look through commits one by one and will place my comments for each commit in it's own GitHub's review. I guess it'll help us to navigate over the discussion (and also will help to deliver first comments earlier).

Don't get me wrong. I will not insist on fixing every line and every comment according to my preferences. No. However I want to *share* preferences and thoughts during review to learn something new from each other.

## 1/23 Explicit imports of needed objects

So called 'star import' (the first link) is irrelevant here. `import foo` is not a star import. 'Import everything from a module' is not the same as 'import a module'.

The second link refers the question, where the accepted answer does not state that this approach 'allows to keep code clean for readers'. The answer author prefers the opposite (at least in the reduced context of their discussion -- regarding `sys`) and explicitly said that it is the code style question. I would treat it as that there is no general rule of thumb (and brief looking around relevant SO discussions confirms this).

I guess you personally prefer this style and it is okay. Don't hesitate to declare your vision as the reason of the change.

What regarding me... I don't remember why exactly I did this change in the draft. I don't have a personal preference, but it seems that `from foo import bar` is the preferred way within the test-run's code. Improving consistency looks as enough motivation for me in the scope of this project.

One more note: I tends to mark all commits with `Part of #xxx` even when they're just side effect of looking around code to fix / implement `#xxx` (and it is easier to keep it within the patchset). There are two reasons:

* Git (w/o GitHub) have no any mechanism to mark a consequent set of commits as part of one patch series (I don't count merge commits, because of my preference to keep a history linear).
* I like to read stories and don't like changes, which appears from nothing. When there is a linked issue, we can guess that an author found a bad code during looking into the issue and decided to fix it. That's the nice storyline.

Too much text, sorry.

Totktonada (commented):
## 02/23 Remove unused variable

I tested it. flake8 complains regarding this line when called from Python 3, but does not complain when called from Python 2. So it is certainly part of #20.

My flake8 versions:

```sh
$ flake8 --version  # Python 2, flake8 from pip
3.8.4 (mccabe: 0.6.1, pycodestyle: 2.6.0, pyflakes: 2.2.0) CPython 2.7.18 on
Linux
$ /usr/bin/flake8 --version  # Python 3, flake8 from portage
3.8.4 (mccabe: 0.6.1, pycodestyle: 2.6.0, pyflakes: 2.2.0) CPython 3.9.1 on Linux
```

LGTM otherwise.

## 03/23 python3: make print calls compatible with Python 3.x

LGTM.

Totktonada (commented):
## 04/23 python3: fix path import lib/box_connection.py

This change is unclear for me: see comments below for details.

I see that I'm co-author here, so maybe the reason is the content of my [experimental patch](https://github.com/tarantool/test-run/commit/2887259844431008bf868049ff564322010b841e). But this fact alone does not advocate the change anyhow. Sorry if I unintentionally directed you to a wrong way.

lib/pytap13.py
```diff
@@ -17,10 +17,13 @@
...
 import re
+
 try:
-    from CStringIO import StringIO
-except ImportError:
+    # Python 2
     from StringIO import StringIO
+except ImportError:
+    # Python 3
+    from io import StringIO
```
Totktonada, line 26: This looks okay for me.

I didn't know all details and had questions. I summarized questions, answers, details and possible implementatios below.

----

The main questions were the following:

1. Why we need `cStringIO` and fallback to `StringIO`? Should not we save this logic on Python 2?
2. Whether we should use `io.StringIO` or `io.BytesIO` on Python 3?
3. Can we use the same `io.StringIO` or `io.BytesIO` on Python 2?

Answers:

1. For performance matters. Does not matter much here (especially on Python 2 -- not our primary target).
2. The former (`io.StringIO`).
3. Seems so, it should be supported on Python 2.6+.

----

A brief list of facts.

Python 2 facts ([source 1][1], [source 2][2], [source 3][3], [source 4][4], [source 5][5]):

1. `cStringIO.StringIO` is the fast implementation of a string buffer with file-like API. It does not accept a unicode string and have other restrictions (they are not interesting for us here). It is stated as 'optional' and there are suggestions to fallback to `StringIO.StringIO` (see sources 3 and 4; don't know more regarding variety across imterpreter implementations, builds and versions).
2. `StringIO.StringIO` (without `c`) is mostly the same, but supports unicode strings. This, however, is declared as less performant alternative. So we prefer `cStringIO`.
3. `io.StringIO` accepts only unicode strings. It does not accept `str`.
4. `io.BytesIO` accepts only binary strings. It does not accept `unicode`.

Python 3 facts:

1. Now there are two buffer implementations with file-like API: `io.BytesIO` for byte strings and `io.StringIO` for unicode string. (And they both supported in Python 2.6+.)

test-run facts (this particular snippet of the code):

1. There is the typo: `CStringIO` should be `cStringIO`. It is hidden due to the fallback to `StringIO`.
2. When working on Python 2 we're not insterested in decoding from UTF-8 to a unicode string representation and back to UTF-8. We have no problem introduced in Python 3: all string literals becomes unicode and so working with byte strings requires extra attention.
3. Moreover, the only exchange between the `pytap13.py` file and all code outside is accepting a string as input in the `parse()` method.
4. As I see on the top commit on your branch (`open(self.tmp_result, 'r')`), we pass `str` (a binary string) as input on Python 2 and, hmm, `str` (a unicode string) as input on Python 3. Okay.
5. We'll anyway require more weird changes in order to work with both binary and unicode strings inputs: at least have two variants of compiled regexps on Python 3. So let's target only binary strings on Python 2 and only unicode strings on Python 3.

Given those facts we can save the `cStringIO` (and fix `C` to `c`)  usage for performance on Python 2 (but why?):

```python
try:
    # Python 2.
    try:
        # Fast implementation (supports only binary strings).
        from cStringIO import StringIO
    except ImportError:
        # Fallback (supports both string types, but we use it for
        # binary strings).
        from StringIO import StringIO
except:
    # Python 3 (unicode strings).
    from io import StringIO
```

Or we can move to `io.BytesIO` and `io.StringIO` (Python 2.6+):

```python
if sys.version_info[0] == 3:
    # Python 3. Accept a unicode string as input.
    from io import StringIO
else:
    # Python 2. Accept a binary string as input.
    from io import BytesIO as StringIO
```

It looks more straightforward. It'll not raise questions 'hey, Python 2 compatible code should prefer `cStringIO`'. It is a bit weird to rename BytesIO to StringIO, though. Alternative:

```python
if sys.version_info[0] == 3:
    # Python 3. Accept a unicode string as input.
    from io import StringIO as StringBuffer
else:
    # Python 2. Accept a binary string as input.
    from io import BytesIO as StringBuffer
```

However it introduces the new term 'StringBuffer' that is not nice.

We can use `six`:

```python
from six.moves import cStringIO
```

It'll choose `cStringIO.StringIO` on Python 2 and `io.StringIO` on Python 3. It is the ideal variant.

Alternative:

```python
from six import StringIO
```

It'll choose `StringIO.StringIO` on Python 2 (acceptable) and `io.StringIO` on Python 3 (ok).

----

BTW, interesting thread about a performance problem on Python 3 with `BytesIO`, which was resolved in Python 3.5: https://mail.python.org/pipermail/python-dev/2014-July/135542.html. Is it backported to Python 2? Don't know.

----

To sum up, what I would recommend as the general Python 2 & 3 solution in the (quite common) situation, when you need to process binary strings on Python 2 and unicode strings on Python 3? Use `from six.moves import cStringIO` where possible. If `six` usage is discouraged, use the three-cases snippet above. If performance on Python 2 does not matter, use anything.

If performance on Python 3 is matter, then I don't know. Is it okay to operate on unicode strings so? Maybe it worth to do refactoring and use `BytesIO`?

[1]: https://docs.python.org/2/library/stringio.html
[2]: https://docs.python.org/2/library/io.html#io.StringIO
[3]: https://books.google.ru/books?id=PIlQzfnpn5cC
[4]: http://histo.ucsf.edu/BMS270/diveintopython3-r802.pdf
[5]: https://docs.python.org/2/library/io.html#io.BytesIO

lib/tarantool_server.py
```diff
@@ -42,6 +37,13 @@
 from lib.utils import prefix_each_line
 from test import TestRunGreenlet, TestExecutionError
 
+try:
+    # Python 2
+    from StringIO import StringIO
+except ImportError:
+    # Python 3
+    from io import StringIO
```
Totktonada: I would keep 'platform' imports first and 'site modules' imports next. This is generally followed in the test-run code, but not everywhere. Why do you moved this import downside? Why only this one?

I had guess that it is to work around 'E402 module level import not at top of file' flake8 warning, but it seems it does not complain against imports on this kind.

Totktonada (commented):
## 05/23 python3: fix import of StringIO module

I don't get why do you moved one code block (see below).

I left some details regarding all those buffer implementations, but primarily for myself -- to better understand what would be an ideal solution in a common case like our one.

LGTM.

lib/tarantool_server.py
```diff
@@ -761,7 +761,7 @@ def deploy(self, silent=True, **kwargs):
     def copy_files(self):
         if self.script:
             shutil.copy(self.script, self.script_dst)
-            os.chmod(self.script_dst, 0777)
+            os.chmod(self.script_dst, 0o777)
```
Totktonada, line 775: The change is okay. However the explanation is misleading. It states:

> For Python 3 permission mode must have a prefix with 0o [1].
>
> 1. https://docs.python.org/3/library/os.html

First, it just the octal number syntax that is accepted on both Python 2.6+ and Python 3. It is not a special permission object or kinda.

Second, the link does not explain anything regarding deprecation of the `0777` octal number syntax.

Totktonada (commented):
## 06/23 fix mode format in chmod() call

The change itself is okay, but the explanation is misleading. See below.

dispatcher.py
```diff
@@ -7,7 +7,12 @@
 import yaml
 
 import multiprocessing
-from multiprocessing.queues import SimpleQueue
+try:
+    # Python 3.7+
+    from multiprocessing import SimpleQueue
+except ImportError:
+    # Python 2
+    from multiprocessing.queues import SimpleQueue
```
Totktonada, line 18: I understand, it is from my old experimental [patch](https://github.com/tarantool/test-run/commit/2887259844431008bf868049ff564322010b841e), but, to be honest, I see no sense in this change.

First of all, the 'Python 3.7+' comment is misleading. The `SimpleQueue` class was exposed from the `multiprocessing` module since Python 3.3. See https://bugs.python.org/issue11836.

However it still [present](https://github.com/python/cpython/blob/a6192635f1e62af2bb8a435487ebb51800edd671/Lib/multiprocessing/queues.py#L10) in `multiprocessing.queues` and I don't heard anything regarding removal plans. So the change looks useless.

I didn't test the code, because now (on the first review iteration) I'm only looking around the patches. So, maybe, I don't know about some subtle difference. If there is some kind of this, it worth to mention it in the commit message.

Totktonada (commented):
## 07/24 python3: fix import of SimpleQueue

This change looks useless. See details below.

Totktonada (commented):
## 08/24 python3: fix configparser module import

There are several comments regarding this change, see details below.

Totktonada (commented):
## 09/24 python3: replace xrange() with xrange()

The change is okay (the `size` is small: it is always `3`), but there is a typo in the commit title: the last `xrange()` should be `range()`.

If you want to keep `xrange()` usage on Python 2, that's [easy](https://github.com/msgpack/msgpack-python/blob/cfae52437b0d146f74a14d69cb706712c1fa2c95/msgpack/fallback.py#L18). I would bother in code that tends to be universal, but here it is optional.

Totktonada (commented):
## 10/24 python3: using six to iteritems()

Some code that work on Python 2 and Python 3 was changed. I would not change it if we'll not going to add six-like `iteritems()` and `iterkeys()` functions to `lib/utils.py` (see below).

New `six` usages were added. I would avoid this: please, consider this as my little wish. Since we can easily eliminate the dependency I hope we'll have less problems with RPM / Deb dependencies in tarantool and a few connectors and modules that also use test-run.

There are few `six` usages now and everything that we need is just the following code in `lib/utils.py`:

```python
if sys.version_info[0] == 3:
    string_types = str,
    integer_types = int,
    from shlex import quote as shlex_quote
else:
    string_types = basestring,
    integer_types = (int, long)
    from pipes import quote as shlex_quote
```

Aside of this we need to change `.iteritems()` and `.iterkeys()` to just `.items()` and `.keys()`. Or we can copy the corresponding [functions](https://github.com/benjaminp/six/blob/42636b15dd1a5b85de56eac98e47954d4c776576/six.py#L587-L622) from `six` and use them (if someone really bother about invisible performance aberrations when running test-run on Python 2). In this case the `lib/utils.py` file will gain 16 SLOC instead of 8 SLOC, that's negligible.

I commented four cases with more details. The rest is the same.

lib/pytap13.py
```diff
@@ -176,7 +177,7 @@ def _parse(self, source):
             self.tests.append(Test('not ok', len(self.tests), comment=comment))
 
     def parse(self, source):
-        if isinstance(source, (str, unicode)):
+        if isinstance(source, six.string_types):
```
Totktonada, line 180: It should not be part of the 'python3: using six to iteritems()' commit, because it is not about iterating over a dictionary.

listeners.py
```diff
@@ -1,9 +1,11 @@
+from __future__ import absolute_import
+
```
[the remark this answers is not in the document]
ligurio: moved change to appropriate patch (`python3: use relative imports`)

test-run.py
```diff
@@ -94,16 +97,16 @@ def main_loop_parallel():
                  format(args.no_output_timeout), schema='tr_text')
     color_stdout("\n", schema='tr_text')
 
-    task_groups = lib.worker.get_task_groups()
-    if lib.Options().args.reproduce:
+    task_groups = get_task_groups()
+    if Options().args.reproduce:
         task_groups = lib.worker.reproduce_task_groups(task_groups)
```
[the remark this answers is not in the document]
ligurio: > should be trimmed here too

Agree, fixed.

> flake8 complains against this line.

Yep, there was an error `Module level import not at top of file (E402)` and it was fixed.

test-run.py
```diff
@@ -173,15 +176,15 @@ def main_loop_consistent(failed_test_ids):
             short_status = worker.run_task(task_id)
             if short_status == 'fail':
                 reproduce_file_path = \
-                    lib.worker.get_reproduce_file(worker.name)
+                    worker.get_reproduce_file(worker.name)
```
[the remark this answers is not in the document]
ligurio: Fixed:
```
@@ -175,7 +173,7 @@ def main_loop_consistent(failed_test_ids):
             short_status = worker.run_task(task_id)
             if short_status == 'fail':
                 reproduce_file_path = \
-                    worker.get_reproduce_file(worker.name)
+                    get_reproduce_file(worker.name)
                 color_stdout('Reproduce file %s\n' %
                              reproduce_file_path, schema='error')
                 if show_reproduce_content:
```

dispatcher.py
```diff
@@ -10,7 +10,7 @@
 from multiprocessing.queues import SimpleQueue
 
 import listeners
```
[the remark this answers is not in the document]
ligurio: Fixed:

```python
@@ -9,11 +9,16 @@ import yaml
 import multiprocessing
 from multiprocessing.queues import SimpleQueue
 
-import listeners
 from lib import Options
 from lib.utils import set_fd_cloexec
 from lib.worker import WorkerTaskResult, WorkerDone
 from lib.colorer import color_stdout
+from listeners import ArtifactsWatcher
+from listeners import FailWatcher
+from listeners import HangWatcher
+from listeners import LogOutputWatcher
+from listeners import OutputWatcher
+from listeners import StatisticsWatcher

 class TcpPortDispatcher:
@@ -130,20 +135,20 @@ class Dispatcher:
             not args.valgrind
         watch_fail = not Options().args.is_force
 
-        log_output_watcher = listeners.LogOutputWatcher()
-        self.statistics = listeners.StatisticsWatcher(
+        log_output_watcher = LogOutputWatcher()
+        self.statistics = StatisticsWatcher(
             log_output_watcher.get_logfile)
-        self.artifacts = listeners.ArtifactsWatcher(
+        self.artifacts = ArtifactsWatcher(
             log_output_watcher.get_logfile)
-        output_watcher = listeners.OutputWatcher()
+        output_watcher = OutputWatcher()
         self.listeners = [self.statistics, log_output_watcher, output_watcher, self.artifacts]
         if watch_fail:
-            self.fail_watcher = listeners.FailWatcher(
+            self.fail_watcher = FailWatcher(
                 self.terminate_all_workers)
             self.listeners.append(self.fail_watcher)
         if watch_hang:
             warn_timeout = 60.0 if args.long else 10.0
-            hang_watcher = listeners.HangWatcher(
+            hang_watcher = HangWatcher(
                 output_watcher.not_done_worker_ids, self.kill_all_workers,
                 warn_timeout, float(args.no_output_timeout))
             self.listeners.append(hang_watcher)

@@ -315,8 +320,8 @@ class Dispatcher:
         # leave only output listeners in self.listeners
         new_listeners = []
         for listener in self.listeners:
-            if isinstance(listener, (listeners.LogOutputWatcher,
-                                     listeners.OutputWatcher)):
+            if isinstance(listener, (LogOutputWatcher,
+                                     OutputWatcher)):
                 listener.report_at_timeout = False
                 new_listeners.append(listener)
         self.listeners = new_listeners
```

lib/box_connection.py
```diff
@@ -21,22 +21,31 @@
...
 from tarantool import Schema                        # noqa: E402
 
+from tarantool_connection import TarantoolConnection
+from lib.utils import check_libs
 
 SEPARATOR = '\n'
 
+# monkey patch tarantool and msgpack
+check_libs()
+
```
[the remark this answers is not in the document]
ligurio: You are right, there is no problem that this patch solves.

lib/tarantool_server.py
ligurio: Updated commit message:
```
    python3: fix mode format in chmod() call
    
    flake8 warn about SyntaxError E999: "leading zeros in decimal integer
    literals are not permitted; use an 0o prefix for octal integers".
    Patch fixes an error.
```

dispatcher.py
ligurio: Without patch I got an error on start `test-run.py`:
```python
  File "../../test/test-run.py", line 233, in <module>
    status = main_parallel()
  File "../../test/test-run.py", line 148, in main_parallel
    res = main_loop_parallel()
  File "../../test/test-run.py", line 106, in main_loop_parallel
    dispatcher = Dispatcher(task_groups, jobs, randomize)
  File "/home/sergeyb/sources/MRG/tarantool/test-run/dispatcher.py", line 105, in __init__
    task_queue_disp = TaskQueueDispatcher(key, task_group, randomize)
  File "/home/sergeyb/sources/MRG/tarantool/test-run/dispatcher.py", line 378, in __init__
    self.result_queue = SimpleQueue()
TypeError: __init__() missing 1 required keyword-only argument: 'ctx'
```
There is a solution in a [Python bugtracker](https://bugs.python.org/msg230716):
```python
try:
    from multiprocessing import SimpleQueue
except ImportError:
    from multiprocessing.queues import SimpleQueue
```

lib/test_suite.py
```diff
@@ -1,4 +1,16 @@
-import ConfigParser
+from __future__ import print_function
+from __future__ import absolute_import
```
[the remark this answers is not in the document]
ligurio: First one import removed, second one moved to `python3: use relative imports`.

lib/test_suite.py
```diff
@@ -1,4 +1,16 @@
...
+
+from six.moves import configparser
+import six
+
+if six.PY2:
+    # Python2
+    ConfigParser = configparser.SafeConfigParser
+else:
+    # Python3
+    ConfigParser = configparser.ConfigParser
```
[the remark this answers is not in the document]
ligurio: removed dead code with patch:
```python
@@ -1,12 +1,4 @@
 from six.moves import configparser
-import six
-
-if six.PY2:
-    # Python2
-    ConfigParser = configparser.SafeConfigParser
-else:
-    # Python3
-    ConfigParser = configparser.ConfigParser
 
 import json
 import os
```

lib/test_suite.py
```diff
@@ -1,4 +1,16 @@
-import ConfigParser
+from __future__ import print_function
+from __future__ import absolute_import
+
+from six.moves import configparser
+import six
+
+if six.PY2:
+    # Python2
+    ConfigParser = configparser.SafeConfigParser
```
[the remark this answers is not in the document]
ligurio: Code has been removed because it was unused. Problem solved with `from six.moves import configparser`

lib/test_suite.py
```diff
@@ -1,4 +1,16 @@
-import ConfigParser
+from __future__ import print_function
+from __future__ import absolute_import
+
+from six.moves import configparser
```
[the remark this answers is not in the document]
ligurio: Well, I don't mind regarding solution, both are ok for me. Rewrote a patch and new version uses different imports for Python 2 and Python 3.

```python
--- a/lib/test_suite.py
+++ b/lib/test_suite.py
@@ -1,4 +1,10 @@
-import ConfigParser
+try:
+    # Python 3
+    import configparser
+except ImportError:
+    # Python 2
+    import ConfigParser as configparser
+
 import json
 import os
 import re
@@ -90,7 +96,7 @@ class TestSuite:
             raise RuntimeError("Suite %s doesn't exist" % repr(suite_path))
 
         # read the suite config
-        config = ConfigParser.ConfigParser()
+        config = configparser.ConfigParser()
         config.read(os.path.join(suite_path, "suite.ini"))
         self.ini.update(dict(config.items("default")))
         self.ini.update(self.args.__dict__)
```
Regarding using `six`: yes, probably it's good to drop it from dependencies list, but benefits are not clear for me. Anyway we have other dependencies in `requirements.txt`...

lib/test_suite.py
```diff
@@ -90,7 +102,7 @@ def __init__(self, suite_path, args):
             raise RuntimeError("Suite %s doesn't exist" % repr(suite_path))
 
         # read the suite config
-        config = ConfigParser.ConfigParser()
+        config = configparser.ConfigParser()
```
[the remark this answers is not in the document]
ligurio: Added both options to ConfigParser() when running under Python 3 and updated commit message.

lib/test_suite.py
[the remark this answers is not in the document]
ligurio: BTW below you highlighted a good reason to reduce dependencies: `Since we can easily eliminate the dependency I hope we'll have less problems with RPM / Deb dependencies in tarantool and a few connectors and modules that also use test-run.`.

listeners.py
```diff
@@ -225,7 +226,7 @@ def process_result(self, obj):
             self.buffer[obj.worker_id] = bufferized + obj.output
 
     def not_done_worker_ids(self):
-        return self.buffer.keys()
+        return list(self.buffer.keys())
```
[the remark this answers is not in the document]
ligurio: Well, reverted change.

dispatcher.py
```diff
@@ -90,7 +91,7 @@ def __init__(self, task_groups, max_workers_cnt, randomize):
 
         tasks_cnt = 0
         self.task_queue_disps = dict()
-        for key, task_group in task_groups.items():
+        for key, task_group in six.iteritems(task_groups):
```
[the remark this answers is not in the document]
ligurio: reverted to `items()`

lib/app_server.py
```diff
@@ -252,7 +253,7 @@ def is_correct(run):
                     test_suite.ini,
                     params=params,
                     conf_name=conf_name
-                ) for conf_name, params in runs.iteritems()
+                ) for conf_name, params in six.iteritems(runs)
```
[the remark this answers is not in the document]
ligurio: reverted to `items()`

lib/pytap13.py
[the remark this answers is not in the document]
ligurio: `iterkeys()` converted to `list()`, it works fine in Python 2 and Python 3:
```
Python 2.7.18 (default, Aug  4 2020, 11:16:42) 
[GCC 9.3.0] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> d = {'a': 2, 'b': 4}
>>> d.keys()
['a', 'b']
>>> list(d)
['a', 'b']
>>> 
sergeyb@pony:~/sources/MRG/tarantool/test-run$ python3
Python 3.8.5 (default, Jul 28 2020, 12:59:40) 
[GCC 9.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> d = {'a': 2, 'b': 4}
>>> d.keys()
dict_keys(['a', 'b'])
>>> list(d)
['a', 'b']
>>> 
```

ligurio: replaced to `string_types` imported from `lib/utils.py`

test-run.py
```diff
@@ -54,13 +53,17 @@
 import sys
 import time
 
-import lib
-import lib.worker
+sys.path.append(os.path.dirname(os.path.realpath(__file__)))
```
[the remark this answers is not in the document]
Totktonada: I don't see this line anymore in the commit, so looks resolved.

lib/box_connection.py
```diff
@@ -21,22 +21,31 @@
...
-from lib.utils import check_libs
-check_libs()
+sys.path.append(os.path.join(
+    os.path.dirname(os.path.realpath(__file__)),
+    'msgpack-python'))
 
 from tarantool import Connection as tnt_connection  # noqa: E402
 from tarantool import Schema                        # noqa: E402
 
+from tarantool_connection import TarantoolConnection
```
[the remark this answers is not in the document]
Totktonada: Since the commit was removed at all, I'll mark this comment as resolved.

dispatcher.py
Totktonada: Okay, got it. However the comment 'Python 3.7+' still incorrect. Can I propose a wording?

```python
# SimpleQueue is available from multiprocessing.queues on
# all Python versions known at the moment of writting the code
# (up to 3.9).
#
# It was additionally exposed directly from the multiprocessing
# module since Python 3.3 ([1]).
#
# However the mandatory argument 'ctx'
# (see multiprocessing.get_context()) was added to the constructor
# of SimpleQueue from multiprocessing.queues since Python 3.4
# ([2]).
#
# So we should import SimpleQueue from multiprocessing on
# Python 3.3+ (and must to do so on Python 3.4+) to uniformly
# instantiate it (without constructor arguments).
#
# [1]: https://bugs.python.org/issue11836
# [2]: https://bugs.python.org/issue18999
try:
    # Python 3.3+
    from multiprocessing import SimpleQueue
except ImportError:
    # Python 3.2 and below (including Python 2)
    from multiprocessing.queues import SimpleQueue
```

lib/test_suite.py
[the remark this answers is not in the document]
Totktonada: You had added motivation to use the `SafeConfigParser` into the commit message and use `ConfigParser` in the code.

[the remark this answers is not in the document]
Totktonada: Ouch, nope, it is regarding `strict=True`, sorry. Everything is consistent here.

lib/test_suite.py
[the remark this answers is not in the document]
Totktonada: I'll cite the new code:

```python
        # read the suite config
        packer_kwargs = dict()
        packer_kwargs['inline_comment_prefixes='] = ';' 
        packer_kwargs['strict='] = True
        try:
            # Python 3.2+
            config = configparser.ConfigParser(**packer_kwargs)
        except TypeError:
            # Python 2
            config = configparser.ConfigParser()
```

It is equivalent to `config = configparser.ConfigParser()`. Details:

1. `packer_` prefix is copy-pasted from the tarantool-python code, where keyword arguments for an msgpack packer instance keyword arguments. Here it has no sense.
2. `'='` is not part of an option name. There is no `inline_comment_prefixes=` option. That is why you got the `TypeError` even on Python 3.
3. According to the documentation the `inline_comment_prefixes` should be a tuple, e.g. `(';',)`.
4. You don't need to do any exception handling. Just add the Python 3 options into the dictionary when you're on Python 3. This is why the dictionary is needed: to pass options conditionally.

I have a little suggestion, which helps me to avoid most of mistakes of this kind. I often test code snippets in REPL when I'm in doubt regarding precise option names/types/format and so on. For example I catched the problem with multiline options (I mentioned it in the previous comment) this way.

lib/pytap13.py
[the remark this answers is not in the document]
Totktonada: It seems, your comment is about some other change. Changes re `string_types` / `integer_types` were extracted to another commit, so I'll resolve this discussion.

If you're about iterating by a dictionary keys using `for key in list(<dictionary>)` or checking of a key presence in a dictionary using `if <key> in list(<dictionary>)`, I can say that `if len(str(bool(<condition>))) % 2 == 0` works for a condition checking too. So what?

The usual suggestion is to iterate using `for key in <dictionary>` and check presence using `if <key> in <dictionary>`.

Totktonada:
My apologizes for the long delay!

Most of the commits are on me to review. The overall status is below.

----

* 01/23 Remove unused variable: **LGTM**.
* 02/23 Explicit imports of needed objects:
  - The code is okay now, but the explanation is not relevant to the code. I explained the problem in my initial comment regarding the commit.
* 03/23 python3: replace xrange() with range(): **LGTM**.
* 04/23 python3: make print calls compatible with Python 3.x: **LGTM**.
* 05/23 python3: fix mode format in chmod() call: **LGTM**.
* 06/23 python3: integer and string types: to be reviewed.
* 07/23 python3: replace execfile() with exec() and compile(): **LGTM**.
* 08/23 python3: fix import of StringIO module:
  - LGTM, but a small question regarding a code jiggle still here. See my initial comment.
* 09/23 python3: fix import of SimpleQueue:
  - The code is okay (I was not right), but a comment in the code is incorrect.
* 10/23 python3: fix configparser module import:
  - I looked at the new code and there are some comments to solve (added above).
* 11/23 python3: conversion iteritems() -> items(): **LGTM**.
* 12/23 python3: decode to utf8: to be reviewed.
* 13/23 python3: use relative imports: to be reviewed.
* 14/23 python3: rename func_name to \_\_name\_\_: **LGTM**.
* 15/23 python3: convert dict keys to a list: to be reviewed.
* 16/23 python3: convert map object to a list: to be reviewed.
* 17/23 python3: use string byte literal:
  - LGTM, but I would mention what was broken and becomes fixed, e.g. 'fix tarantool connection liveliness check'.
* 18/23 python3: use byte representation of test output: to be reviewed.
* 19/23 python3: use python 3 in test-run.py's shebang
  - **LGTM**. You know, I would [choose](https://github.com/packpack/packpack-docker-images/pull/68#issuecomment-763953885) another way, but this one is acceptable too.
 * 20/23 python3: bump gevent version: to be reviewed.
 * 21/23 python3: fix multiprocessing module bug with serialization: to be reviewed.
 * 22/23 travis: switch to python 3: to be reviewed.
 * 23/23 lib: make path to unit test absolute: to be reviewed.

On Sergey:

* 02/23 Explicit imports of needed objects
* 09/23 python3: fix import of SimpleQueue
* 10/23 python3: fix configparser module import

Approved, but with minor wishes (on Sergey too):

* 08/23 python3: fix import of StringIO module
* 17/23 python3: use string byte literal

On me:

* 06/23 python3: integer and string types
* 12/23 python3: decode to utf8
* 13/23 python3: use relative imports
* 15/23 python3: convert dict keys to a list
* 16/23 python3: convert map object to a list
* 18/23 python3: use byte representation of test output
* 20/23 python3: bump gevent version
* 21/23 python3: fix multiprocessing module bug with serialization
* 22/23 travis: switch to python 3: to be reviewed.
* 23/23 lib: make path to unit test absolute: to be reviewed.

Totktonada (commented):
## 06/23 python3: integer and string types

There are several comments regarding the code, but the idea is okay.

lib/admin_connection.py
```diff
@@ -28,6 +28,9 @@
 from tarantool_connection import TarantoolPool
 from tarantool_connection import TarantoolAsyncConnection
 
+if sys.version[0] == '2':
+    reload(sys)         # noqa: F821
+    sys.setdefaultencoding('utf8')
```
Totktonada, line 36: We both know that it is 'a bad trick' and that it is not guaranteed to be reliable. But I want to discuss cons and pros in practice.

First, what problem it solves? Everything was good on Python 2. I guess it is mostly to call `.encode()` and `.decode()` both on `str` and `unicode` both and don't be hit by `UnicodeDecodeError`. A kind of interoperability between `str` and `unicode`. It'll make not only `'abc' == u'abc'` being `True`, but also `'абв' == u'абв'`.

What disquets me on the practical side: dictionaries do not work this way:

```python
>>> {'blablabla': True}.get(u'blablabla', False)
True
>>> import sys
>>> reload(sys)
<module 'sys' (built-in)>
>>> sys.setdefaultencoding('utf-8')
>>> {'ляляля': True}.get(u'ляляля', False)
False
```

So if we'll not take care whether a data is `str` / `unicode`, we can receive hard to debug problems, where an expected value is not found in a dictionary.

What disquiets me on the theoretical side: it is neither in the spirit on Python 2, nor Python 3. The latter is even more strict:

- `<bytes>.decode()` and `<str>.encode()`, not vice versa
- `b'abc' != 'abc'`
- `{b'foo': True}.get('foo', False)` is `False`.

It seems meaningful to going toward strictness of Python 3, not the opposite.

Okay, what I propose? It should not be something large and vague like 'analyze all data flows in the code and propose a refactoring plan'.

I see that we have not so much places, where `.encode()` / `.decode()` were added. So we can create wrappers like:

```python
def as_text(s):
    """ Convert the binary string to a 'native string'.

        It converts the argument to a unicode string on Python 3
        and just return it without changes on Python 2.

        The function performs argument validation: proper use
        is to pass a binary string as input (<bytes> on Python 3
        and <str> on Python 2).
    """

    if type(s) != bytes:
        raise ValueError('Internal error: expected {}, got {}: {}'.format(
            str(bytes), str(type(s))), repl(s))

    if PY2:
        return s

    return s.decode('utf-8')


# If there is a need.
def as_bytes(b):
    <...>

def open_as_text(name, mode='r', buffering=-1):
    """ Open a file as text.

        Reading from the resulting stream gives 'native strings'
        on both Python versions (<str> type). On Python 2 it is
        a binary string, on Python 3 it is a unicode string.
    """
    if PY2:
        return open(name, mode, buffering)

    return open(name, mode, buffering, encoding='utf-8')
```

And use them when we need a 'native string' (`<str>`) or a stream of such strings.

So we'll keep data types on Python 2 unchanged, while Python 3 data flow will be more explicit and easier to follow.

Not sure, though. I need to experiment around this proposal to say something for sure.

Totktonada (commented):
## 12/23 python3: decode to utf8

Not a real review, just general thoughts around.

I think one of us (likely me) should experiment around the proposal below to understand how easy it would be to keep data flow `str` type centric.

I'll pass over for now, review other commits and return back here.

lib/__init__.py
```diff
@@ -1,11 +1,13 @@
...
 import shutil
 
-from lib.options import Options
-from lib.tarantool_server import TarantoolServer
-from lib.unittest_server import UnittestServer
-from utils import warn_unix_sockets_at_start
+from .options import Options
+from .tarantool_server import TarantoolServer
+from .unittest_server import UnittestServer
+from .utils import warn_unix_sockets_at_start
```
Totktonada, line 10: This line is actually becomes broken on Python 3 and there are two ways to fix. Let's looks at cons and pros of both approaches.

## from lib.utils import warn_unix_sockets_at_start

Pros:

1. Consistent with existing code.
2. Minimal change (only broken lines will be changed).
3. It follows majority of the existing code (we already use `from lib.foo import bar` in `lib/*.py` in the vast majority of imports).
4. Easy to explain.

Cons:

1. Using relative import would be nice, it looks a bit more natural.

(The patch below may require small tweaks depending on a place in git history, where we intend to apply it.)

<details>
<summary>Patch</summary>

```diff
commit d45edc0b22dfb2280ac7051b102bfc85a9861a49
Author: Alexander Turenko <alexander.turenko@tarantool.org>
Date:   Fri Feb 19 03:32:26 2021 +0300

    python3: get rid of package level imports
    
    Unlike Python 2, Python 3 does not treat `import bar` from a module in a
    package `foo` as importing of `foo.bar`.
    
    Absolute import that is relative to a test-run root (like `import
    lib.utils`) still work, because `sys.path` contains a script directory
    (after symlinks resolution) on both Python versions.
    
    https://www.python.org/dev/peps/pep-0328/

diff --git a/lib/__init__.py b/lib/__init__.py
index c03a6d8..0dffc40 100644
--- a/lib/__init__.py
+++ b/lib/__init__.py
@@ -5,7 +5,7 @@ import shutil
 from lib.options import Options
 from lib.tarantool_server import TarantoolServer
 from lib.unittest_server import UnittestServer
-from utils import warn_unix_sockets_at_start
+from lib.utils import warn_unix_sockets_at_start
 
 
 __all__ = ['Options']
diff --git a/lib/admin_connection.py b/lib/admin_connection.py
index 062d7c7..5551437 100644
--- a/lib/admin_connection.py
+++ b/lib/admin_connection.py
@@ -24,9 +24,9 @@ __author__ = "Konstantin Osipov <kostja.osipov@gmail.com>"
 import re
 import sys
 
-from tarantool_connection import TarantoolConnection
-from tarantool_connection import TarantoolPool
-from tarantool_connection import TarantoolAsyncConnection
+from lib.tarantool_connection import TarantoolConnection
+from lib.tarantool_connection import TarantoolPool
+from lib.tarantool_connection import TarantoolAsyncConnection
 
 if sys.version[0] == '2':
     reload(sys)         # noqa: F821
diff --git a/lib/app_server.py b/lib/app_server.py
index a37ef18..2343a0b 100644
--- a/lib/app_server.py
+++ b/lib/app_server.py
@@ -21,7 +21,7 @@ from lib.utils import find_port
 from lib.utils import format_process
 from lib.utils import signame
 from lib.utils import warn_unix_socket
-from test import TestRunGreenlet, TestExecutionError
+from lib.test import TestRunGreenlet, TestExecutionError
 from threading import Timer
 
 if sys.version[0] == '2':
diff --git a/lib/box_connection.py b/lib/box_connection.py
index dad84a2..9dd212d 100644
--- a/lib/box_connection.py
+++ b/lib/box_connection.py
@@ -25,7 +25,7 @@ import errno
 import ctypes
 import socket
 
-from tarantool_connection import TarantoolConnection
+from lib.tarantool_connection import TarantoolConnection
 
 # monkey patch tarantool and msgpack
 from lib.utils import check_libs
diff --git a/lib/connpool.py b/lib/connpool.py
index 3e9d654..b80ba58 100644
--- a/lib/connpool.py
+++ b/lib/connpool.py
@@ -10,7 +10,7 @@ from collections import deque
 from contextlib import contextmanager
 from functools import wraps
 
-from test import TestRunGreenlet
+from lib.test import TestRunGreenlet
 
 __all__ = ["ConnectionPool", "retry"]
 
diff --git a/lib/tarantool_connection.py b/lib/tarantool_connection.py
index f2a5267..b246d4f 100644
--- a/lib/tarantool_connection.py
+++ b/lib/tarantool_connection.py
@@ -30,10 +30,10 @@ from contextlib import contextmanager
 import gevent
 from gevent import socket as gsocket
 
-from connpool import ConnectionPool
-from test import TestRunGreenlet
-from utils import warn_unix_socket
-from utils import set_fd_cloexec
+from lib.connpool import ConnectionPool
+from lib.test import TestRunGreenlet
+from lib.utils import warn_unix_socket
+from lib.utils import set_fd_cloexec
 
 
 class TarantoolPool(ConnectionPool):
diff --git a/lib/tarantool_server.py b/lib/tarantool_server.py
index 47631df..38a1ab3 100644
--- a/lib/tarantool_server.py
+++ b/lib/tarantool_server.py
@@ -35,7 +35,7 @@ from lib.utils import safe_makedirs
 from lib.utils import signame
 from lib.utils import warn_unix_socket
 from lib.utils import prefix_each_line
-from test import TestRunGreenlet, TestExecutionError
+from lib.test import TestRunGreenlet, TestExecutionError
 
 try:
     # Python 2
diff --git a/lib/test.py b/lib/test.py
index e6192b5..0fc4c02 100644
--- a/lib/test.py
+++ b/lib/test.py
@@ -2,20 +2,20 @@ import filecmp
 import gevent
 import os
 import pprint
-import pytap13
 import re
 import shutil
 import sys
 import traceback
 from functools import partial
 from hashlib import md5
-from utils import safe_makedirs
 
+from lib import pytap13
 from lib import Options
 from lib.colorer import color_stdout
 from lib.utils import non_empty_valgrind_logs
 from lib.utils import print_tail_n
 from lib.utils import print_unidiff as utils_print_unidiff
+from lib.utils import safe_makedirs
 
 if sys.version[0] == '2':
     reload(sys)     # noqa: F821
```
</details>

## from .utils import warn_unix_sockets_at_start

Pros:

1. This way looks a bit more sane and natural.

Cons:

1. All other imports should be changed in the same way for consistency.
2. So the patch will change not only problematic imports, but all, and it'll be harder to spot where a problem actually was.
3. And so we should cleanly describe the problem, alternative solution and a reason why we decide to use relative imports in the commit message.

## My thoughts

Personally I would prefer the minimal change. However I'll accept any way if it is done consistently and the commit message responds to questions 'what was broken' and 'why it is fixed this way'.

lib/admin_connection.py
```diff
@@ -1,3 +1,6 @@
+from __future__ import print_function
```
Totktonada, line 1: Should not be part of this commit.

Totktonada (commented):
## 13/23 python3: use relative imports

There are questions regarding consistency of the resulting code state: when we introduce an approach we should apply it with some system in the mind, e.g. to all applicable places. I unable to express a general rule, that the commit would follow.

<details>
<summary>Background info to read comments below</summary>

You likely already know that all, but when I started the review I didn't know exact mechanics. So I'll describe it briefly just in case.

In comments below assume knowledge how `import lib.foo` and `import foo` works on Python 2 and Python 3 from a `lib/bar.py`. In short:

1. An absolute import is relative to `sys.path` paths (both Python 2 & 3). So `import lib.foo` works from any location in test-run.
2. An absolute import like `import foo` from `lib/bar.py` (to bring `lib/foo.py` code) works on Python 2, because `lib` is a package, but does not work on Python 3, because an absolute import is not relative to a package anymore (considered as clash incurring practice).

Personally I read [how exactly](https://docs.python.org/2/tutorial/modules.html) import works on Python 2 and the 'Abstract' of [PEP-0328](https://www.python.org/dev/peps/pep-0328/). That's enough.
</details>

lib/tarantool_server.py
```diff
@@ -663,7 +663,7 @@ def __init__(self, _ini=None, test_suite=None):
         # or passed through execfile() for PythonTest
         self.current_test = None
         caller_globals = inspect.stack()[1][0].f_globals
-        if 'test_run_current_test' in caller_globals.keys():
+        if 'test_run_current_test' in list(caller_globals):
```
Totktonada, line 666: That's a dictionary object (I verified on both Python 2 and Python 3). `if 'foo' in <dictionary>` is enough. There is no logic in transformation to a list, another dict, a set, a string, some custom class or any other structure to check whether a key is in a dictionary.

In fact, `caller_globals.keys()` works on both Python 2 and Python 3, so this change is just unnecessary.

See also my previous [comment](https://github.com/tarantool/test-run/pull/263#discussion_r575164506) on this topic. 'It works' can advocate so wide variety of meaningless solutions that I cannot consider it as a valuable argument.

lib/test_suite.py
```diff
@@ -185,7 +185,7 @@ def collect_tests(self):
         return self.tests
 
     def get_fragile_list(self):
-        return self.fragile['tests'].keys()
+        return list(self.fragile['tests'])
```
Totktonada, line 188: I would just replace `self.get_fragile_list()` in `stable_tests()` and `fragile_tests()` with `self.fragile['tests']`. So no extra list will be created. However, okay, it is up to you, because `_list` is in the function name.

However, as I see, the change is not necessary: everything works good with `.keys()` on Python 3. So it is better to don't change it.

lib/worker.py
```diff
@@ -287,7 +287,7 @@ def task_get(self, task_queue):
 
     @staticmethod
     def is_joinable(task_queue):
-        return 'task_done' in task_queue.__dict__.keys()
+        return 'task_done' in list(task_queue.__dict__)
```
Totktonada, line 290: Same as above. Don't transform a dict to something else to check a key presence. Use `'task_done' in task_queue.__dict__`. Or, better, leave `.keys()` when it is already there: it works on Python 2 and Python 3 both.

Totktonada (commented):
## 15/23 python3: convert dict keys to a list

Useless change. The code works on Python 2 and Python 3 without this commit. Please, drop the commit.

lib/app_server.py
```diff
@@ -245,8 +245,8 @@ def is_correct(run):
 
         test_names = sorted(glob.glob(os.path.join(suite_path, "*.test.lua")))
         test_names = Server.exclude_tests(test_names, test_suite.args.exclude)
-        test_names = sum(map((lambda x: patterned(x, test_suite.args.tests)),
-                             test_names), [])
+        test_names = sum(list(map((lambda x: patterned(x, test_suite.args.tests)),
+                                  test_names)), [])
```
Totktonada, line 249: Useless change. The existing code works on Python 2 and Python 3.

```python
$ python2
Python 2.7.18 (default, Jan 27 2021, 00:52:11) 
<...>
>>> sum(map((lambda x: [x]), ['a', 'b', 'c']), [])
['a', 'b', 'c']

$ python3
Python 3.9.1 (default, Jan 27 2021, 00:36:29) 
<...>
>>> sum(map((lambda x: [x]), ['a', 'b', 'c']), [])
['a', 'b', 'c']
```

Reverted the commit and verified on test-run too:

```sh
$ git checkout origin/ligurio/gh-15-support-python-3-wip
$ git revert 975335b4d78a94d022deac4378e49e99be66b73f
$ python2 ./test/test-run.py app-tap/msgpack. app/crypto. unit/rtree.
<..all passed..>
$ python3 ./test/test-run.py app-tap/msgpack. app/crypto. unit/rtree.
<..all passed..>
```

Totktonada (commented):
## 16/23 python3: convert map object to a list

Useless change. Please, drop.

lib/test.py
```diff
@@ -83,6 +75,10 @@ def write(self, fragment):
                 self.stream.write(line)
 
     def push_filter(self, pattern, replacement):
+        if sys.version[0] == '3' and isinstance(pattern, str):
+            pattern = pattern.encode('utf-8')
+        if sys.version[0] == '3' and isinstance(replacement, str):
+            replacement = replacement.encode('utf-8')
```
Totktonada, line 81: As I see, we always receive an `str` here as for `pattern` as well as for `replacement`. I verified it on the whole tarantool test collection. It seems, the patch was created for some intermediate version of the patchset, when it was not so. That means that we can remove `isinstance()` checks. If we'll introduce the `as_bytes()` helper, it'll look even more nice:

```python
def push_filter(self, pattern, replacement):
    self.filters.append([as_bytes(pattern), as_bytes(replacement)])
```

Totktonada (commented):
## 18/23 python3: use byte representation of test output

The approach is okay for me. A few minor comments are below.

requirements.txt
```diff
@@ -1,5 +1,5 @@
 PyYAML==5.1
 argparse==1.1
 msgpack-python==0.4.6
-gevent==1.1b5
+gevent==21.1.1
```
Totktonada, line 4: NB: 21.1.2 is [out](http://www.gevent.org/changelog.html).

Totktonada, line 4: The change is okay, but the explanation 'gevent-1.1b5 is incompatible with Mac OS X 15' is wrong. It is not related to Mac OS. gevent 1.2 and below does not support Python 3.7 and above. See https://github.com/gevent/gevent/issues/1297

Totktonada (commented):
## 20/23 python3: bump gevent version

The change is okay, but the explanation requires a change. See below.

lib/test_suite.py
```diff
@@ -125,10 +125,10 @@ def __init__(self, suite_path, args):
...
                 if i in self.ini else dict()
         for i in ["lua_libs"]:
-            self.ini[i] = map(
-                lambda x: os.path.join(suite_path, x),
-                dict.fromkeys(self.ini[i].split())
-                if i in self.ini else dict())
+            lua_libs = self.ini.get(i, "")
+            self.ini[i] = []
+            for p in lua_libs.split():
+                self.ini[i].append(os.path.join(suite_path, p))
```
Totktonada, line 131: It is not necessary if we'll use 'fork' multiprocessing start method. However, it seems, it is better to keep 'fork' (if'll leave my investigations in another comment).

Totktonada, line 131: Nitpicking (sorry). There is a kind of inaccuracy: the loop body is supposed to be option unaware, but it uses certain option name (`lua_libs`) inside. Since we're here I would rewrite it with proper names and leave a comment. So Une may just add an option of the same format here.

```python
# For given options in the following format:
#   
#  | key = v1 v2
#   
# Set self.ini[key] to the following value:
# ['<suite_name>/v1', '<suite_name>/v2'].
for option_name in ["lua_libs"]:
    option_raw_value = self.ini.get(option_name, "") 
    self.ini[option_name] = []
    for item in option_raw_value.split():
        self.ini[option_name].append(os.path.join(suite_path, item))
```

However you can leave it unchanged due to usage of 'fork' start method. Even if you'll apply it, it'll be just refactoring and should not be part of this patchset. So it is just note, ignore it.

dispatcher.py
```diff
@@ -86,6 +88,13 @@ class Dispatcher:
     ```
     """
     def __init__(self, task_groups, max_workers_cnt, randomize):
+        if platform == "darwin":
+            try:
+                # Python 3
+                multiprocessing.set_start_method('fork')
+            except AttributeError:
+                # Python 2
+                pass
```
Totktonada, line 97: I tried 'spawn' start method on Linux and was hit by 'No such file or directory' from `LogOutputWatcher.process_result()`. Sure, the `test/var/log` directory that is created in the costructor was removed. I found that it is because of reevaluation of the `lib/__init__.py` content in the child process (it removes `test/var`).

So 'spawn' causes re-execution of code in child processes and it is not easy (at least for me, at least now) to say what else will be executed in each process in addition to execution in the main process. We don't expect such behaviour: the code it not ready for that.

So I vote to keep 'fork' everywhere.

dispatcher.py
```diff
@@ -6,6 +6,8 @@
 import functools
 import yaml
 
+from sys import platform
```
Totktonada, line 9: We follow `import foo` for the standard modules across the code.

dispatcher.py
```diff
@@ -6,6 +6,8 @@
 import functools
```
Totktonada, line 6: Regarding commit message:

1. 'fix multiprocessing module bug with serialization' is not what we're doing here. We set 'fork' start method (and the change re lambda elimination becomes unnecessary and should be removed).
2. Why `Part of #265`? It seems, it'll be resolved by using the 'fork' start method. And we unlikely will investigate how to adapt code to be ready for using 'spawn'.
3. Details are desirable:
    - 'spawn' becomes default on Mac OS and we need to set 'fork' explicitly (this can be included right into the code).
    - Why 'spawn' method does not work for us (you can use my investigation from another comment or, if you know another critical problem, use it to advocate the change); in fact, this also can be included right into the code.

Totktonada (commented):
## 21/23 python3: fix multiprocessing module bug with serialization

I vote up the direction you choose: using of 'fork' start method everywhere. Some code jiggle and comments are desirable, however.

.travis.yml
```diff
@@ -1,14 +1,21 @@
 language: python
-python: 2.7
+
+python:
+  - "3.4"
+  - "3.5"
+  - "3.6"
+  - "3.7"
+  - "3.8"
+  - "3.8-dev"
```
Totktonada, line 9: I would keep `2.7` too, however, up to you. Now flake8 is passed on both Python versions (verified locally).

.travis.yml
```diff
@@ -1,14 +1,21 @@
...
 before_install:
   - sudo apt update -y
   - sudo apt-get -y install lua5.1 luarocks
   - sudo luarocks install luacheck
 
 install:
-  - pip install -r requirements.txt
-  - pip install -r requirements-test.txt
+  - pip3 install -r requirements.txt
+  - pip3 install -r requirements-test.txt
```
Totktonada, line 18: Is `pip3` actually necessary? Whether that means that just `pip` does not correspond to the Python version from the matrix above? We use just `python` in the `Makefile`, not `python3`, and I guess it works. So using of `pip3` looks lopsided.

I unable to verify, Travis CI is gone. So, if there is a reason to the change, I'm okay.

Totktonada (commented):
## 22/23 travis: switch to python 3

I'm okay in general, but a couple of questions are unclear for me, see below.

lib/unittest_server.py
```diff
@@ -17,6 +17,7 @@ def __init__(self, *args, **kwargs):
     def execute(self, server):
         server.current_test = self
         execs = server.prepare_args()
+        execs[0] = os.path.abspath(execs[0])
```
Totktonada, line 20: I looked over `TarantoolServer` and `AppServer` and construct an absolute path in `prepare_args()`. So I[ll follow this approach here too:

```diff
diff --git a/lib/unittest_server.py b/lib/unittest_server.py
index 9d57558..5770941 100644
--- a/lib/unittest_server.py
+++ b/lib/unittest_server.py
@@ -50,7 +50,7 @@ class UnittestServer(Server):
     def prepare_args(self, args=[]):
         executable_path = os.path.join(self.builddir, "test",
                                        self.current_test.name)
-        return [executable_path] + args
+        return [os.path.abspath(executable_path)] + args
 
     def deploy(self, vardir=None, silent=True, wait=True):
         self.vardir = vardir
```

However you way is okay too. The decision is up to you.

Totktonada, line 20: It would be nice to include a rationale into the commit message. The documentation for Python 2 and Python 3 explicitly describes the different behaviour:

From https://docs.python.org/2/library/subprocess.html#subprocess.Popen:

> If cwd is not None, the child’s current directory will be changed to cwd before it is executed. Note that this directory is not considered when searching the executable, so you can’t specify the program’s path relative to cwd.

From https://docs.python.org/3/library/subprocess.html#subprocess.Popen:

> If cwd is not None, the function changes the working directory to cwd before executing the child. <...> In particular, the function looks for executable (or for the first item in args) relative to cwd if the executable path is a relative path.

Totktonada (commented):
## 23/23 lib: make path to unit test absolute

LGTM with a few comments (they are up to you).

Totktonada:
I looked over all patches now. However I'll keep it on me for a while to experiment around encode/decode. I'll either prove that my proposal regarding elimination of `setdefaultencoding()` and using of wrappers such as `as_text()` and `as_binary()` is workable or will consider it as 'would be good in theory, but unreachable in practice'.

lib/admin_connection.py
Totktonada: I experimented around strings and bytes and the idea with sticking to a 'native string' (`<str>`) in APIs seems to work pretty good. See 17ca0888e0107de6b295e86ba8402c5895220670. It follows the Python 3 idea about decoupling binary and text data, so the code becomes easier to follow.

Please, note: I used the import style from the surrounding code (relative), but I would prefer to keep it absolute (reasons in the corresponding review comment). If you'll follow my proposal regarding imports, it should be changed here as well.

lib/admin_connection.py
```diff
@@ -39,7 +42,8 @@ def get_handshake(sock, length=128, max_try=100):
     result = ""
     i = 0
     while len(result) != length and i < max_try:
-        result = "%s%s" % (result, sock.recv(length-len(result)))
+        recv = sock.recv(length-len(result)).decode('utf-8')
+        result = "%s%s" % (result, recv)
```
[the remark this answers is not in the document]
Totktonada: In 17ca0888e0107de6b295e86ba8402c5895220670 I kept `bytes` to `str` conversion here for the unification sake: almost all our APIs operate on strings: not `bytes`, not `unicode`.

Totktonada:
I propose 17ca0888e0107de6b295e86ba8402c5895220670 with decoupling binary data from text data. It replaces the following commits:

* 12/23 python3: decode to utf8
* 18/23 python3: use byte representation of test output

The review iteration is finished.

----

(I restricted myself from abstract reasoning during the review, but I'll leave several words now.)

Programming is about finding and using common patterns for me. Pushing code toward simplicity, where possible. Communication with future developers through code, sharing methods and ideas.

That is why someone can say that I spin around insignificant details. I'm trying to show how I would communicate with a reader of my code. How determining of common rules makes resulting code easier.

I can defend all this strictness about things that're not for a machine with the usual words about maintenance cost. That's important, of course. But what is more important: how good we communicate with others using our code. If we're good here, then a project have chances to be alive.

dispatcher.py
ligurio: Replaced `Python 3.7+` with `Python 3.3+` and added wording proposed by you.

lib/test_suite.py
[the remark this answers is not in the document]
ligurio: Thanks for comments. Fixed 1, 2, 3, 4.:

```
--- a/lib/test_suite.py
+++ b/lib/test_suite.py
@@ -97,15 +97,10 @@ class TestSuite:
             raise RuntimeError("Suite %s doesn't exist" % repr(suite_path))
 
         # read the suite config
-        packer_kwargs = dict()
-        packer_kwargs['inline_comment_prefixes='] = ';'
-        packer_kwargs['strict='] = True
-        try:
-            # Python 3.2+
-            config = configparser.ConfigParser(**packer_kwargs)
-        except TypeError:
-            # Python 2
-            config = configparser.ConfigParser()
+        parser_kwargs = dict()
+        parser_kwargs['inline_comment_prefixes'] = (';',)
+        parser_kwargs['strict'] = True
+        config = configparser.ConfigParser(**parser_kwargs)
         config.read(os.path.join(suite_path, "suite.ini"))
         self.ini.update(dict(config.items("default")))
         self.ini.update(self.args.__dict__)
```

lib/pytap13.py
ligurio: Reverted block with imports in `lib/pytap13.py`, no sense to change it.

requirements.txt
ligurio: Updated

ligurio: Updated commit message.

dispatcher.py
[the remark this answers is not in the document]
ligurio: Moved `set_start_method` to `test-run.py` script and removed try/except around it.

lib/test_suite.py
ligurio: Agree, it was a part of another attempt to fix. Removed.

dispatcher.py
ligurio: no action is required

dispatcher.py
ligurio: code is removed in updated patch

dispatcher.py
ligurio: 1. Agree. Initially there was another attempt to fix issue.
2. convert to 'Fixes"
3. add wording to commit message and to source code inline

.travis.yml
ligurio: Done, added 2.7.

.travis.yml
ligurio: > Is pip3 actually necessary? 

`pip3` used here to explicitly run it with python3. Some environment may have pip that points to pip2 actually.

ligurio: If you are still confused and not satisfied by comment above I can check it for at least env on test-run CI and replace pip3 to pip if everything well there. But again, our CI is broken :(

lib/unittest_server.py
ligurio: I like you fix :) so I updated a patch.

ligurio: Thanks! Added it to commit message.

lib/pytap13.py
```diff
@@ -23,6 +23,7 @@
     from StringIO import StringIO
 
 import yaml
+from .utils import string_types
```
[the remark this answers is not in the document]
ligurio: What's wrong here? I need to import exactly one function and do it with relative import. Other existed imports are 'system' imports and there is no absolute imports for 'local'  modules.

lib/utils.py
```diff
@@ -21,6 +20,18 @@
 
 UNIX_SOCKET_LEN_LIMIT = 107
 
+# Useful for very coarse version differentiation.
+PY2 = sys.version_info[0] == 2
+PY3 = sys.version_info[0] == 3
+PY34 = sys.version_info[0:2] >= (3, 4)
```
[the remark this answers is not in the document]
ligurio: removed

lib/test_suite.py
```diff
@@ -68,7 +69,7 @@ def parse_bool_opt(self, name, default):
             self.ini[name] = default
         elif isinstance(val, bool):
             pass
-        elif isinstance(val, str) and val.lower() in ('true', 'false'):
+        elif isinstance(val, string_types) and val.lower() in ('true', 'false'):
```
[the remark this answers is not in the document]
ligurio: revert changes for this file

lib/pytap13.py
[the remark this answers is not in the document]
Totktonada: It is about consistency with other imports from `lib.utils`. There are two variants:

1. `from lib.utils import string_types`
2. `from .utils import string_types`

Both work, but the former is consistent with all other test-run code. See more information in the review of the '13/23 python3: use relative imports' commit.

.travis.yml
Totktonada: I expect that Travis CI setup correct environement, when `python: [<versions>]` is used. However, yep, there is no simple way to check it, so feel free to skip this comment.

BTW, since you added 2.7, there is another reason to use `pip` here.

ligurio: > BTW, since you added 2.7, there is another reason to use pip here.

Yep, replaced `pip3` to `pip`.

lib/pytap13.py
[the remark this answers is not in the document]
ligurio:
```
--- a/lib/pytap13.py
+++ b/lib/pytap13.py
@@ -23,7 +23,7 @@ except ImportError:
     from StringIO import StringIO
 
 import yaml
-from .utils import string_types
+from lib.utils import string_types
 
 
 RE_VERSION = re.compile(r"^\s*TAP version 13\s*$")
```

Updated.

lib/tarantool_server.py
ligurio: Moved StringIO import before 'site modules' imports.

lib/worker.py
ligurio: reverted change

lib/test_suite.py
ligurio: > However, as I see, the change is not necessary: everything works good with .keys() on Python 3. 

Just reverted change

lib/tarantool_server.py
ligurio: reverted change

lib/box_connection.py
```diff
@@ -1,3 +1,6 @@
+from __future__ import print_function
```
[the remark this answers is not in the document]
ligurio: reverted

lib/tarantool_server.py
```diff
@@ -1,3 +1,6 @@
+from __future__ import print_function
```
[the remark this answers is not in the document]
ligurio: reverted

lib/__init__.py
```diff
@@ -1,11 +1,13 @@
+from __future__ import absolute_import
```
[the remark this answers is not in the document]
ligurio: reverted `import absolute_import` in 18 out of 23 files

lib/admin_connection.py
ligurio: reverted

ligurio:
> I propose 17ca088 with decoupling binary data from text data. It replaces the following commits:
> 
> 12/23 python3: decode to utf8
> 18/23 python3: use byte representation of test output
> The review iteration is finished.

Replaced with a small change - added to commit `PY2` definition in `lib/utils.py`.

lib/admin_connection.py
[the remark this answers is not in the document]
Totktonada: Now the received byted are converted to a native string (unicode on Python 3) at end of the function. It is just to unify test-run functions APIs and use just `str` (almost) everywhere.

lib/admin_connection.py
Totktonada: Sergey included the patch into the patchset and made appropriate changes.

lib/__init__.py
```diff
@@ -1,11 +1,13 @@
...
 import sys
 import shutil
 
-from lib.options import Options
-from lib.tarantool_server import TarantoolServer
-from lib.unittest_server import UnittestServer
-from utils import warn_unix_sockets_at_start
+from .options import Options
+from .tarantool_server import TarantoolServer
+from .unittest_server import UnittestServer
```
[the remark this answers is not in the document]
Totktonada: Now everything looks consistent (at least on the first brief glance).

lib/test.py
```diff
@@ -59,19 +52,18 @@ def __init__(self, filename):
...
 
     def write(self, fragment):
         """Apply all filters, then write result to the undelrying stream.
         Do line-oriented filtering: the fragment doesn't have to represent
         just one line."""
-        if isinstance(fragment, (bytes, bytearray)):
-            fragment = fragment.decode('utf-8')
-        fragment_stream = StringIO(fragment)
+        if sys.version[0] == '3' and isinstance(fragment, str):
+            fragment = fragment.encode('utf-8')
```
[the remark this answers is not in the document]
Totktonada: Now the changes around handling strings and bytes in our output stream become part of the 'python3: decouple bytes and strings' commit. Here we use `str_to_bytes()` (referred as `as_bytes()` previously) in `.write()` and assert that the incoming object is `bytes` in `.write_bytes()`.

lib/test.py
Totktonada: The change is part of the 'python3: decouple bytes and strings' commit now.

lib/test_suite.py
Totktonada: Filed #272 for this.

dispatcher.py
Totktonada: In fact, it is the primary reason of choosing 'fork', not because we need to eliminate one lambda function. I would leave several words regarding this somewhere.

dispatcher.py
Totktonada: Object serialization is not the primary problem. Code reexecution in a child process is the reason.

Totktonada:
> 2. open result file in binary mode and convert line using byte_to_str()

I see, several result files contains characters that are not in ASCII, but in UTF-8:

* test/engine/select.result
* test/sql/collation.result
* test/sql/icu-upper-lower.result

Don't know how I missed it. Thanks!

Totktonada:
Hey, not I know how I missed this :) The fail is due to [non UTF-8 locale](https://github.com/packpack/packpack-docker-images/blob/23ca21bd189e710476a0f8803909f10814bec88e/centos/8/Dockerfile#L16) in our images.

The [commit message](https://github.com/packpack/packpack-docker-images/commit/4a295815a300678c86e2ea03d13a970aa7e39831) ([PR #39](https://github.com/packpack/packpack-docker-images/pull/39)) states the following:

> - at CentOS 8 the localization environment changed to "C";
>    (needed by backport repository);

So I have several questions here:

1. What the exact problem was?
2. Can it be solved by a temporary changing of the environment during the image build?

@avtikhon Can you clarify?

(Anyway, I don't mind the change in test-run.)

dispatcher.py
Totktonada: I'm going to change the comment this way:

```diff
diff --git a/test-run.py b/test-run.py
index ce3558d..63392ff 100755
--- a/test-run.py
+++ b/test-run.py
@@ -219,11 +219,19 @@ def main_consistent():
 
 if __name__ == "__main__":
     # In Python 3 start method 'spawn' in multiprocessing module becomes
-    # default on Mac OS [1]. However 'spawn' method requires object serialization
-    # that doesn't work when objects use lambdas, whose for example used in
-    # class TestSuite (lib/test_suite.py). So this method is not acceptable
-    # for using with test-run. To workaround it can be possible to force
-    # fork mode with multiprocessing.set_start_method('fork').
+    # default on Mac OS [1].
+    #
+    # The 'spawn' method causes re-execution of some code, which is already
+    # executed in the main process. At least it is seen on the
+    # lib/__init__.py code, which removes the 'var' directory. Some other
+    # code may have side effects too, it requires investigation.
+    #
+    # The method also requires object serialization that doesn't work when
+    # objects use lambdas, whose for example used in class TestSuite
+    # (lib/test_suite.py).
+    #
+    # The latter problem is easy to fix, but the former looks more
+    # fundamental. So we stick to the 'fork' method now.
     if PY3:
         multiprocessing.set_start_method('fork')
```

I'll do the same change for the commit message. I hope you don't mind.

Totktonada: And, while I'm here, the tiny change for consistency:

```diff
diff --git a/test-run.py b/test-run.py
index c2af31e..f820272 100755
--- a/test-run.py
+++ b/test-run.py
@@ -55,7 +55,8 @@ import time
 
 from lib import Options
 from lib.colorer import color_stdout
-from lib.utils import print_tail_n, PY3
+from lib.utils import print_tail_n
+from lib.utils import PY3
 from lib.worker import get_task_groups
 from lib.worker import get_reproduce_file
 from lib.worker import reproduce_task_groups
```

dispatcher.py
Totktonada: I'm going to change the wording. See the [comment](https://github.com/tarantool/test-run/pull/263#discussion_r593350504) above.

Totktonada:
> Hey, not I know how I missed this :) The fail is due to [non UTF-8 locale](https://github.com/packpack/packpack-docker-images/blob/23ca21bd189e710476a0f8803909f10814bec88e/centos/8/Dockerfile#L16) in our images.

So I should change the commit message of the 'python3: decouple bytes and strings' commit, because it states the following now:

> We can use unicode or binary streams for files (unicode by default[^1]).
>
>     [^1]: Technically it depends on a system locale, but, hey, does anyone
>           see a non UTF-8 locale after the millennium?

Since we don't lean on this property anymore, we should reflect it here.

Totktonada:
> > Hey, not I know how I missed this :) The fail is due to [non UTF-8 locale](https://github.com/packpack/packpack-docker-images/blob/23ca21bd189e710476a0f8803909f10814bec88e/centos/8/Dockerfile#L16) in our images.
> 
> So I should change the commit message of the 'python3: decouple bytes and strings' commit, because it states the following now:
> 
> > We can use unicode or binary streams for files (unicode by default[^1]).
> > ```
> > [^1]: Technically it depends on a system locale, but, hey, does anyone
> >       see a non UTF-8 locale after the millennium?
> > ```
> 
> Since we don't lean on this property anymore, we should reflect it here.

It appears that the problem is not only here. We read test files in the same way. We read result files when a hung worker is detected, which is not quite usual situation, so it'll not be covered by our usual CI run in tarantool.

I decided to work the problem around for all text streams uniformly. So, I reverted the change from the 'python3: decouple bytes and strings' commit and slightly changed its wording:

> [^1]: Technically it depends on a Python version and a system locale.
>       There are situations, when default text file streams encoding is
>       not 'utf-8'. They will be handled in the next commit.

I added one more commit to handle the problem. It is named 'python3: set text file streams encoding to utf-8', I'll push it here soon.

dispatcher.py
Totktonada: Pushed.

Totktonada (approved):
The code looks okay for me now.

I did some testing on Python 2 and Python 3, but not on full set of our platforms. I see that you tested the patchset thoroughly, so I don't worry.

I'll think about better order of pushing relevant test-run and tarantool patches and will proceed.

Totktonada:
Updated the test-run submodule in tarantool in 2.8.0-127-gf203bf33a (part of [PR #5751](https://github.com/tarantool/tarantool/pull/5751)).

Backports to 2.7, 2.6 and 1.10 are in progress.

Totktonada:
2.7 in 2.7.1-116-ga71248988.
2.6 in 2.6.2-113-g7421d0f43.

Totktonada:
1.10 in 1.10.9-75-g949e86b7e.

PR #267, opened 2021-02-18
readme: list all our users

ligurio:


README.md
```diff
@@ -379,3 +379,10 @@ The following files will be removed:
 * `[0-9]*/`
 
 For a `core = unittest` test suite this option does not change any behaviour.
+
+### Used By
+
+- [Tarantool](https://github.com/tarantool/tarantool) - is an in-memory database and application server
+- [tarantool-python](https://github.com/tarantool/tarantool-python) - Python driver for Tarantool
```
Totktonada, line 386: It does not anymore: https://github.com/tarantool/tarantool-python/pull/189

That was just one test case, so don't count it.

README.md
```diff
@@ -379,3 +379,10 @@ The following files will be removed:
 * `[0-9]*/`
 
 For a `core = unittest` test suite this option does not change any behaviour.
+
+### Used By
+
+- [Tarantool](https://github.com/tarantool/tarantool) - is an in-memory database and application server
```
Totktonada, line 385: Nit: I would remove 'is an' for consistency with the vshard description below.

README.md
Totktonada: I applied the change myself to don't bother you. Hope you don't mind.

README.md
Totktonada: I removed 'is an' here.

Totktonada (approved):
LGTM.

I made several minor changed to don't bother you with extra context switch. If you're okay, I'll push the change.

ligurio:
Sure, I'm ok. Thanks!

PR #266, opened 2021-02-10
bump tarantool-python version

ligurio:
commit daff045 ('Add luacheck config and fix warnings found by luacheck')
fixes luacheck warnings in Lua source code. Bumping version
allows to remove test-run directory from exclusions in .luacheckrc in
tarantool repository.

PR #268, opened 2021-02-24
Bump tarantool-python submodule (remove test-run)

Totktonada:
This change removes the test-run submodule from `lib/tarantool-python`.

The idea is to get rid of this 'recursive' nesting of test-run and
tarantool-python submodules into each other. Those nested submodules
contain old resivions of the code. It confuses linters, which looking
for code recursively (such as luacheck). See [1] for details.

After this commit we can eliminate the `test-run/**/*.lua` exclusion
from `.luacheckrc` in tarantool.

As the side effect, `git clone --recursive <...>` will be faster.

[1]: https://github.com/tarantool/tarantool-python/pull/189

Follows up PR #266


ligurio (approved):
Thanks for the patch!

LGTM

Totktonada:
Updated the test-run submodule in tarantool in the following commits:

* `master`: [2.8.0-106-g6a640ee9e](https://github.com/tarantool/tarantool/commit/6a640ee9ea267244cab02592442bb7b66eb9ba6a)
* `2.7`: 2.7.1-96-gfc0d4c874
* `2.6`: 2.6.2-93-ge001cac12
* `1.10`: 1.10.9-51-gc6fe07572

Issue #226, opened 2020-10-17
test_run:wait_lsn() fails with "attempt to compare nil with number"


Gerold103:
```
[006] Test failed! Result content mismatch:
[006] --- replication/election_qsync.result	Fri Oct 16 00:12:02 2020
[006] +++ replication/election_qsync.reject	Sat Oct 17 19:08:11 2020
[006] @@ -89,6 +89,8 @@
[006]  -- Wait replication to the other instance.
[006]  test_run:wait_lsn('default', 'replica')
[006]   | ---
[006] + | - error: '...sitories/tarantool/test/var/006_replication/test_run.lua:68: attempt
[006] + |     to compare nil with number'
[006]   | ...
[006]  
[006]  test_run:switch('default')
```

avtikhon:
PR #269

PR #269, opened 2021-03-01
Add check in wait_lsn for nil value

avtikhon:
Found issue in testing:
```    
      [006] --- replication/election_qsync.result   Fri Oct 16 00:12:02 2020
      [006] +++ replication/election_qsync.reject   Sat Oct 17 19:08:11 2020
      [006] @@ -89,6 +89,8 @@
      [006]  -- Wait replication to the other instance.
      [006]  test_run:wait_lsn('default', 'replica')
      [006]   | ---
      [006] + | - error: '...sitories/tarantool/test/var/006_replication/test_run.lua:68: attempt
      [006] + |     to compare nil with number'
      [006]   | ...
```    
Found that wait_lsn() routine from test_run.lua script failed on
comparing returned result from get_lsn() routine with real integer.
It happened because get_lsn() returned 'nil' on the freshly
bootstrapped instance. It happened when replica vclock [1:x, 2:1],
while master vclock [1:x, 2:0], but in box.info.vclock[2] is nil
instead of 0. To avoid of it in wait_lsn() routine added check that
get_lsn() routine returned not 'nil' value. Complete explanation of
the issue and reproducer provided in [1].

This test showed the issue in test-run which could happen with any
test that uses test_run:wait_lsn() routine. This patch fixes this
routine to be able to work correctly with 'nil' value returned by
get_lsn() routine.
    
Closes tarantool/test-run#226
    
Co-authored-by: Alexander Turenko <alexander.turenko@tarantool.org>
    
[1]: https://github.com/tarantool/test-run/pull/269#issuecomment-789373803

### DIFF
--- a/test_run.lua
+++ b/test_run.lua
@@ -64,9 +64,10 @@ end
 local function wait_lsn(self, waiter, master)
     local sid = self:get_server_id(master)
     local lsn = self:get_lsn(master, sid)
-
-    while self:get_lsn(waiter, sid) < lsn do
+    local remote_lsn = self:get_lsn(waiter, sid)
+    while remote_lsn == nil or remote_lsn < lsn do
         fiber.sleep(0.001)
+        remote_lsn = self:get_lsn(waiter, sid)
     end
 end
 


LeonidVas:
Hi! Thank you for the patch.
AFAIU in the "election_qsync.test.lua" `wait_lsn()` is used for the old master, and the "bootstrap" problem does not seem relevant in this case.
Also in this test, the replica was started with options `wait=True, wait_load=True`, and if I understand correctly, this means that `replicaset_connect()` was done and `vclock != nil`.

Totktonada:
I wondered how `box.info.vclock` may give a result (not raise an error), where the first item is `nil`. It seems, it is possible on early box configuration stage (for an instance with existing snap/xlog with some operations). It can be verified from a console:

```lua
tarantool> require('fiber').new(function() if box.info.vclock[1] == nil then print(require('json').encode(box.info)) else print('success') end end) box.cfg{}
```

The `box.info.vclock` value is `{}`.

Also I found that when an instance is freshly bootstrapped, it also has `box.info.vclock` value `{}` (easy to check).

----

The code does not fix this particular problem. You check whether `box.info.vclock` is `nil` (not `{}`). If it would `nil`, you would got the 'attempt to index nil value' error on the line 44, not 'attempt to compare nil with number' on line 68.

It seems, you (again, and again, and again...) trying to blindly fix a problem without any reproducer. Ask for help if you need, but don't produce meaningless patches.

Reproducer:

```diff
diff --git a/test/app-tap/test-run-pr-269.test.lua b/test/app-tap/test-run-pr-269.test.lua
new file mode 100755
index 000000000..2ebb108b0
--- /dev/null
+++ b/test/app-tap/test-run-pr-269.test.lua
@@ -0,0 +1,17 @@
+#!/usr/bin/env tarantool
+
+local json = require('json')
+local test_run = require('test_run').new()
+
+test_run:cmd("create server foo with script='box/box.lua'")
+test_run:cmd('start server foo')
+--test_run:eval('foo', "box.once('first dml', function() end)")
+
+local res = test_run:get_param('foo', 'vclock')
+print(json.encode(res))
+local res = test_run:get_lsn('foo', 1)
+print(json.encode(res))
+
+test_run:cmd('stop server foo')
+test_run:cmd('cleanup server foo')
+test_run:cmd('delete server foo')
```

How to run it:

```sh
$ ./test/test-run.py --verbose app-tap/test-run-pr-269.test.lua
<...>
[001] [{}]
[001] null
```

I'll add inline comments regarding the code.

test_run.lua
```diff
@@ -40,8 +40,12 @@ local function get_param(self, node, param)
 end
 
 local function get_lsn(self, node, sid)
-    local nodes = self:get_param(node, 'vclock')
-    return tonumber(nodes[1][tonumber(sid)])
+    local node = self:get_param(node, 'vclock')[1]
```
Totktonada, line 43: The terminology was broken before your patch, but, please, fix it. Otherwise the code will be confusing as for you as well as for others.

It is not a 'node' (a vclock component). It is the whole vclock. `test_run:cmd([[eval foo "return 42"]])` gives a table as result (`{42}`), because in the general case evaluation of a Lua statement may give several values as result (so called multireturn).

test_run.lua
```diff
@@ -40,8 +40,12 @@ local function get_param(self, node, param)
...
+    if node == nil then
```
Totktonada, line 44: Since it is not a vclock component, but a full vclock, the change does not fix the issue. I described it in my [comment](https://github.com/tarantool/test-run/pull/269#issuecomment-789373803).

test_run.lua
```diff
@@ -40,8 +40,12 @@ local function get_param(self, node, param)
...
+        return -1 -- bootstrap in progress
```
Totktonada, line 45: To be honest, giving -1 here does not look natural for me. When we get a vclock component on a freshly bootstrapped instance (`box.info.vclock[1]`) in a local console, we'll receive `nil`. I would keep the behaviour consistent with tarantool's Lua API.

What does it mean in context of the patch? We should revert the changes in `get_lsn()` and handle `nil` result gracefully in `wait_lsn()`.

avtikhon:
> I wondered how `box.info.vclock` may give a result (not raise an error), where the first item is `nil`. It seems, it is possible on early box configuration stage (for an instance with existing snap/xlog with some operations). It can be verified from a console:
> 
> ```lua
> tarantool> require('fiber').new(function() if box.info.vclock[1] == nil then print(require('json').encode(box.info)) else print('success') end end) box.cfg{}
> ```
> 
> The `box.info.vclock` value is `{}`.
> 
> Also I found that when an instance is freshly bootstrapped, it also has `box.info.vclock` value `{}` (easy to check).
> 
> The code does not fix this particular problem. You check whether `box.info.vclock` is `nil` (not `{}`). If it would `nil`, you would got the 'attempt to index nil value' error on the line 44, not 'attempt to compare nil with number' on line 68.
> 
> It seems, you (again, and again, and again...) trying to blindly fix a problem without any reproducer. Ask for help if you need, but don't produce meaningless patches.
> 
> Reproducer:
> 
> ```diff
> diff --git a/test/app-tap/test-run-pr-269.test.lua b/test/app-tap/test-run-pr-269.test.lua
> new file mode 100755
> index 000000000..2ebb108b0
> --- /dev/null
> +++ b/test/app-tap/test-run-pr-269.test.lua
> @@ -0,0 +1,17 @@
> +#!/usr/bin/env tarantool
> +
> +local json = require('json')
> +local test_run = require('test_run').new()
> +
> +test_run:cmd("create server foo with script='box/box.lua'")
> +test_run:cmd('start server foo')
> +--test_run:eval('foo', "box.once('first dml', function() end)")
> +
> +local res = test_run:get_param('foo', 'vclock')
> +print(json.encode(res))
> +local res = test_run:get_lsn('foo', 1)
> +print(json.encode(res))
> +
> +test_run:cmd('stop server foo')
> +test_run:cmd('cleanup server foo')
> +test_run:cmd('delete server foo')
> ```
> 
> How to run it:
> 
> ```shell
> $ ./test/test-run.py --verbose app-tap/test-run-pr-269.test.lua
> <...>
> [001] [{}]
> [001] null
> ```
> 
> I'll add inline comments regarding the code.

Thanks a lot for the explanation. I've added your comment link to commit message and used you reproducer to check the fix.

avtikhon:
> Hi! Thank you for the patch.
> AFAIU in the "election_qsync.test.lua" `wait_lsn()` is used for the old master, and the "bootstrap" problem does not seem relevant in this case.
> Also in this test, the replica was started with options `wait=True, wait_load=True`, and if I understand correctly, this means that `replicaset_connect()` was done and `vclock` must not be empty.
> In other words, I don’t understand why the `vclock` became empty on the old master.

I've changed the patch as @Totktonada suggested.

test_run.lua
avtikhon: Ok, I see, rolled-back.

test_run.lua
avtikhon: Used your reproducer to check that tonumber routine removed 'nil' on given '{}'. Added check in wait_lsn routine that get_lsn routine returned not 'nil' value. Checked the fix with test that uses wait_lsn routine:
```
./test-run.py replication/gh-4730-applier-rollback.test.lua
```
with reproducer of the nil value to be sure that the issue fixed:
```
diff --git a/test_run.lua b/test_run.lua
index 7425b4a..eff78dc 100644
--- a/test_run.lua
+++ b/test_run.lua
@@ -68,6 +68,7 @@ local function wait_lsn(self, waiter, master)
     repeat
         fiber.sleep(0.001)
         local waiter_lsn = self:get_lsn(waiter, sid)
+        waiter_lsn = nil
     until waiter_lsn ~= nil and waiter_lsn >= lsn
 end
 
```

test_run.lua
avtikhon: Thanks for the explanation, I've moved the changes from get_lsn to wait_lsn as you suggested.

LeonidVas:
> > Hi! Thank you for the patch.
> > AFAIU in the "election_qsync.test.lua" `wait_lsn()` is used for the old master, and the "bootstrap" problem does not seem relevant in this case.
> > Also in this test, the replica was started with options `wait=True, wait_load=True`, and if I understand correctly, this means that `replicaset_connect()` was done and `vclock` must not be empty.
> > In other words, I don’t understand why the `vclock` became empty on the old master.
> 
> I've changed the patch as @Totktonada suggested.

Hi!
My question: "Why is the `vclock` empty on the old master?"
And Alexander's comments don't answer on this question.

About the patch:
I had a conversation with @Totktonada and I don't object against updating `wait_lsn()` function, but I think that this change doesn't fix the problem with the `election_qsync.test.lua` test.

test_run.lua
```diff
@@ -65,9 +65,10 @@ local function wait_lsn(self, waiter, master)
     local sid = self:get_server_id(master)
     local lsn = self:get_lsn(master, sid)
 
-    while self:get_lsn(waiter, sid) < lsn do
+    repeat
         fiber.sleep(0.001)
-    end
+        local waiter_lsn = self:get_lsn(waiter, sid)
+    until waiter_lsn ~= nil and waiter_lsn >= lsn
```
LeonidVas, line 71: Maybe `waiter_lsn < lsn`?
AFAIU @Totktonada suggests to continue waiting in case of an empty `vclock`.

LeonidVas (changes requested):
If you agree with https://github.com/tarantool/test-run/pull/269#issuecomment-789677097 - update the commit message.
Please, don't ignore:"The terminology was broken before your patch, but, please, fix it." from https://github.com/tarantool/test-run/pull/269#discussion_r586057480

test_run.lua
avtikhon: Seems you missed that I've used until in place of while.

avtikhon:
> If you agree with [#269 (comment)](https://github.com/tarantool/test-run/pull/269#issuecomment-789677097) - update the commit message.

Added comment in commit message and PR.

> Please, don't ignore:"The terminology was broken before your patch, but, please, fix it." from [#269 (comment)](https://github.com/tarantool/test-run/pull/269#discussion_r586057480)

Seems this fix need to be done in separate commit.

LeonidVas:
> > Please, don't ignore:"The terminology was broken before your patch, but, please, fix it." from [#269 (comment)](https://github.com/tarantool/test-run/pull/269#discussion_r586057480)
> 
> Seems this fix need to be done in separate commit.
Yep. The PR can contain 2 commits.
It's not a fix. It something like "stylistic correction".
Up to you.

LeonidVas:
> > If you agree with [#269 (comment)](https://github.com/tarantool/test-run/pull/269#issuecomment-789677097) - update the commit message.
> 
> Added comment in commit message and PR.

Hi!
About the commit message:
- At the moment this change doesn't seem to fix anything, it just changes the behavior, and I don't believe the new one will be better, because before you were getting an error in the place where something went wrong, but now you will have a hang test, which will be killed after 120 seconds. As for me, I think the old behavior was clearer. But this change could make it easier to write tests (potentially), so I don't mind.
- If the first is true. This change is not related with issue https://github.com/tarantool/test-run/issues/226 at all, and the commit message should only contain a description of the behavior and motivation change.

test_run.lua
LeonidVas: D'oh! I confused the behavior of `do {...} while (condition)` with the `repeat ... until condition`.

Totktonada, line 71: (Regarding the code itself.) This way we sleep first, then make attempt to get the LSN. It is better to check LSN first and sleep if the attempt is unsuccessful:

```lua
local remote_lsn = self:get_lsn(waiter, sid)
while remote_lsn == nil or remote_lsn < lsn do
    fiber.sleep(0.001)
    remote_lsn = self:get_lsn(waiter, sid)
end
```

avtikhon: Ok, let's use your variant.

avtikhon:
> > > If you agree with [#269 (comment)](https://github.com/tarantool/test-run/pull/269#issuecomment-789677097) - update the commit message.
> > 
> > 
> > Added comment in commit message and PR.
> 
> Hi!
> About the commit message:
> 
> * At the moment this change doesn't seem to fix anything, it just changes the behavior, and I don't believe the new one will be better, because before you were getting an error in the place where something went wrong, but now you will have a hang test, which will be killed after 120 seconds. As for me, I think the old behavior was clearer. But this change could make it easier to write tests (potentially), so I don't mind.
> * If the first is true. This change is not related with issue #226 at all, and the commit message should only contain a description of the behavior and motivation change.

This patch does not fix the found issue in tests, but changes the
way it works with the issue. Instead of breaking the test run, for
now it will wait for the lsn value different from nil in loop till
available timeout.

test_run.lua
LeonidVas: > (Regarding the code itself.) This way we sleep first, then make attempt to get the LSN. It is better to check LSN first and sleep if the attempt is unsuccessful:
> 
> ```lua
> local remote_lsn = self:get_lsn(waiter, sid)
> while remote_lsn == nil or remote_lsn < lsn do
>     fiber.sleep(0.001)
>     remote_lsn = self:get_lsn(waiter, sid)
> end
> ```

And now you get lsn, sleep and after do the check. Why do you think it's better? As for me I am comfortable with both.

LeonidVas:
> > > > If you agree with [#269 (comment)](https://github.com/tarantool/test-run/pull/269#issuecomment-789677097) - update the commit message.
> > > 
> > > 
> > > Added comment in commit message and PR.
> > 
> > 
> > Hi!
> > About the commit message:
> > 
> > * At the moment this change doesn't seem to fix anything, it just changes the behavior, and I don't believe the new one will be better, because before you were getting an error in the place where something went wrong, but now you will have a hang test, which will be killed after 120 seconds. As for me, I think the old behavior was clearer. But this change could make it easier to write tests (potentially), so I don't mind.
> > * If the first is true. This change is not related with issue #226 at all, and the commit message should only contain a description of the behavior and motivation change.
> 
> This patch does not fix the found issue in tests, but changes the
> way it works with the issue. Instead of breaking the test run, for
> now it will wait for the lsn value different from nil in loop till
> available timeout.

Yep, I wrote the same in the first paragraph.

test_run.lua
LeonidVas: In the first variant, you are more likely to get a larger lsn.

Totktonada: > Why do you think it's better?

No sleep when suitable data is already available.

LeonidVas: > > Why do you think it's better?
> 
> No sleep when suitable data is already available.

Ok. Sorry, I think you are suggesting to simply swap `sleep` with `get_lsn`.

Totktonada:
> This patch does not fix the found issue in tests, but changes the
> way it works with the issue.

Since the change is described in context of the `replication/election_qsync.test.lua` problem and it does not fix the problem, I don't see any reason to push it.

avtikhon:
> > This patch does not fix the found issue in tests, but changes the
> > way it works with the issue.
> 
> Since the change is described in context of the `replication/election_qsync.test.lua` problem and it does not fix the problem, I don't see any reason to push it.

Once more discussed this issue with @sergepetrenko and found that this situation could happen
when replica vclock [1:x, 2:1], while master vclock [1:x, 2:0], but in box.info.vclock[2] is nil instead of 0.

Discussed once more with @Totktonada this fix and decided to correct commit message with more clear
information what this fix is really for.

Totktonada:
I propose to consider #226 as the problem with inability to wait until a vclock component will going above zero and close the issue with this PR. If there are other problems in the test, let's investigate them out of scope of the issue.

Totktonada:
Since @sergepetrenko approved that the described situation is possible in the test, I'm okay with wording that describes zero vclock component problem and the particular test that was hit by the problem.

Totktonada (approved):
LGTM (aside that I vote for closing the issue from this PR).

avtikhon:
> LGTM (aside that I vote for closing the issue from this PR).

Agree to close the issue, corrected commit message.

LeonidVas (approved):
LGTM.
I wrote optional fixes for the commit message in a private conversation.

Totktonada:
Updated the test-run submodule in tarantool in [2.8.0-134-g81c663335](https://github.com/tarantool/tarantool/commit/81c6633358d0739d9459423cb1ed6441a2b6d2e2), 2.7.1-123-ge3d1d9e7a, 2.6.2-120-g790b55a92, 1.10.9-78-g1a4e6d963.

Issue #156, opened 2019-04-08
pretest_clean: revoke non-default privileges of default users


Totktonada:
And grant revoked ones.

Totktonada:
Not actual anymore, see PR #309.

PR #273, opened 2021-03-12
test: add initial test suite

ligurio:
test-run supports three types of tests:

- tarantool - Test-Suite for Functional Testing
- app - Another functional Test-Suite
- unittest - Unit-Testing Test Suite

Patch adds tests for two of supported test types:

- test-app
- test-tarantool

How-to run:

$ mkdir build && cd build
$ make test_integration

### DIFF
--- a/.luacheckrc
+++ b/.luacheckrc
@@ -26,4 +26,5 @@ include_files = {
 
 exclude_files = {
     "lib/tarantool-python",
+    "test/test-tarantool/*.test.lua",
 }
--- a/Makefile
+++ b/Makefile
@@ -1,11 +1,20 @@
+TEST_RUN_EXTRA_PARAMS?=
+PYTHON?=python
+
 default:
 	false
 
-.PHONY: lint flake8 luacheck
 lint: flake8 luacheck
 
 flake8:
-	python -m flake8 *.py lib/*.py
+	$(PYTHON) -m flake8 *.py lib/*.py
 
 luacheck:
 	luacheck --config .luacheckrc .
+
+test_integration:
+	$(PYTHON) test/test-run.py --force $(TEST_RUN_EXTRA_PARAMS)
+
+test: test_integration
+
+.PHONY: lint flake8 luacheck test test_integration
--- a/test/test-app/cfg.test.lua
+++ b/test/test-app/cfg.test.lua
@@ -1,0 +1,19 @@
+#!/usr/bin/env tarantool
+
+-- Test is an example of TAP test
+
+local tap = require('tap')
+local test = tap.test('cfg')
+test:plan(4)
+
+box.cfg{listen = box.NULL}
+test:is(nil, box.info.listen, 'no cfg.listen - no info.listen')
+
+box.cfg{listen = '127.0.0.1:0'}
+test:ok(box.info.listen:match('127.0.0.1'), 'real IP in info.listen')
+test:ok(not box.info.listen:match(':0'), 'real port in info.listen')
+
+box.cfg{listen = box.NULL}
+test:is(nil, box.info.listen, 'cfg.listen reset drops info.listen')
+
+os.exit(test:check() and 0 or 1)
--- a/test/test-tarantool/engine.cfg
+++ b/test/test-tarantool/engine.cfg
@@ -1,0 +1,15 @@
+{
+    "set_language.test.sql": {
+        "memtx": {"engine": "memtx"}
+    },
+    "setopt_delimeter.test.lua": {
+        "memtx": {"engine": "memtx"}
+    },
+    "worker_hang_when_gc_triggered_inside_colorer.test.lua": {
+        "vinyl": {"engine": "vinyl"}
+    },
+    "*": {
+        "memtx": {"engine": "memtx"},
+        "vinyl": {"engine": "vinyl"}
+    }
+}
--- a/test/test-tarantool/iproto.test.py
+++ b/test/test-tarantool/iproto.test.py
@@ -1,0 +1,55 @@
+"""
+Original Tarantool's test box-py/iproto.test.py had a problem when output with
+running under Python 2 was not the same as with running under Python 3.
+
+Fixed in commit 697b79781cc63e2d87d86d43713998261d602334
+"test: make output of box-py/iproto.test.py deterministic".
+"""
+
+from __future__ import print_function
+
+import msgpack
+from tarantool.const import *
+from tarantool import Connection
+from tarantool.response import Response
+from lib.tarantool_connection import TarantoolConnection
+
+# Note re IPROTO_SQL_INFO_* keys: they cannot appear in the
+# response map at the top level, but have the same codes as other
+# IPROTO_* constants. Exclude those names so.
+key_names = {}
+for (k,v) in list(globals().items()):
+    if type(k) == str and k.startswith("IPROTO_") and \
+            not k.startswith("IPROTO_SQL_INFO_") and type(v) == int:
+        key_names[v] = k
+
+def repr_dict(todump):
+    d = {}
+    for (k, v) in todump.items():
+        k_name = key_names.get(k, k)
+        d[k_name] = v
+    return repr(sorted(d.items()))
+
+
+def test(header, body):
+    # Connect and authenticate
+    c = Connection("localhost", server.iproto.port)
+    c.connect()
+    print("query", repr_dict(header), repr_dict(body))
+    header = msgpack.dumps(header)
+    body = msgpack.dumps(body)
+    query = msgpack.dumps(len(header) + len(body)) + header + body
+    # Send raw request using connected socket
+    s = c._socket
+    try:
+        s.send(query)
+    except OSError as e:
+        print("   => ", "Failed to send request")
+    c.close()
+    print(iproto.py_con.ping() > 0)
+
+print("IPROTO_UPDATE")
+test({ IPROTO_CODE : REQUEST_TYPE_UPDATE }, { IPROTO_SPACE_ID: 280 })
+test({ IPROTO_CODE : REQUEST_TYPE_UPDATE },
+     { IPROTO_SPACE_ID: 280, IPROTO_KEY: (1, )})
+print("\n")
--- a/test/test-tarantool/setopt_delimeter.test.lua
+++ b/test/test-tarantool/setopt_delimeter.test.lua
@@ -1,0 +1,22 @@
+-- Simple test that uses 'setopt delimiter' command.
+-- Command introduced in commit 6e38b88eb6bbe543a1e3ba0a6a0be2f6f58abc86
+-- ('Implement SQL driver')
+
+test_run = require('test_run').new()
+
+-- Using delimiter
+_ = test_run:cmd("setopt delimiter ';'")
+function test_a()
+    local a = 1
+end;
+_ = test_run:cmd("setopt delimiter ''");
+
+box.cfg{}
+
+-- Using multiline
+box.cfg{                     \
+    coredump = false,        \
+    log_format = 'plain',    \
+    log_level = 5,           \
+    strip_core = true        \
+}
--- a/test/test-unit/broken_unicode.test
+++ b/test/test-unit/broken_unicode.test
@@ -1,0 +1,15 @@
+#!/bin/sh
+
+# Test is needed to check how test-run handle
+# tests with result files. Since commit
+# 395edeb6b743c4479a62dd2183062124973d2b2a
+# 'python3: decouple bytes and strings' test-run
+# reads test output and result files in byte mode.
+# Output of this test contains broken UTF-8 sequences
+# and test-run will fail if it will try to decode output.
+
+printf 'TAP version 13\n'
+printf '1..3\n'
+printf 'ok 1 - \302\302\n'
+printf 'ok 2 - \302\302\n'
+printf 'ok 3 - \342\230\240\n'
--- a/test/test-unit/suite.ini
+++ b/test/test-unit/suite.ini
@@ -1,0 +1,4 @@
+[default]
+core = unittest
+description = unit tests
+is_parallel = True
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -14,10 +14,17 @@ jobs:
       fail-fast: false
       matrix:
         python-version: [2.7, 3.5, 3.6, 3.7, 3.8, 3.8-dev]
+        tarantool-version: [2.6, 2.7]
 
     steps:
     - uses: actions/checkout@v2
-    - name: Set up Python ${{ matrix.python-version }}
+    - name: update git submodules
+      run: git submodule update --init --recursive
+    - name: set up Tarantool ${{ matrix.tarantool-version }}
+      uses: tarantool/setup-tarantool@v1
+      with:
+        tarantool-version: ${{ matrix.tarantool-version }}
+    - name: set up Python ${{ matrix.python-version }}
       uses: actions/setup-python@v2
       with:
         python-version: ${{ matrix.python-version }}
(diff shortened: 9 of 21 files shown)


ligurio:
ready to review

ligurio:
```
  File "/home/runner/work/test-run/test-run/lib/tarantool_server.py", line 2, in <module>
    import gevent
ModuleNotFoundError: No module named 'gevent'
make: *** [Makefile:19: test_integration] Error 1
```

.luacheckrc
```diff
@@ -26,4 +26,6 @@ include_files = {
 
 exclude_files = {
     "lib/tarantool-python",
+    "test/test-tarantool/panic_on_broken_lsn.test.lua",
+    "test/test-tarantool/setopt_delimeter.test.lua",
```
Totktonada, line 30: I think we should exclude the whole test suite, because it is 'core = tarantool' (so, it is not a lua interpreter input, but tarantool console input).

test/.tarantoolctl
```diff
@@ -0,0 +1,14 @@
...
+    wal_dir    = workdir,
+    snap_dir   = workdir,
+    vinyl_dir  = workdir,
+    logger     = workdir,
+    background = false,
+}
+
+instance_dir = workdir
+
+-- vim: set ft=lua :
```
Totktonada, line 14: I would prefer to use the tarantoolctl config from test-run. See https://github.com/tarantool/tarantool/commit/6c8efa8491ed3fc2834938ec65d91dc1f7ec7e34 and https://github.com/tarantool/tarantool/commit/54667f6fa86ff1ab0030501abc9a59d0d7aea82f. Is there a reason to place our own config here?

(I would also highlight that `logger` and `snap_dir` options are [deprecated](https://www.tarantool.io/en/doc/latest/reference/configuration/#deprecated-parameters).)

test/test-app/cfg.test.lua
```diff
@@ -0,0 +1,20 @@
...
+
+box.cfg{listen = '127.0.0.1:0'}
+test:ok(box.info.listen:match('127.0.0.1'), 'real IP in info.listen')
+test:ok(not box.info.listen:match(':0'), 'real port in info.listen')
+
+box.cfg{listen = box.NULL}
+test:is(nil, box.info.listen, 'cfg.listen reset drops info.listen')
+
+test:check()
+os.exit(0)
```
Totktonada, line 20: The [rule of thumb](https://github.com/tarantool/doc/issues/1004) is to set a non-zero exit code, when the check for the plan fails.

test/test-tarantool/iproto.test.py
```diff
@@ -0,0 +1,55 @@
+"""
+Original Tarantool's test box-py/iproto.test.py had a problem when output with
+running under Python 2 was not the same as with running under Python 3.
+
+Fixed in commit 697b79781cc63e2d87d86d43713998261d602334
+"test: make output of box-py/iproto.test.py deterministic".
```
Totktonada, line 6: Is this test requires a special handling on the test-run side? Or here we just verify the `Connection` class from the tarantool-python connector provided by test-run?

Similarly to the [comment](https://github.com/tarantool/test-run/pull/283#discussion_r603731146) regarding `call.test.py`: you link a problem around Python 3 compatibility *of the test itself*. It is interesting, but it does not reveal motivation of your choice to add it here.

test/test-tarantool/setopt_delimeter.test.lua
```diff
@@ -0,0 +1,22 @@
...
+    local a = 1
+end;
+_ = test_run:cmd("setopt delimiter ''");
+
+box.cfg{}
+
+-- Using multiline
+box.cfg{                     \
+    coredump = false,        \
+    log_format = plain,      \
```
Totktonada, line 19: Did you mean `'plain'` (in quotes)?

test/test-tarantool/panic_on_broken_lsn.test.lua
```diff
@@ -0,0 +1,9 @@
+-- regression test for https://github.com/tarantool/tarantool-qa/issues/96
```
Totktonada, line 1: 1. I would link a particular problem. The QA issue is too general: it links two problems now and we maybe will add more ones on this kind to it (if we'll see them in CI). There is no relevant issue in test-run, because I just filed PR with the problem description: https://github.com/tarantool/test-run/pull/275 (not a best practice, sure).
2. I would rename the test case to something relevant to the problem (there is nothing about a broken snapshot here).
3. And the test case reproduces the problem with a small probability (without patching `multiprocessing/connection.py`), so I don't know, whether it ever worth to add it. At least we should make this fact clear.

test/test-tarantool/panic_on_broken_lsn.test.lua
```diff
@@ -0,0 +1,9 @@
...
+
+ok, err = pcall(box.schema.user.grant, 'guest', 'replication')
```
Totktonada, line 3: I would write a test like it is run in a clean tarantool instance (without `pcall()` in this case). Tests we add here are a kind of examples for others and it is better to fix the [problem](https://github.com/tarantool/test-run/issues/156) rather than suggest a workaround.

The rule of thumb is to revoke privileges we granted in the test. If we'll add `box.schema.user.revoke('guest', 'replication')` at the end, the test may be run several times on one instance (and we'll not need `pcall()`).

Makefile
```diff
@@ -1,11 +1,23 @@
...
 luacheck:
 	luacheck --config .luacheckrc .
+
+build_unit_tests:
+	make -C $(UNIT_TEST_DIR)
+
+test_integration: build_unit_tests
+	test/test-run.py --builddir=build/ --var=build/test/var --force $(TEST_RUN_EXTRA_PARAMS)
+
+test: test_integration
```
Totktonada, line 22: Regarding the PR description: maybe it would be better to suggest `make test` rather than direct ./test-run.py invocation with parameters?

test/test-unit/Makefile
```diff
@@ -0,0 +1,16 @@
+BIN_DIR=../../build/test/test-unit
+TEST_BINARIES=read_bytes.test
+
+all: $(TEST_BINARIES)
+
+read_bytes.test: read_bytes.c unit.c
+	gcc -o $@ $^
```
Totktonada, line 7: I guess `$(CC)` would be useful for ones, who have clang installed (but not gcc).

test/test-tarantool/proxy.lua
```diff
@@ -0,0 +1,11 @@
+#!/usr/bin/env tarantool
```
Totktonada, line 1: According to the [code](https://github.com/tarantool/test-run/blob/d986990b9a8a8f48f05ba03690877c4e5823aa28/lib/preprocessor.py#L329), it will not be used.

test/test-unit/suite.ini
```diff
@@ -0,0 +1,5 @@
+[default]
+core = unittest
+description = unit tests
+is_parallel = True
+disabled = read_bytes.test
```
Totktonada, line 5: Is it disabled by intention?

test/test-tarantool/read_bytes.result
```diff
@@ -0,0 +1,16 @@
...
+ok 5 - error on position 1 for <''>
+ok 6 - error on position 1 for < field1>
+ok 7 - error on position 1 for <1field>
+ok 8 - error on position 2 for <.1field>
+ok 9 - error on position 8 for <['field>
+ok 10 - error on position 9 for <['field'>
+ok 11 - error on position 5 for <[123>
+ok 12 - error on position 3 for <['']>
+ok 13 - error on position 2 for <.[123]>
+ok 22 - invalid token for index_base 1
```
Totktonada, line 16: There is no such test in the `test-tarantool` test suite.

test/test-tarantool/engine.cfg
```diff
@@ -0,0 +1,16 @@
...
+    "set_language.test.lua": {
+        "memtx": {"engine": "memtx"}
+    },
+    "setopt_delimeter.test.lua": {
+        "memtx": {"engine": "memtx"}
+    },
+    "panic_on_broken_lsn.test.lua": {
+        "vinyl": {"engine": "vinyl"}
+    },
+    "manage_cluster.test.lua": {},
```
Totktonada, line 11: Hm. There is no such test. Is it intentional?

test/test-unit/Makefile
```diff
@@ -0,0 +1,16 @@
+BIN_DIR=../../build/test/test-unit
```
Totktonada, line 1: Should it be added to .gitignore?

test/test-unit/Makefile
Totktonada, line 7: Just side note: we can add any executable as a unit test, including one that is written on a scripting language (Bourne shell, Python). Only exit code, stdout and stderr are important for test-run.

So maybe it would be easier to verify different things around unit tests using scripts.

test/.tarantoolctl
ligurio: There are no reason for this.
Ok, removed custom config to use `.tarantoolctl` in a root dir.

test/test-app/cfg.test.lua
ligurio: Updated.

test/test-tarantool/setopt_delimeter.test.lua
ligurio: Yes, fixed.

test/test-unit/Makefile
ligurio: Fixed.

test/test-tarantool/engine.cfg
ligurio: No, removed.

test/test-unit/Makefile
ligurio: `/build` is already there

test/test-unit/suite.ini
ligurio: Forgot to enable it back, thanks.

test/test-tarantool/proxy.lua
ligurio: removed.

Makefile
ligurio: Updated description

BTW commit message describes `make test`:
```
commit e74f813d41c9f1cad9e4e3a84bc5b34253253e96
Author: Sergey Bronnikov <sergeyb@tarantool.org>
Date:   Fri Mar 12 14:15:44 2021 +0000

    test: add integration tests
    
    test-run supports three types of tests:
    
    - tarantool - Test-Suite for Functional Testing
    - app - Another functional Test-Suite
    - unittest - Unit-Testing Test Suite
    
    Patch adds tests for two of supported test types:
    
    - test-app for type 'app'
    - test-tarantool for type 'tarantool'
    
    How-to run:
    
    $ make test_integration
    
    - test-tarantool/panic_on_broken_lsn.test.lua [1]
    
    1. https://github.com/tarantool/tarantool-qa/issues/96
```

.luacheckrc
ligurio: Excluded by mask.

test/test-app/cfg.test.lua
Totktonada: `os.exit(<boolean>)` always gives zero exit code. I suggest to use `os.exit(test:check() and 0 or 1)`.

test/test-unit/Makefile
Totktonada: (`printf` — not `echo` — is for bash / dash portability.)

Totktonada: (Sergey pointed me that printf in dash [does not support](https://man7.org/linux/man-pages/man1/dash.1.html) hexadecimal byte values. We can use octal ones instead: `printf '\300\000\n'`.)

test/test-tarantool/proxy.lua
Totktonada: Hm. It is here now.

test/test-unit/broken_unicode.test
```diff
@@ -0,0 +1,15 @@
...
+# Test is needed to check how test-run handle
+# tests with result files. Since commit
+# 395edeb6b743c4479a62dd2183062124973d2b2a
+# 'python3: decouple bytes and strings' test-run
+# reads test output and result files in byte mode.
+# Output of this test contains broken unicode sequences
+# and test-run will fail if it will try to decode output.
+
+printf 'TAP version 13\n'
+printf '1..2\n'
```
Totktonada, line 12: Should be `1..3`?

test/test-tarantool/iproto.test.py
ligurio: You are right, nothing special for test-run itself in a test. To support Python 3 in call.test.py there were commits to tarantool-python and test itself. But I would keep test as integration check to make sure call.test.py is not broken. Althugh probably it is better to add a unit test to tarantool-python to cover this case.

test/test-unit/broken_unicode.test
```diff
@@ -0,0 +1,15 @@
+#!/bin/sh
+
+# Test is needed to check how test-run handle
+# tests with result files. Since commit
+# 395edeb6b743c4479a62dd2183062124973d2b2a
+# 'python3: decouple bytes and strings' test-run
+# reads test output and result files in byte mode.
+# Output of this test contains broken unicode sequences
```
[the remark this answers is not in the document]
ligurio: Fixed

test/test-unit/broken_unicode.test
ligurio: Fixed

test/test-unit/Makefile
```diff
@@ -0,0 +1,10 @@
+BIN_DIR := ../../test/build/test/test-unit/
+TESTS := broken_unicode.test
+
+all: install
+
+install: $(TESTS)
+	[ -e $(BIN_DIR) ] || mkdir -p $(BIN_DIR)
+	cp $^ $(BIN_DIR)
+
+.PHONY: all clean $(TESTS)
```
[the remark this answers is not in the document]
ligurio: cleaned up things related to build C test in a root Makefile and removed `test/test-unit/Makefile` at all

test/test-tarantool/panic_on_broken_lsn.test.lua
ligurio: removed pcall() wrapper and added command with rights revoke

test/test-unit/broken_unicode.test
Totktonada: It seems, the result file was not updated.

test/test-unit/Makefile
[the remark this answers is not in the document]
Totktonada: `--builddir` should be removed too (or we should return copying of the the test back).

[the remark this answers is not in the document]
ligurio: `--builddir` removed as well as `--vardir `

test/test-app/cfg.test.lua
ligurio:
```
--- a/test/test-app/cfg.test.lua
+++ b/test/test-app/cfg.test.lua
@@ -16,4 +16,4 @@ test:ok(not box.info.listen:match(':0'), 'real port in info.listen')
 box.cfg{listen = box.NULL}
 test:is(nil, box.info.listen, 'cfg.listen reset drops info.listen')
 
-os.exit(test:check())
+os.exit(test:check() and 0 or 1)
```

test/test-unit/Makefile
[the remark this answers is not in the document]
Totktonada: Nice!

test/test-tarantool/iproto.test.py
Totktonada: Okay.

test/test-tarantool/engine.cfg
```diff
@@ -0,0 +1,15 @@
...
+        "memtx": {"engine": "memtx"}
+    },
+    "worker_hang_when_gc_triggered_inside_colorer.test.lua": {
+        "vinyl": {"engine": "vinyl"}
+    },
+    "*": {
+        "memtx": {"engine": "memtx"},
+        "vinyl": {"engine": "vinyl"}
+    }
+}
```
Totktonada, line 15: I updated this file according to recent changes (I hope you don't mind):

```diff
diff --git a/test/test-tarantool/engine.cfg b/test/test-tarantool/engine.cfg
index e6c80a1..b8b138b 100644
--- a/test/test-tarantool/engine.cfg
+++ b/test/test-tarantool/engine.cfg
@@ -1,11 +1,11 @@
 {
-    "set_language.test.lua": {
+    "set_language.test.sql": {
         "memtx": {"engine": "memtx"}
     },
     "setopt_delimeter.test.lua": {
         "memtx": {"engine": "memtx"}
     },
-    "panic_on_broken_lsn.test.lua": {
+    "worker_hang_when_gc_triggered_inside_colorer.test.lua": {
         "vinyl": {"engine": "vinyl"}
     },
     "*": {
```

Totktonada (approved):
LGTM.

test/test-tarantool/engine.cfg
Totktonada: (Resolved.)

PR #281, opened 2021-03-24
Implement new option, which allows to set allocator for memtx

EvgenyMekhanik:
The ability to select an allocator for memtx has been added to
tarantool. To test a new type of allocator, all tests must also
be run with it. Implemented new option, which allows to set allocator
for memtx. If you wan't to choose allocator type for tests, run test-run.py
with --memtx-allocator="small" or --memtx-allocator="system". Allocator type
is passed via MEMTX_ALLOCATOR environment variable to the test.

Part of https://github.com/tarantool/tarantool/issues/5419

### DIFF
--- a/lib/options.py
+++ b/lib/options.py
@@ -252,6 +252,11 @@ class Options:
                 action="store_true",
                 default=False,
                 help="""Disable schema upgrade on testing with snapshots.""")
+        parser.add_argument(
+                "--memtx-allocator",
+                dest="memtx_allocator",
+                default=os.environ.get("MEMTX_ALLOCATOR", "small"),
+                help="""Memtx allocator type for tests""")
 
         # XXX: We can use parser.parse_intermixed_args() on
         # Python 3.7 to understand commands like
--- a/lib/__init__.py
+++ b/lib/__init__.py
@@ -58,6 +58,7 @@ def module_init():
     os.environ["LUA_PATH"] = SOURCEDIR+"/?.lua;"+SOURCEDIR+"/?/init.lua;;"
     os.environ["LUA_CPATH"] = BUILDDIR+"/?."+soext+";;"
     os.environ["REPLICATION_SYNC_TIMEOUT"] = str(args.replication_sync_timeout)
+    os.environ['MEMTX_ALLOCATOR'] = args.memtx_allocator
 
     TarantoolServer.find_exe(args.builddir)
     UnittestServer.find_exe(args.builddir)


Mons:
Do not use `TT_` prefix for environment option.
It is only for tarantool itself.

Also typo in `ALLOCATOTR`

lib/options.py
```diff
@@ -252,6 +252,11 @@ def __init__(self):
                 action="store_true",
                 default=False,
                 help="""Disable schema upgrade on testing with snapshots.""")
+        parser.add_argument(
+                "--allocator",
+                dest="allocator",
+                default=env_int('TEST_RUN_MEMTX_ALLOCATOR', "small"),
```
Totktonada, line 258: `env_int()` gives an integer (or set a default value).

lib/options.py
```diff
@@ -15,6 +15,14 @@ def env_int(name, default):
         return default
 
 
+def env_string(name, default):
+    try:
+        value = os.environ.get(name)
+        return default if value is None else value
+    except ValueError:
+        return default
```
Totktonada, line 23: `ValueError` is not possible here. Can be rewritten this way:

```python
return os.environ.get(name, default)
```

However we don't need to introduce the special function for this. It is enough to just pass `default=os.environ.get("TEST_RUN_MEMTX_ALLOCATOR", "memtx")` to `add_arguments()`.

Totktonada:
Part of https://github.com/tarantool/tarantool/issues/5419

lib/options.py
```diff
@@ -252,6 +260,11 @@ def __init__(self):
                 action="store_true",
                 default=False,
                 help="""Disable schema upgrade on testing with snapshots.""")
+        parser.add_argument(
+                "--allocator",
+                dest="allocator",
+                default=env_string('TEST_RUN_MEMTX_ALLOCATOR', "small"),
```
Totktonada, line 266: The name of the test-run parameter (`--allocator`) is not consistent with the name of test-run's environment variable (`TEST_RUN_MEMTX_ALLOCATOR`). The name of the environment variable is not consistent with the name of the box.cfg parameter (`allocator`). I would use the same name (ignoring lower/upper case, dash/underscore, `--` or `TEST_RUN_` prefix) everywhere. Is there decision how the `box.cfg` will be named: `allocator` or `memtx_allocator`?

Totktonada, line 266: Nit: Different quote styles within one statement.

Totktonada, line 266: Regarding prefixing the environment variable name with `TEST_RUN_` or not.

1. There are `TEST_RUN_TESTS` and `TEST_RUN_JOBS` variables, which configures test-run itself. However more recent `TEST_TIMEOUT` and `NO_OUTPUT_TIMEOUT` were added without the prefix. To be honest, I just don't find any gains and decided to get rid of the prefix to make it more convenient in use.
2. There are several other `TEST_RUN_*` variables, which are set by test-run itself for child processes (such as `TEST_RUN_WORKER_ID`).
3. There is `REPLICATION_SYNC_TIMEOUT` environment variable (and `--replication-sync-timeout` parameter) that reflects the box.cfg option name. It have no `TEST_RUN_` prefix.

Our case is quite similar to the third one. So, if there are no complains, I would follow this naming.

test-run.py
```diff
@@ -272,6 +272,9 @@ def open_as_utf8(*args, **kwargs):
 
     status = 0
 
+    # set allocator for memtx, default small
+    os.environ['TEST_RUN_MEMTX_ALLOCATOR'] = Options().args.allocator
```
Totktonada, line 276: The similar `REPLICATION_SYNC_TIMEOUT` is set in `lib/__init.py`, I would do the same for this environment variable.

lib/options.py
EvgenyMekhanik: memtx_allocator

lib/options.py
Totktonada: Fixed.

lib/options.py
Totktonada: Fixed.

lib/options.py
Totktonada: Now the names are consistent. Thanks!

Totktonada: Now it is `MEMTX_ALLOCATOR` without the prefix. I think it is good choice.

test-run.py
Totktonada: Moved.

Totktonada:
Everything looks good on the first glance. I'll test it a bit and proceed if I'll not find any problems.

EvgenyMekhanik:
> Everything looks good on the first glance. I'll test it a bit and proceed if I'll not find any problems.

Thanks, I think that this patch should be pushed after LGTM in corresponding patch 

> Everything looks good on the first glance. I'll test it a bit and proceed if I'll not find any problems.

Thanks, I hink this patch should be pushed after LGTM in corresponding patch https://github.com/tarantool/tarantool/pull/5670

Totktonada:
If there is an agreement about the option name (it seems, we have it) I'm okay to push the test-run side support before tarantool's patchset.

Totktonada:
## Testing

Added a couple of tests, which read the `MEMTX_ALLOCATOR` environment variable in different ways:

* From a 'core = unittest' test.
* From a 'core = app' test.
* From a 'core = tarantool' test.
* From a non-default instance of a 'core = tarantool' test.
* From an instance file of a 'core = tarantool' test suite.

...and verify that the variable's value is "system".

The patch with tests is at end of the comment.

Performed the following runs with current test-run master (cb5adde9d399bdf478082a061739b78647752c85), on current tarantool master (2.8.0-166-g234472522):

```sh
./test/test-run.py test-run-pr-281 test_run_pr_281
<..all tests failed (as expected)..>
```

```sh
MEMTX_ALLOCATOR=system ./test/test-run.py test-run-pr-281 test_run_pr_281
<..all tests passed (as expected)..>
```

Cherry-picked the Evgeny's commit on top of the test-run master and performed the folllowing runs:

```sh
./test/test-run.py test-run-pr-281 test_run_pr_281
<..all tests failed (as expected)..>
```

```sh
MEMTX_ALLOCATOR=system ./test/test-run.py test-run-pr-281 test_run_pr_281
<..all tests passed (as expected)..>
```

```sh
./test/test-run.py --memtx-allocator system test-run-pr-281 test_run_pr_281
<..all tests passed (as expected)..>
```

My conclusion: everything works as expected.

----

<details>
<summary>The patch with tests</summary>

```diff
diff --git a/test/box-tap/test-run-pr-281-memtx-allocator.test.lua b/test/box-tap/test-run-pr-281-memtx-allocator.test.lua
new file mode 100755
index 000000000..78c3cc4b0
--- /dev/null
+++ b/test/box-tap/test-run-pr-281-memtx-allocator.test.lua
@@ -0,0 +1,15 @@
+#!/usr/bin/env tarantool
+
+local tap = require('tap')
+
+local test = tap.test('test-run-pr-281-memtx-allocator')
+test:plan(1)
+
+-- Get rid of extra output in case of a failure.
+test.trace = false
+
+-- Verify from 'core = app' test.
+test:is(os.getenv('MEMTX_ALLOCATOR'), 'system',
+        'MEMTX_ALLOCATOR environment variable')
+
+os.exit(test:check() and 0 or 1)
diff --git a/test/engine/box.lua b/test/engine/box.lua
index 8558c9ac0..503b5dc9e 100644
--- a/test/engine/box.lua
+++ b/test/engine/box.lua
@@ -15,3 +15,6 @@ box.cfg{
 }
 
 require('console').listen(os.getenv('ADMIN'))
+
+_G.memtx_allocator = os.getenv('MEMTX_ALLOCATOR')
+_G.protected_globals = {'memtx_allocator'}
diff --git a/test/engine/engine.cfg b/test/engine/engine.cfg
index 3bfc44dcf..090860961 100644
--- a/test/engine/engine.cfg
+++ b/test/engine/engine.cfg
@@ -11,7 +11,10 @@
     },
     "gh-5216-small_alloc_factor.test.lua" : {
         "memtx": {"engine": "memtx"}
-    }
+    },
+    "test-run-pr-281-memtx-allocator-from-instance-file.test.lua": {},
+    "test-run-pr-281-memtx-allocator-from-test.test.lua": {},
+    "test-run-pr-281-memtx-allocator-from-nondefault.test.lua": {}
 }
 
 
diff --git a/test/engine/test-run-pr-281-memtx-allocator-from-instance-file.result b/test/engine/test-run-pr-281-memtx-allocator-from-instance-file.result
new file mode 100644
index 000000000..bc766c72f
--- /dev/null
+++ b/test/engine/test-run-pr-281-memtx-allocator-from-instance-file.result
@@ -0,0 +1,6 @@
+-- test-run result file version 2
+-- Verify from an instance file (box.lua).
+memtx_allocator == 'system' or memtx_allocator
+ | ---
+ | - true
+ | ...
diff --git a/test/engine/test-run-pr-281-memtx-allocator-from-instance-file.test.lua b/test/engine/test-run-pr-281-memtx-allocator-from-instance-file.test.lua
new file mode 100644
index 000000000..89e93bce3
--- /dev/null
+++ b/test/engine/test-run-pr-281-memtx-allocator-from-instance-file.test.lua
@@ -0,0 +1,2 @@
+-- Verify from an instance file (box.lua).
+memtx_allocator == 'system' or memtx_allocator
diff --git a/test/engine/test-run-pr-281-memtx-allocator-from-nondefault.result b/test/engine/test-run-pr-281-memtx-allocator-from-nondefault.result
new file mode 100644
index 000000000..fa8be39ab
--- /dev/null
+++ b/test/engine/test-run-pr-281-memtx-allocator-from-nondefault.result
@@ -0,0 +1,43 @@
+-- test-run result file version 2
+test_run = require('test_run').new()
+ | ---
+ | ...
+
+test_run:cmd('create server nondefault with script="engine/test-run-pr-281-nondefault.lua"')
+ | ---
+ | - true
+ | ...
+test_run:cmd('start server nondefault')
+ | ---
+ | - true
+ | ...
+test_run:cmd('switch nondefault')
+ | ---
+ | - true
+ | ...
+
+-- Verify from non-default instance.
+x = os.getenv('MEMTX_ALLOCATOR')
+ | ---
+ | ...
+x == 'system' or x
+ | ---
+ | - true
+ | ...
+
+test_run:cmd('switch default')
+ | ---
+ | - true
+ | ...
+test_run:cmd('stop server nondefault')
+ | ---
+ | - true
+ | ...
+test_run:cmd('cleanup server nondefault')
+ | ---
+ | - true
+ | ...
+test_run:cmd('delete server nondefault')
+ | ---
+ | - true
+ | ...
diff --git a/test/engine/test-run-pr-281-memtx-allocator-from-nondefault.test.lua b/test/engine/test-run-pr-281-memtx-allocator-from-nondefault.test.lua
new file mode 100644
index 000000000..f669c5e30
--- /dev/null
+++ b/test/engine/test-run-pr-281-memtx-allocator-from-nondefault.test.lua
@@ -0,0 +1,14 @@
+test_run = require('test_run').new()
+
+test_run:cmd('create server nondefault with script="engine/test-run-pr-281-nondefault.lua"')
+test_run:cmd('start server nondefault')
+test_run:cmd('switch nondefault')
+
+-- Verify from non-default instance.
+x = os.getenv('MEMTX_ALLOCATOR')
+x == 'system' or x
+
+test_run:cmd('switch default')
+test_run:cmd('stop server nondefault')
+test_run:cmd('cleanup server nondefault')
+test_run:cmd('delete server nondefault')
diff --git a/test/engine/test-run-pr-281-memtx-allocator-from-test.result b/test/engine/test-run-pr-281-memtx-allocator-from-test.result
new file mode 100644
index 000000000..2e44ea5fd
--- /dev/null
+++ b/test/engine/test-run-pr-281-memtx-allocator-from-test.result
@@ -0,0 +1,9 @@
+-- test-run result file version 2
+-- Verify from 'core = tarantool' test.
+x = os.getenv('MEMTX_ALLOCATOR')
+ | ---
+ | ...
+x == 'system' or x
+ | ---
+ | - true
+ | ...
diff --git a/test/engine/test-run-pr-281-memtx-allocator-from-test.test.lua b/test/engine/test-run-pr-281-memtx-allocator-from-test.test.lua
new file mode 100644
index 000000000..5c0ed1218
--- /dev/null
+++ b/test/engine/test-run-pr-281-memtx-allocator-from-test.test.lua
@@ -0,0 +1,3 @@
+-- Verify from 'core = tarantool' test.
+x = os.getenv('MEMTX_ALLOCATOR')
+x == 'system' or x
diff --git a/test/engine/test-run-pr-281-nondefault.lua b/test/engine/test-run-pr-281-nondefault.lua
new file mode 100644
index 000000000..618b1f6b8
--- /dev/null
+++ b/test/engine/test-run-pr-281-nondefault.lua
@@ -0,0 +1,7 @@
+#!/usr/bin/env tarantool
+
+box.cfg({
+    listen = os.getenv('LISTEN'),
+})
+
+require('console').listen(os.getenv('ADMIN'))
diff --git a/test/unit/CMakeLists.txt b/test/unit/CMakeLists.txt
index eb5c8347e..25eef2f41 100644
--- a/test/unit/CMakeLists.txt
+++ b/test/unit/CMakeLists.txt
@@ -272,3 +272,6 @@ add_executable(popen-child popen-child.c)
 
 add_executable(popen.test popen.c core_test_utils.c)
 target_link_libraries(popen.test misc unit core)
+
+add_executable(test_run_pr_281_memtx_allocator.test
+               test_run_pr_281_memtx_allocator.c unit)
diff --git a/test/unit/test_run_pr_281_memtx_allocator.c b/test/unit/test_run_pr_281_memtx_allocator.c
new file mode 100644
index 000000000..4bc8f7c68
--- /dev/null
+++ b/test/unit/test_run_pr_281_memtx_allocator.c
@@ -0,0 +1,20 @@
+#include <stdbool.h>
+#include <stdlib.h>
+#include <string.h>
+
+#include "unit.h"
+
+int
+main(int argc, char *argv[])
+{
+	plan(1);
+	header();
+
+	/* Verify from a 'core = unittest' test. */
+	bool ok = !strcmp("system", getenv("MEMTX_ALLOCATOR"));
+	ok(ok, "MEMTX_ALLOCATOR environment variable");
+
+	int rc = check_plan();
+	footer();
+	return rc;
+}
diff --git a/test/unit/test_run_pr_281_memtx_allocator.result b/test/unit/test_run_pr_281_memtx_allocator.result
new file mode 100644
index 000000000..cd633f21a
--- /dev/null
+++ b/test/unit/test_run_pr_281_memtx_allocator.result
@@ -0,0 +1,4 @@
+1..1
+	*** main ***
+ok 1 - MEMTX_ALLOCATOR environment variable
+	*** main: done ***
```

The same as a file: [test-run-pr-281-testing.patch.txt](https://github.com/tarantool/test-run/files/6212213/test-run-pr-281-testing.patch.txt).
</details>

Korablev77 (approved):
LGTM

Totktonada:
Updated the test-run submodule in tarantool in the following commits: [2.8.0-167-g6d6a153b0](https://github.com/tarantool/tarantool/commit/6d6a153b007652254683e0ed0da73ea256e49a98), 2.7.1-149-gfcb4f06af, 2.6.2-146-g1090c7cdb, 1.10.9-82-g467a9e3e6.

PR #283, opened 2021-03-26
test: add initial unit tests and code coverage support

ligurio:


test/test-tarantool/call.test.py
```diff
@@ -0,0 +1,217 @@
+"""
+Original Tarantool's test box-py/call.test.py had a problem when output with
+running under Python 2 was not the same as with running under Python 3.
+
+Fixed in commit e97425e46a77a4ac9d102c5f88062cb37e5a8da4
+"test: make box-py/call.test.py compatible with Python 2 and 3".
```
Totktonada, line 6: Here the test aims to verify test-run, so I would mention 72f8b89332b552cf20b2143e6dae7fe5042a9f79, 110f681ac0a56e3016e9a95eb1db544e353fa812 and d421020eeb1b3c2ec66a49a3be0e8a363dbe8d3d. Problems around porting of the test to Python 3 are less relevant here (at least).

requirements-test.txt
```diff
@@ -1 +1,2 @@
 flake8==3.7.9
+hypothesis==4.57.0
```
Totktonada, line 2: How about relaxed requirements?

```
hypothesis >= 4.57, == 4.*
```

It seems, we don't even need [environment markers](https://pip.pypa.io/en/stable/reference/pip_install/#requirement-specifiers) to support Python 2.7 and 3.* both.

lib/utils.py
```diff
@@ -286,8 +286,8 @@ def xlog_rows(xlog_path):
     cmd = ['tarantoolctl', 'cat', xlog_path, '--format=json', '--show-system']
     with open(os.devnull, 'w') as devnull:
         process = subprocess.Popen(cmd, stdout=subprocess.PIPE, stderr=devnull)
-    for line in process.stdout.readlines():
-        yield json.loads(line)
+        for line in process.stdout.readlines():
+            yield json.loads(bytes_to_str(line))
```
Totktonada, line 290: The stderr pipe (one that points to `/dev/null`) will be referenced by the Popen instance and the file descriptor will be closed by the garbage collector. Since we don't use it directly in those two lines, I think it is okay (and even logical) to keep them outside of the `with` block.

test/unittest/test_lib_utils.py
```diff
@@ -0,0 +1,36 @@
...
+            s = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+            s.bind(('localhost', p))
+            s.listen(0)
+            return s
+        status = utils.check_port(port, False, ipv4=True, ipv6=False)
+        self.assertEqual(status, True)
+        s = open_socket(port)
+        status = utils.check_port(port, False, ipv4=True, ipv6=False)
+        s.close()
+        self.assertEqual(status, False)
```
Totktonada, line 32: (It is my pet peeve, so I can't pass away. It is not a review comment and not something against your change.)

To be honest, I would prefer to [put it to the torch](https://github.com/tarantool/test-run/issues/141). Not sure, thought, it'll be possible entirely: at least there is 1.10 w/o logging of an actual port and we likely will meet other problem on the way.

The test depends on the environment: if someone listens on such port, the test will fail. However the range that you choosen should more or less protect us from this.

It also will fail with some probability if many instances of the test will be run in parallel or even one-by-one (AFAIR, `TIME_WAIT` is around two minutes and we don't use `SO_REUSEADDR` / zero `SO_LINGER` here). Hopefully, we'll not run the test several times during a short time. However we should remember: the test is flaky by design. If it'll be problematic, we unlikely will be able to propose a full fix.

Anyway, I don't mind adding the test case. Just... it is my old fault: it would be better if I would drop this machinery in 2017, when I worked on test-run parallelization.

BTW, you can safely drop the `rais` parameter: it is not used anywhere.

test/unittest/test_lib_utils.py
```diff
@@ -0,0 +1,36 @@
...
+import lib.utils as utils
+
+
+class TestUtils(unittest.TestCase):
+    def test_extract_schema_from_snapshot(self):
+        snapshot_path = '00000000000000000003.snap'
+        v = utils.extract_schema_from_snapshot(snapshot_path)
+        # assertListEqual() is available since 3.1
+        is_equal = v == [u'version', 2, 3, 1]
+        self.assertTrue(is_equal)
```
Totktonada, line 17: Nit: Can we use `assertEqual()`?

Totktonada: Ah, missed the comment. However, nope, we can use `assertEqual()`: it is more general than `assertListEqual()`, but works for lists (on Python 2 too).

test/unittest/test_lib_utils.py
```diff
@@ -0,0 +1,36 @@
...
+
+import lib.utils as utils
+
+
+class TestUtils(unittest.TestCase):
+    def test_extract_schema_from_snapshot(self):
+        snapshot_path = '00000000000000000003.snap'
+        v = utils.extract_schema_from_snapshot(snapshot_path)
+        # assertListEqual() is available since 3.1
+        is_equal = v == [u'version', 2, 3, 1]
```
Totktonada, line 16: (Again, not about the test per se. Feel free to postpone.)

There are two points that I dislike about the API of the function:

1. It returns a unicode string unlike all other test-run code (which sticks to `str`).
2. It would be better to return just version tuple, like `(2, 3, 1)` -- quite usual for representing a version in Python. And it works nice with less-great comparisons.

The first could be resolved by using pyyaml instead of json:

```python
# (python 2)
>>> import json
>>> import yaml
>>> json.loads('["version", 1, 2, 3]')
[u'version', 1, 2, 3]
>>> yaml.safe_load('["version", 1, 2, 3]')
['version', 1, 2, 3]
```

However if we'll implement the second point, it is not necessary. It is quite simple:

```diff
diff --git a/lib/tarantool_server.py b/lib/tarantool_server.py
index 45c5148..a0ad6da 100644
--- a/lib/tarantool_server.py
+++ b/lib/tarantool_server.py
@@ -912,8 +912,8 @@ class TarantoolServer(Server):
         # Verify that the schema actually was not upgraded.
         if self.disable_schema_upgrade:
             expected_version = extract_schema_from_snapshot(self.snapshot_path)
-            actual_version = yaml.safe_load(self.admin.execute(
-                'box.space._schema:get{"version"}'))[0]
+            actual_version = tuple(yaml.safe_load(self.admin.execute(
+                'box.space._schema:get{"version"}'))[0][1:])
             if expected_version != actual_version:
                 color_stdout('Schema version check fails: expected '
                              '{}, got {}\n'.format(expected_version,
diff --git a/lib/utils.py b/lib/utils.py
index a92db07..f88ebb5 100644
--- a/lib/utils.py
+++ b/lib/utils.py
@@ -303,7 +303,7 @@ def extract_schema_from_snapshot(snapshot_path):
        "BODY": {"space_id": 272, "tuple": ["version", 2, 3, 1]}
      }
 
-    :returns: [u'version', 2, 3, 1]
+    :returns: (2, 3, 1)
     """
     BOX_SCHEMA_ID = 272
     for row in xlog_rows(snapshot_path):
@@ -311,7 +311,7 @@ def extract_schema_from_snapshot(snapshot_path):
            row['BODY']['space_id'] == BOX_SCHEMA_ID:
             res = row['BODY']['tuple']
             if res[0] == 'version':
-                return res
+                return tuple(res[1:])
     return None
```

test/unittest/test_lib_utils.py
ligurio: done

ligurio: works fine:
```
  File "test/unittest/test_lib_utils.py", line 15, in test_extract_schema_from_snapshot
    self.assertEqual(v, [u'version', 2, 3, 1])
AssertionError: (2, 3, 1) != ['version', 2, 3, 1]
```

test/unittest/test_lib_utils.py
ligurio: Applied patch, thanks!

test/unittest/test_lib_utils.py
ligurio: > BTW, you can safely drop the rais parameter: it is not used anywhere.

Didn't get it.

test/test-tarantool/call.test.py
ligurio: added

test/unittest/test_lib_utils.py
Totktonada: The second argument of the `check_port()` function is not used anywhere (neither in test-run, nor in tarantool's python tests).

requirements-test.txt
ligurio:
```
--- a/requirements-test.txt
+++ b/requirements-test.txt
@@ -1 +1,2 @@
 flake8==3.7.9
+hypothesis==4.*
```

test/unittest/test_lib_utils.py
ligurio: removed

ligurio:
rebased to master

requirements-test.txt
Totktonada: Seems legit, nice!

test/unittest/test_lib_utils.py
Totktonada: I meant to remove it from the source code, however, okay: it is out of scope here.

test/test-tarantool/call.test.py
```diff
@@ -0,0 +1,216 @@
+"""
+Tarantool's test box-py/call.test.py had a number of problem fixed in commits:
+72f8b89332b552cf20b2143e6dae7fe5042a9f79 ("Update tarantool-python to 0.7.1-1-g8847b8c"),
+110f681ac0a56e3016e9a95eb1db544e353fa812 ("python3: use string byte literal") and
+d421020eeb1b3c2ec66a49a3be0e8a363dbe8d3d ("python3: fix printing of called box command")
```
Totktonada, line 5: The first one is correct, but reverting of the second and third commits does not break the test case.

.gitignore
```diff
@@ -41,6 +41,7 @@ htmlcov/
 .coverage
```
Totktonada, line 41: > Unit test that pass tarantoolctl output to JSON serializer failed [4]
> with Python 3.5

This part of the commit message becomes a bit outdated and now it is more relevant to the second commit.

ligurio: updated
```
commit 73a78c309621bfc312dac3dd5e146c7e928b7267 (HEAD -> ligurio/add-unittest)
Author: Sergey Bronnikov <sergeyb@tarantool.org>
Date:   Fri Mar 26 10:45:12 2021 +0300

    test: add unit tests
    
    Tests written using unittest [1] and hypothesis [2] modules. We use
    hypothesis v4.57.0 because it is a latest version with Python 2.7
    support [3].
    
    How-to run:
    
    $ make test_unittest
    
    1. https://docs.python.org/3/library/unittest.html
    2. https://hypothesis.readthedocs.io/en/latest/
    3. https://hypothesis.readthedocs.io/en/latest/changes.html#v4-57-0

```

Totktonada: (Forgot to push it?)

test/test-tarantool/call.test.py
ligurio: Fixed

.gitignore
ligurio: Pushed.

lib/utils.py
ligurio: Done

lib/utils.py
```diff
@@ -305,6 +305,8 @@ def extract_schema_from_snapshot(snapshot_path):
 
     :returns: (2, 3, 1)
     """
+    if not os.path.exists(snapshot_path):
+        raise(OSError)
```
Totktonada, line 310: It is verified in lib/options.py::check() before running any test:

> $ ./test/test-run.py --snapshot xxx
>
> Path /home/alex/projects/tarantool-meta/r/t-6/xxx does not exist

Even if we'll remove the check (let's assume that we unintentionally broke test-run or that there is a race condition and the file can be removed during the testing), we'll get the following:

> <..backtrace..>
> [001] FileNotFoundError: [Errno 2] No such file or directory: '/home/alex/projects/tarantool-meta/r/t-6/xxx'

I guess you want this check, because of the unit test, but it worth to mention it in the commit message. At least one may understand this commit as fix of the actual test-run behaviour.

I removed all beforehand checks to hit into the new one and look how it'll be shown in the test-run output:

> <..backtrace..>
> [001] OSError

The error message is not descriptive. In fact, it raises a class instead of its instance. Proper way to raise the exception would look so (see [SO](https://stackoverflow.com/a/36077407/1598057) for the explanation):

```python
# <..somewhere at top of the file..>
if PY2:
    FileNotFoundError = IOError
# <...>
    if not os.path.exists(snapshot_path):
        raise FileNotFoundError(errno.ENOENT, os.strerror(errno.ENOENT), snapshot_path)
```

I also was wondered, why it is not raised by some `open(snapshot_path, 'rb')` call deeper by the code and I see: we read the file via tarantoolctl in `xlog_rows()`. So the check should be in this function: we can use it more in a future (I do that in my [WIP patch](https://github.com/tarantool/test-run/commit/5fbe2195d4d1d662d8553af668a91acc12483a5f) for example).

lib/utils.py
```diff
@@ -284,10 +284,11 @@ def xlog_rows(xlog_path):
         Assume tarantool and tarantoolctl is in PATH.
     """
     cmd = ['tarantoolctl', 'cat', xlog_path, '--format=json', '--show-system']
+    process = None
     with open(os.devnull, 'w') as devnull:
         process = subprocess.Popen(cmd, stdout=subprocess.PIPE, stderr=devnull)
     for line in process.stdout.readlines():
-        yield json.loads(line)
+        yield json.loads(bytes_to_str(line))
```
Totktonada, line 291: I would extract it from the 'test: add unit test' commit, because this change fixes the real problem. It worth to quote the documentation to make it clear that you solve the Python 3.5 problem here.

lib/utils.py
```diff
@@ -284,10 +284,11 @@ def xlog_rows(xlog_path):
         Assume tarantool and tarantoolctl is in PATH.
     """
     cmd = ['tarantoolctl', 'cat', xlog_path, '--format=json', '--show-system']
+    process = None
```
Totktonada, line 287: I don't see a reason for this line. Can you clarify, please?

test/test-tarantool/call.result
```diff
@@ -0,0 +1,599 @@
...
+
+call  myinsert (3, 'old', 2)
+- [3, 'old', 2]
+call  myinsert (3, 'old', 2)
+{
+    "error": {
+        "code": "ER_TUPLE_FOUND", 
+        "reason": "Duplicate key exists in unique index 'primary' in space 'tweedledum'"
+    }
+}
```
Totktonada, line 256: It gives the following miscompare on a recent tarantool version:

```diff
--- test/test-tarantool/call.result	2021-04-30 01:26:24.708845662 +0300
+++ test/var/rejects/test-tarantool/call.reject	2021-04-30 02:21:26.230637668 +0300
@@ -251,7 +251,7 @@
 {
     "error": {
         "code": "ER_TUPLE_FOUND", 
-        "reason": "Duplicate key exists in unique index 'primary' in space 'tweedledum'"
+        "reason": "Duplicate key exists in unique index \"primary\" in space \"tweedledum\" with old tuple - [3, \"old\", 2] and new tuple - [3, \"old\", 2]"
     }
 }
 space:update({3}, {{'=', 1, 4}, {'=', 2, 'new'}})
```

See https://github.com/tarantool/tarantool/commit/d11fb3061e15faf4e0eb5375fb8056b4e64348ae

You can use a 'push filter' / 'clear filter' test-run commands to preprocess the message (or do it manually in the Lua code, or just remove the field if you don't need it for testing of test-run).

Makefile
```diff
@@ -15,6 +15,9 @@ luacheck:
 test_integration:
 	$(PYTHON) test/test-run.py --force $(TEST_RUN_EXTRA_PARAMS)
 
-test: test_integration
+test_unittest:
+	$(PYTHON) -m unittest discover test/unittest/
 
-.PHONY: lint flake8 luacheck test test_integration
+test: test_integration test_unittest
```
Totktonada, line 26: I would run unit tests first (a kind of smoke testing).

test/test-tarantool/call.test.py
Totktonada: Reverting of those two commits still doesn't fail the test for me:

```
$ git checkout origin/ligurio/add-unittest # 9087358a6a1653a8cda450d5c884ce41aabaa1d0
$ git revert 110f681ac0a56e3016e9a95eb1db544e353fa812
$ git revert d421020eeb1b3c2ec66a49a3be0e8a363dbe8d3d
$ make test
<...>
[008] test-tarantool/call.test.py                                     [ pass ]
<...>
```

lib/utils.py
ligurio: moved to a separate commit

lib/utils.py
ligurio: It is actually not needed here. I added it because I'm always forget about Python scope rules and define variables before loops and `with` if I want to use variable outside of them.

P.S. I found similar useless definition in https://github.com/tarantool/test-run/blob/master/lib/utils.py#L243

Makefile
ligurio: agree, reordered dependencies

test/test-tarantool/call.result
ligurio: removed 
```
--- a/test/test-tarantool/call.result
+++ b/test/test-tarantool/call.result
@@ -247,13 +247,6 @@ call  space:delete (2,)
 
 call  myinsert (3, 'old', 2)
 - [3, 'old', 2]
-call  myinsert (3, 'old', 2)
-{
-    "error": {
-        "code": "ER_TUPLE_FOUND", 
-        "reason": "Duplicate key exists in unique index 'primary' in space 'tweedledum'"
-    }
-}
 space:update({3}, {{'=', 1, 4}, {'=', 2, 'new'}})
 ---
 - error: Attempt to modify a tuple field which is part of index 'primary' in space
```

lib/utils.py
ligurio: Update exception to `FileNotFoundError`

lib/utils.py
Totktonada: Note: The definition of `fh` in `print_unidiff()` is not useless: `open()` may raise an error, we'll catch it, continue execution of the function and access the variable below the try-except block.

lib/utils.py
Totktonada: I was stuck for a while trying to understand why everything works without the change (just because I have Python 3.9 installed on my system), so I still think that we should cite the json module documentation or mention how Python 3.5 and 3.6+ are different here in some other way. Sorry for nitpicking.

lib/utils.py
Totktonada: I still think that we should perform the check in the same function, where Popen() is called, (it is `xlog_rows()`), and still think that a reason of the change should be mentioned in the commit message.

lib/utils.py
ligurio: Updated commit message:
```
    lib: use bytes_to_str() in xlog_rows()
    
    Since Python 3.6+ in json.loads() object that should be deserialized can
    be of type bytes or bytearray. The input encoding should be UTF-8,
    UTF-16 or UTF-32 [1]. Patch follows up
    commit 395edeb6b743c4479a62dd2183062124973d2b2a
    ('python3: decouple bytes and strings') and it is a part of
    task with switching to Python 3.
    
    1. https://docs.python.org/3/library/json.html
    
    Follows up: #20
```

lib/utils.py
ligurio: Could you explain why we should place code to check the existence of a snapshot path to `xlog_rows()` if we can check it earlier in `extract_schema_from_snapshot()`?

Popen can raise exception `FileNotFound` only when the command not found:
```
--- a/lib/utils.py
+++ b/lib/utils.py
@@ -287,7 +287,7 @@ def xlog_rows(xlog_path):
 
         Assume tarantool and tarantoolctl is in PATH.
     """
-    cmd = ['tarantoolctl', 'cat', xlog_path, '--format=json', '--show-system']
+    cmd = ['tarantoolctl1', 'cat', xlog_path, '--format=json', '--show-system']
     with open(os.devnull, 'w') as devnull:
         process = subprocess.Popen(cmd, stdout=subprocess.PIPE, stderr=devnull)
     for line in process.stdout.readlines():
```
```
======================================================================
ERROR: test_extract_schema_from_snapshot (test_lib_utils.TestUtils)
----------------------------------------------------------------------
  File "/home/sergeyb/sources/MRG/tarantool/test-run/test/unittest/test_lib_utils.py", line 12, in test_extract_schema_from_snapshot
    v = utils.extract_schema_from_snapshot(snapshot_path)
  File "/home/sergeyb/sources/MRG/tarantool/test-run/lib/utils.py", line 315, in extract_schema_from_snapshot
    for row in xlog_rows(snapshot_path):
  File "/home/sergeyb/sources/MRG/tarantool/test-run/lib/utils.py", line 292, in xlog_rows
    process = subprocess.Popen(cmd, stdout=subprocess.PIPE, stderr=devnull)
  File "/usr/lib/python3.8/subprocess.py", line 854, in __init__
    self._execute_child(args, executable, preexec_fn, close_fds,
  File "/usr/lib/python3.8/subprocess.py", line 1702, in _execute_child
    raise child_exception_type(errno_num, err_msg, err_filename)
FileNotFoundError: [Errno 2] No such file or directory: 'tarantoolctl1'

----------------------------------------------------------------------
Ran 2 tests in 0.010s

FAILED (errors=1)
```


And Popen() cannot detect that path passed as a parameter doesn't exist. 
```
--- a/lib/utils.py
+++ b/lib/utils.py
@@ -287,7 +287,7 @@ def xlog_rows(xlog_path):
 
         Assume tarantool and tarantoolctl is in PATH.
     """
-    cmd = ['tarantoolctl', 'cat', xlog_path, '--format=json', '--show-system']
+    cmd = ['tarantoolctl', 'cat', 'path', '--format=json', '--show-system']
     with open(os.devnull, 'w') as devnull:
         process = subprocess.Popen(cmd, stdout=subprocess.PIPE, stderr=devnull)
     for line in process.stdout.readlines():
```
```
======================================================================
FAIL: test_extract_schema_from_snapshot (test_lib_utils.TestUtils)
----------------------------------------------------------------------
  File "/home/sergeyb/sources/MRG/tarantool/test-run/test/unittest/test_lib_utils.py", line 13, in test_extract_schema_from_snapshot
    self.assertEqual(v, (2, 3, 1))
AssertionError: None != (2, 3, 1)

----------------------------------------------------------------------
Ran 2 tests in 0.027s

FAILED (failures=1)
make: *** [Makefile:21: test_unittest] Error 1
```

So what is a reason to move condition with `if os.path.exists()` to a `xlog_rows()`?

test/test-tarantool/call.test.py
ligurio: I have reverted d421020eeb1b3c2ec66a49a3be0e8a363dbe8d3d and run call.test.py in test-run and box-py/call.test.py in tarantool repo. Both tests passed.

Seems the real fix for `box-py/call.test.py` was done in commit for tarantool-python (https://github.com/tarantool/tarantool-python/pull/186/commits/13b8dfd83be9d87e196bfefc909dbdfc4362192e) and in commit 45972c99a5607b74984edcdfdab5fd2a79c78eba "test: make output of box-py/call.test.py deterministic" (https://github.com/tarantool/tarantool/commit/45972c99a5607b74984edcdfdab5fd2a79c78eba)
and commit to tarantool-python is a single reason to keep that regression test for test-run.

lib/utils.py
Totktonada: The important thing is that it cannot by bytes in Python 3.5 (it is the same as 'since Python 3.6+ is can', but the accent differs), but anyway now it is much easier to spot the reason of the change. Thanks.

lib/utils.py
Totktonada: > And Popen() cannot detect that path passed as a parameter doesn't exist.

That's my key point. If we'll imagine that `xlog_rows()` uses more typical way to read a file (using built-in `open()`), it would return an error on its own. It is meaningful: this function reads the file, so it raises an error, when the file is not exists. It looks logical to lean on this in `extract_schema_from_snapshot()`: usage of Popen() is the implementation detail of `xlog_rows()`.

It would be more important, if we would use `xlog_rows()` from several places.

Anyway, that's not critical and I don't insist.

test/test-tarantool/call.test.py
Totktonada: Sure. But its commit hash in tarantool-python is https://github.com/tarantool/tarantool-python/commit/8847b8c8bd1092d87601987d858e3cb9ff54f9f6 (13b8dfd83be9d87e196bfefc909dbdfc4362192e is from the [PR #186](https://github.com/tarantool/tarantool-python/pull/186), it was rebased at pushing to master).

ligurio: True, thanks for pointing to this!
```
commit 67bfb309c23b8e35aecaa8ea33189fbcac5fe19f
Author: Sergey Bronnikov <sergeyb@tarantool.org>
Date:   Fri Mar 26 10:29:07 2021 +0300

    test: add regression test for box-py/call.test.py
    
    Added test that covers fix in tarantool-python library:
    8847b8c8bd1092d87601987d858e3cb9ff54f9f6
    ("python3: make json.dumps compatible with Python 2") [1].
    
    1. https://github.com/tarantool/tarantool-python/pull/186/commits/8847b8c8bd1092d87601987d858e3cb9ff54f9f6

```

lib/utils.py
ligurio: Moved to `xlog_rows()`:

```python
--- a/lib/utils.py
+++ b/lib/utils.py
@@ -1,3 +1,4 @@
+import errno
 import os
 import sys
 import collections
@@ -24,6 +25,9 @@ UNIX_SOCKET_LEN_LIMIT = 107
 PY3 = sys.version_info[0] == 3
 PY2 = sys.version_info[0] == 2
 
+if PY2:
+    FileNotFoundError = IOError
+
 if PY3:
     string_types = str,
     integer_types = int,
@@ -283,6 +287,8 @@ def xlog_rows(xlog_path):
 
         Assume tarantool and tarantoolctl is in PATH.
     """
+    if not os.path.exists(xlog_path):
+        raise FileNotFoundError(errno.ENOENT, os.strerror(errno.ENOENT), xlog_path)
     cmd = ['tarantoolctl', 'cat', xlog_path, '--format=json', '--show-system']
     with open(os.devnull, 'w') as devnull:
         process = subprocess.Popen(cmd, stdout=subprocess.PIPE, stderr=devnull)
```

test/test-tarantool/call.test.py
Totktonada: The test file still refers 13b8dfd83be9d87e196bfefc909dbdfc4362192e. I'll update it.

Makefile
```diff
@@ -1,5 +1,7 @@
 TEST_RUN_EXTRA_PARAMS?=
 PYTHON?=python
+MAKEFILE_PATH := $(abspath $(lastword $(MAKEFILE_LIST)))
+PROJECT_DIR := $(patsubst %/,%,$(dir $(MAKEFILE_PATH)))
```
Totktonada, line 4: It is not used anywhere.

.github/workflows/test.yml
```diff
@@ -58,3 +58,7 @@ jobs:
     - name: run regression tests
       run: |
        make test
+    - name: code coverage
+      run: |
+       PYTHON="coverage run" make test_unittest
+       make coverage_report
```
Totktonada, line 64: So, we run unit tests first, then all tests (including unit tests) — because 'coverage_report' target depends of 'test' target. But the statistics will be collected only from the first run of the unit tests.

Makefile
```diff
@@ -20,4 +22,7 @@ test_unittest:
 
 test: test_unittest test_integration
 
+coverage_report: test
+	coverage report -m
```
Totktonada, line 26: Should not we follow `$(PYTHON)` value here like in all other places.

Totktonada, line 26: Whether subprocesses are tracked? Are you plan to measuse coverage using only unit tests?

.github/workflows/test.yml
Totktonada, line 64: Are there plans to visually represent per-line coverage using, say, coveralls.io? It should be easy to implement and will be a way more useful comparing to just numbers.

Totktonada, line 64: I dislike such way to run collect / report coverage. Can it be as easy as `make coverage` or `make test COVERAGE=1`?

Totktonada:
Ready to push except the patch about coverage.

test/test-tarantool/call.test.py
ligurio: Updated.

Makefile
ligurio: removed both `MAKEFILE_PATH` and `PROJECT_DIR`

.github/workflows/test.yml
ligurio: Could you explain why you dislike this way? Why running `make test COVERAGE=1` is better than `make coverage_report`?
Below you suggested pushing code coverage data to coveralls.io because it is "more useful comparing to just numbers".
Why it is important how we run code coverage collecting process if we plan to use it in GH actions and not manually?

Totktonada: > Could you explain why you dislike this way?

It is not what I can remember or find somewhere in the Makefile.

> Why running `make test COVERAGE=1` is better than `make coverage_report`?

The first one doing everything (collecting + reporting), not just reporting.

> Why it is important how we run code coverage collecting process if we plan to use it in GH actions and not manually?

The ability to run manually is what I expect from a project that has a Makefile in the root directory. It is useful for debugging (say, looking why a test does not cover a particulal line of the code). I would not place rules that are solely for CI into the Makefile.

Makefile
ligurio: I don't see any reasons to add it. Everything works without it.

.github/workflows/test.yml
ligurio: Personally, I like the way when we gather code coverage for unit testing and system tests separately. But if we want to use Coveralls we need to combine coverage data for a common report so I updated patch for workflow and Makefile:
```
--- a/Makefile
+++ b/Makefile
@@ -1,3 +1,4 @@
+MAKEFILE_PATH := $(abspath $(lastword $(MAKEFILE_LIST)))
 TEST_RUN_EXTRA_PARAMS?=
 PYTHON?=python
 
@@ -13,11 +14,15 @@ luacheck:
        luacheck --config .luacheckrc .
 
 test_integration:
-       $(PYTHON) test/test-run.py --force $(TEST_RUN_EXTRA_PARAMS)
+       PYTHON="$(PYTHON)" test/test-run.py --force $(TEST_RUN_EXTRA_PARAMS)
 
 test_unittest:
        $(PYTHON) -m unittest discover test/unittest/
 
 test: test_unittest test_integration
 
+coverage:
+       make -f $(MAKEFILE_PATH) test
+       coverage report --skip-empty --precision=1
+
 .PHONY: lint flake8 luacheck test test_integration test_unittest
```

```
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -58,3 +58,8 @@ jobs:
     - name: run regression tests
       run: |
        make test
+    - name: code coverage
+      run: |
+       make coverage
+       pip install coveralls
+       coveralls
```

ligurio: Added command to execute coveralls tool in GH Actions, it is required to pass an auth token in an environment variable. Could you set it there?

https://coveralls-python.readthedocs.io/en/latest/usage/configuration.html

ligurio: now it is enough to execute `make coverage`

Makefile
ligurio: No, I have changed patch to gather code coverage data with all tests.
Also added a patch to enable `coverage.py` for subprocesses:
```
--- a/lib/tarantool_server.py
+++ b/lib/tarantool_server.py
@@ -864,6 +864,12 @@ class TarantoolServer(Server):
 
         # redirect stdout from tarantoolctl and tarantool
         os.putenv("TEST_WORKDIR", self.vardir)
+        if os.environ.get('PYTHON') == 'coverage run':
+            import coverage
+            coveragerc_path = os.path.join(os.path.abspath(self.sourcedir), ".coveragerc")
+            assert(os.path.exists(coveragerc_path))
+            os.environ['COVERAGE_PROCESS_START'] = os.path.join(coveragerc_path)
+            coverage.process_startup()
         self.process = subprocess.Popen(args,
                                         cwd=self.vardir,
                                         stdout=self.log_des,
```

Makefile
ligurio: revert back removing of `MAKEFILE_PATH`, it's needed to run tests with `coverage.py`

ligurio: the same for `PROJECT_DIR`, used to set `PYTHONPATH`

Makefile
Totktonada: I was in doubts about test-run workers (which are separate processes). Maybe it works by default, I don't know. Just asked.

Here we start a tarantool process, so I guess we should not track it in the coverage tool.

.github/workflows/test.yml
Totktonada: [GITHUB_TOKEN](https://docs.github.com/en/actions/reference/authentication-in-a-workflow) is always there. I added `COVERALLS_TOKEN` into secrets just in case.

Makefile
Totktonada: Actually I don't know, but [this](https://coverage.readthedocs.io/en/latest/subprocess.html#subprocess) note looks relevant:

> If you have subprocesses because you are using `multiprocessing`, the `--concurrency=multiprocessing` command-line option should take care of everything for you. See [Execution: coverage run](https://coverage.readthedocs.io/en/latest/cmd.html#cmd-run) for details.

requirements-test.txt
```diff
@@ -1,2 +1,4 @@
+coverage==4.5.3
```
Totktonada, line 1: 4.5.3 looks quite old. Why not 5.5?

.github/workflows/test.yml
```diff
@@ -58,3 +58,11 @@ jobs:
...
       run: |
        make test
+    - name: code coverage
+      run: |
+       make coverage
+    - name: upload coverage data to coveralls.io
+      run: coveralls
+      env:
+        COVERALLS_REPO_TOKEN: ${{ secrets.COVERALLS_TOKEN }}
+        COVERALLS_PARALLEL: true
```
Totktonada, line 68: We should report coverage only for one matrix entry. We can use [matrix.include](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idstrategymatrix) syntax for this ([my example](https://github.com/tarantool/smtp/blob/322254c83169905bc43cb8a25972fa19a03f72ea/.github/workflows/testing.yml#L44-L57)).

Makefile
Totktonada: The relevant configuration option is [run.concurrency](https://coverage.readthedocs.io/en/latest/config.html#config-run-concurrency). As I see, it allows to pass several values at once, so I would try to set both `multiprocessing` and `gevent`. At least it worth to try and look how results will be changed. (I guess that `gevent` assumes `greenlet`. According to the [source code](https://github.com/nedbat/coveragepy/blob/01cbb8751f98e5a7de79699444cbc03647691616/coverage/collector.py#L94-L97), they cannot be set both. It seems, we can combine `multiprocessing` with something else.)

Makefile
```diff
@@ -13,11 +15,18 @@ luacheck:
...
-	$(PYTHON) test/test-run.py --force $(TEST_RUN_EXTRA_PARAMS)
+	PYTHONPATH=$(PROJECT_DIR) $(PYTHON) test/test-run.py --force $(TEST_RUN_EXTRA_PARAMS)
 
 test_unittest:
 	$(PYTHON) -m unittest discover test/unittest/
 
 test: test_unittest test_integration
 
+coverage:
+	PYTHON="$(PYTHON) -m coverage run" make -f $(MAKEFILE_PATH) test
```
Totktonada, line 26: This is not clear for me, to be honest. Can't we run it as `coverage run make -f $(MAKEFILE_PATH) test` (the documentation suggests this variant)? If so, is there a difference with given variant that uses `PYTHON="$(PYTHON) -m coverage run"`?

Makefile
Totktonada: Okay.

.github/workflows/test.yml
Totktonada: Okay, thanks!

Makefile
Totktonada: Okay.

.github/workflows/test.yml
Totktonada: Nice!

requirements-test.txt
ligurio: Hmm, this version was installed on my machine. updated to 5.*

.github/workflows/test.yml
ligurio: done

Makefile
ligurio: > Can't we run it as `coverage run make -f $(MAKEFILE_PATH) test` (the documentation suggests this variant)? 
when you want to enable profiling and gather code coverage you should use `coverage run` instead of `python`,
you don't use python to execute make with certain Makefile, but if you will try you will get an error:
```
sergeyb@pony:~/sources/MRG/tarantool/test-run$ python make -f Makefile 
python: can't open file 'make': [Errno 2] No such file or directory
sergeyb@pony:~/sources/MRG/tarantool/test-run$ 
```

>  If so, is there a difference with given variant that uses PYTHON="$(PYTHON) -m coverage run"?
AFAIR without $PYTHON it doesn't work. Let's try to remove $PYTHON:

```
--- a/Makefile
+++ b/Makefile
@@ -23,7 +23,7 @@ test_unittest:
 test: test_unittest test_integration
 
 coverage:
-       PYTHON="$(PYTHON) -m coverage run" make -f $(MAKEFILE_PATH) test
+       PYTHON="coverage run" make -f $(MAKEFILE_PATH) test
        coverage report
 
 clean:
sergeyb@pony:~/sources/MRG/tarantool/test-run$ make coverage 
PYTHON="coverage run" make -f /home/sergeyb/sources/MRG/tarantool/test-run/Makefile test
make[1]: Entering directory '/home/sergeyb/sources/MRG/tarantool/test-run'
coverage run -m unittest discover test/unittest/
E..
======================================================================
ERROR: test_extract_schema_from_snapshot (test_lib_tarantool_connection.TestTarantoolConnection)
----------------------------------------------------------------------
  File "/home/sergeyb/sources/MRG/tarantool/test-run/test/unittest/test_lib_tarantool_connection.py", line 8, in test_extract_schema_from_snapshot
    v = utils.extract_schema_from_snapshot(snapshot_path)
NameError: global name 'utils' is not defined

----------------------------------------------------------------------
Ran 3 tests in 0.036s

FAILED (errors=1)
```

ligurio: I was wrong, without $PYTHON it works fine:
```
--- a/Makefile
+++ b/Makefile
@@ -23,7 +23,7 @@ test_unittest:
 test: test_unittest test_integration
 
 coverage:
-       PYTHON="$(PYTHON) -m coverage run" make -f $(MAKEFILE_PATH) test
+       PYTHON="coverage run" make -f $(MAKEFILE_PATH) test
        coverage report
 
 clean:
```

Makefile
ligurio: > The relevant configuration option is run.concurrency. As I see, it allows to pass several values at once, so I would try to set both multiprocessing and gevent. 

`multiprocessing` conflicts with `gevent`, so I kept `gevent` only. BTW no changes in overall code coverage number.

```
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,5 +1,7 @@
 [run]
 branch = True
+concurrency =
+    gevent
 
 [report]
 precision = 1
```

Totktonada: > > Here we start a tarantool process, so I guess we should not track it in the coverage tool.
> 
> python tests uses python helpers like `admin` etc so why we shouldn't want to track it?

We execute Python code in tests within a worker process using the [exec()](https://docs.python.org/3/library/functions.html#exec) function in `PythonTest`. Not here. Here we start tarantool processes.

Totktonada: > `multiprocessing` conflicts with `gevent`, so I kept `gevent` only. BTW no changes in overall code coverage number.

Now I see that both are there. That was a mistake?

requirements-test.txt
```diff
@@ -1,2 +1,4 @@
+coverage==5.*
+coveralls==1.*
```
Totktonada, line 2: Missed it, when looked on this first time, but I would hold `coveralls==3.*`. I hope it'll be integrated with coveralls.io better.

Totktonada: I guess it is due to [dropping of Python 2 support](https://github.com/TheKevJames/coveralls-python/blob/master/CHANGELOG.md#200-2020-04-07), but now we run it only for 3.8.

Makefile
Totktonada: My question was why not just `coverage run make test`, but now I see: `coverage run` expects a python script as argument and does not accept a wrapper around python script. Okay so.

Makefile
Totktonada: Anyway, it seems, that the answer to my initial question is 'no': subprocesses are NOT tracked. I found that it is due to changing current directory to `test/`. So, the following patch does the work:

```diff
diff --git a/Makefile b/Makefile
index 4c5f72e..2b0361b 100644
--- a/Makefile
+++ b/Makefile
@@ -24,7 +24,7 @@ test: test_unittest test_integration
 
 coverage:
        PYTHON="coverage run" make -f $(MAKEFILE_PATH) test
-       coverage combine
+       coverage combine . test
        coverage report
 
 clean:
```

.coveragerc
```diff
@@ -0,0 +1,13 @@
...
+    gevent
+    multiprocessing
+
+[report]
+precision = 1
+include =
+    ./*
+omit =
+    lib/tarantool-python/*
+    lib/msgpack-python/*
```
Totktonada, line 13: Should we add `test/*` here as well? The test file test/unittest/test_lib_utils.py is in my report.

Makefile
ligurio: added `coverage combine $(PROJECT_DIR) $(PROJECT_DIR)/test`

ligurio: >> multiprocessing conflicts with gevent, so I kept gevent only. BTW no changes in overall code coverage number.

> Now I see that both are there. That was a mistake?

Possibly I added `gevent` back when it started to work.

requirements-test.txt
ligurio: updated
```
--- a/requirements-test.txt
+++ b/requirements-test.txt
@@ -1,4 +1,4 @@
 coverage==5.*
-coveralls==1.*
+coveralls==3.*
 flake8==3.7.9
 hypothesis==4.*
```

.coveragerc
ligurio: sure, added

```
--- a/.coveragerc
+++ b/.coveragerc
@@ -11,3 +11,4 @@ include =
 omit =
     lib/tarantool-python/*
     lib/msgpack-python/*
+    test/*

```

Makefile
Totktonada: Nice, thanks!

requirements-test.txt
ligurio: Moved coveralls package installation to GH actions workflow

Totktonada: From CI on Python 2:

> ERROR: Could not find a version that satisfies the requirement coveralls==3.* (from -r requirements-test.txt (line 2)) (from versions: 0.1, 0.1.1, 0.2, 0.3, 0.4, 0.4.1, 0.4.2, 0.4.3, 0.4.4, 0.5, 1.0a1, 1.0a2, 1.0b1, 1.0, 1.1, 1.2.0, 1.3.0, 1.4.0, 1.5.0, 1.5.1, 1.6.0, 1.7.0, 1.8.0, 1.8.1, 1.8.2, 1.9.0, 1.9.1, 1.9.2, 1.10.0, 1.11.0, 1.11.1)
> ERROR: No matching distribution found for coveralls==3.* (from -r requirements-test.txt (line 2))

I guess [environment markers](https://www.python.org/dev/peps/pep-0508/#environment-markers) should do the work:

```
coveralls==3.*; python_version > '3.4'
```

Totktonada: > Moved coveralls package installation to GH actions workflow

Or this way, okay.

Totktonada (approved):
LGTM. I have no objections anymore.

I'll look again just in case and push.

Totktonada:
Does not matter much here, but anyway:

Updated the test-run submodule in tarantool in the following commits: [2.9.0-45-g0f44ede23](https://github.com/tarantool/tarantool/commit/0f44ede23c154c2231cbb14e8d8dc1d29ce5a345), 2.8.1-23-g1b632d36c, 2.7.2-18-g9221e0e28, 1.10.10-10-gbc8463d67.

PR #290, opened 2021-04-05
Fix slowdown on Python 3

### LINKED ISSUE #279: python3: "sql" test suite passing slowed down
With python3:

```
time ./test-run.py sql/ --force
```

Result:
```
real	0m36.403s
user	1m15.925s
sys	0m17.432s
```

With python2:
```
python2 time ./test-run.py sql/ --force
```

Result:
```
real	0m12.975s
user	0m45.623s
sys	0m16.097s
```

Totktonada:
Don't parse command line arguments each time `Options()` is called.

The metaclass syntax was changed in Python 3 (see [1]) and it is tricky
to add a metaclass in a Python version agnostic way. Still possible, but 
the code would be hard to understand. So I get rid of the Singleton
metaclass and use \_\_new\_\_() to implement this logic.

Aside of the Option class, we had another singleton: Colorer. However it
is instanciated once (this instance is exposed as color_stdout) and it
does not matter, whether it is a singleton or not.

A brief testing on tarantool's test suite (with --long --force) gives me
~8 minutes after the fix instead of ~11 minutes before the fix. However
it worth to note that we have a lot of tests, which are marked as
fragile and so run without parallelization. Otherwise the difference
would be lower, I guess.

[1]: https://www.python.org/dev/peps/pep-3115/

Fixes #279

### DIFF
--- a/lib/colorer.py
+++ b/lib/colorer.py
@@ -1,7 +1,6 @@
 import os
 import re
 import sys
-from lib.singleton import Singleton
 
 
 # Use it to print messages on the screen and to the worker's log.
@@ -128,7 +127,6 @@ class Colorer(object):
     1. ftp://ftp.cs.utk.edu/pub/shuford/terminal/dec_vt220_codes.txt
     2. http://invisible-island.net/xterm/ctlseqs/ctlseqs.html
     """
-    __metaclass__ = Singleton
     fgcolor = {
         "black":     '0;30',
         "red":       '0;31',
--- a/lib/options.py
+++ b/lib/options.py
@@ -2,7 +2,6 @@ import os
 import sys
 import argparse
 from itertools import product
-from lib.singleton import Singleton
 
 from lib.colorer import color_stdout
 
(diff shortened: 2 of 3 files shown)


LeonidVas (approved):
Additionally, you can add a [link](https://www.python.org/dev/peps/pep-3115) on PEP
LGTM.

Totktonada:
> Additionally, you can add a [link](https://www.python.org/dev/peps/pep-3115) on PEP
> LGTM.

Good idea. Done.

ligurio (approved):
Thanks for the patch!

Change declares performance speedup, but there are no any numbers attached to check it. So I decided to measure overall time required to run tests for sql test suite (because initially performance degradation has been discovered with tests in that suite) and all tests with and without patch. 

`time ../../test/test-run.py --builddir=/home/sergeyb/sources/MRG/tarantool/build --vardir=/home/sergeyb/sources/MRG/tarantool/build/test/var --suite sql`

| iteration | wo patch | with patch |
|-----------------|:-------------------:|:---------------------:|
| 1                | 51,909s    | 16,175s         |
| 2               | 51,338s     | 16,295s       |
| 3               | 57,364s     | 16,358s      |
| 4               | 51,278s      | 16,267s       |
| 5               | 50,738s    | 16,564s       |

`time ../../test/test-run.py --builddir=/home/sergeyb/sources/MRG/tarantool/build --vardir=/home/sergeyb/sources/MRG/tarantool/build/test/var`

| iteration | wo patch | with patch |
|-----------------|:------------------:|:---------------------:|
| 1                | 9m35,228s | 7m23,025s |
| 2               | 9m32,369s | 7m3,614s |

LGTM

avtikhon (approved):
Timings checked on self-hosted and Github Actions hosts, patch LGTM.

Totktonada:
Thank you all for reviews, testing and measurements!

Totktonada:
Updated the test-run submodule in tarantool in [2.8.0-199-g8b3ec7b92](https://github.com/tarantool/tarantool/commit/8b3ec7b928a02a8cc44aca1911cc5d665cf5133d), 2.7.1-182-gafc84a5d2, 2.6.2-178-geedd55789, 1.10.9-110-gef7938933.

Issue #277, opened 2021-03-22
Add memory usage profiling for running tests


avtikhon:
Need to add routines which will collect memory usage in runtime for tests. It is needed to resolve main issue
https://github.com/tarantool/tarantool-qa/issues/98

RSS memory usage file to parse:
/sys/fs/cgroup/memory/memory.limit_in_bytes

Style example of implementation:
https://github.com/tarantool/test-run/blob/5baa4f080dc82cd45bdc619743504a3c7203dd0d/lib/utils.py#L215-L233

Totktonada:
I would highlight that we need to collect RSS per-test × per-configuration and do it always, not under an option. Othwerwise we'll not spot problematic tests.

Collecting just total RSS across the system does not work when we run several tests in parallel. It also don't work when you serve Web in a browser, while testing is going.

----

Regarding implementation. We need to track particular processes and use knowledge about a test name and configuration (BTW, we also can track a server name to spot which server consumes much memory). Python's Popen implementation does not save rusage: it uses waitpid(), not wait3() / wait4().  So we unable to catch a peak RSS after a process termination. Sadly, it is not reported in `/proc/<..pid..>/status` in `VmHWM` for a zombie process.

(Just raw idea: /usr/bin/time uses wait4() and can report peak RSS. Can we just wrap or subclass subprocess.Popen() with a wrapper that calls a necessary command under `/usr/bin/time -f %M <...>`, parses output and set it to a field in the process handler after the process termination? NB: Don't do that under --gdb, --lddb, --gdb-server, --valgrind, --strace.)

So we should collect RSS (`VmHWM` or `VmRSS` from `/proc/<..pid..>/status`) periodically. We can use `VmHWM` (maximum RSS) for app and unittest tests, but we should use `VmRSS` for tarantool tests, because the default server may be reused between several tests.

When a tarantool process is run, we should register it in a watcher with necessary metainformation: pid, suite, test name, configuration name, server name, some unique run identifier (see #255). The watcher should periodically collect information about registered processes. When a process is gone, unregister it (I guess we can assume that PIDs are rarely reused). After testing the collected information may be reported on the terminal. We can also write all collected samples into logs (to allow in-depth analysis).

PR #292, opened 2021-04-13
Add RSS memory status collecting routine to get RSS per tests and workers

avtikhon:
Patch set consists of 2 commits:

1. Add RSS memory status collecting routine
    
    Added RSS memory status collecting routine which parses file:

    `      /sys/fs/cgroup/memory/memory.stat`

    for RSS value and run this routine after each sent test lua command in
    listeners. All results it collects in main tests worker log file:

    `      var/log/<worker name>.log`

    For this routine use added test-run new option:

    `      --collect-statistics`

    which should be used to use this routine.

    Closes tarantool/tarantool-qa#98

2. Enable logging TAP tests on each testing condition

    Found that TAP test do not print any information on tested conditions
    in tests worker log file. It can be used to check test stages for
    data written into this log file.

### DIFF
--- a/lib/options.py
+++ b/lib/options.py
@@ -271,6 +271,14 @@ class Options(object):
                 default=os.environ.get("MEMTX_ALLOCATOR", "small"),
                 help="""Memtx allocator type for tests""")
 
+        parser.add_argument(
+                "--collect-statistics",
+                dest="collect_statistics",
+                action="store_true",
+                default=False,
+                help="""Collect memory/disk host statistics usage.
+                Default: false.""")
+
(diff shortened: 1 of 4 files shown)


Totktonada:
I [clarified](https://github.com/tarantool/test-run/issues/277#issuecomment-833248499) the task: I want a built-in tool that will spot tests that use large amount of memory and will be enabled always: locally and in CI. So it should not be under an option and should be cheap enough (not after each command).

Sure, we may want a tool that will allow in-depth analysis, but it is not my primary wish. It is also possible that we'll get enough information in test-run logs after #298: we'll see commands and samples on the same timeline.

Issue #286, opened 2021-04-02
Collect tests timings


Totktonada:
I want to teach test-run to collect test timings and report several largest ones after testing. This way we can spot tests that're near to [limits](https://github.com/tarantool/test-run/blob/e698240f6a2034138c102e5ce627a46343b6e5ad/lib/options.py#L206-L212) (at least in some environments) and reduce them.

My criterias:

- Collect timings always (not under an option, locally and in CI both).
- Per-test × per-configuration statistics.

Wishful properties:

- Not only on Linux, but on Mac OS as well.
- Distiguish several runs of the same test (see #255).

What better to do with so called 'long_run' tests? Maybe mark them explicitly in the output?

Totktonada:
Clarified the task (see the description above).

Totktonada:
The similar issue in luatest: https://github.com/tarantool/luatest/issues/128

PR #294, opened 2021-04-27
Add test timings collection

avtikhon:
Added ability to show tests timings during tests run on screen output
right after the test result pass or fail in format
```
  ... [ pass ] [ <float timing in secs with 2 digits after point> ]
```
And dump it to log file in format:
```
  TEST:CONF TIMING: <full test name>:<test conf> <timing in secs>
```
Closes tarantool/test-run#286

### DIFF
--- a/lib/test.py
+++ b/lib/test.py
@@ -5,11 +5,13 @@ import pprint
 import re
 import shutil
 import sys
+import time
 import traceback
 from functools import partial
 from hashlib import md5
 
 from lib import Options
+from lib.colorer import color_log
 from lib.colorer import color_stdout
 from lib.utils import assert_bytes
 from lib.utils import non_empty_valgrind_logs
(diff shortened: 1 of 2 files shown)


lib/test.py
```diff
@@ -242,7 +247,7 @@ def run(self, server):
               self.is_equal_result and
               self.is_valgrind_clean):
             short_status = 'pass'
-            color_stdout("[ pass ]\n", schema='test_pass')
+            color_stdout("[ pass ] [ {} ]\n" . format(test_timing), schema='test_pass')
```
Totktonada, line 250: There was agreement to don't make test-run output too wide to allow split the screen vertically on a laptop.

Another point here is that we're interested in large timings, not all. It would be nice to see a list of longest tests after end of testing. It seems, it is done in PR #298, which actually should resolve #286 (BTW, I updated the description of the task). Does PR #298 supersedes this one?

Issue #293, opened 2021-04-14
Binary value in result leads to crashing of sql-tap tests


ImeevMA:
How to reproduce:

1) Add a new test that returns binary value to sql-tap tests. For example:
```
diff --git a/test/sql-tap/sql-errors.test.lua b/test/sql-tap/sql-errors.test.lua
index 469193a2b..0ca58c2ff 100755
--- a/test/sql-tap/sql-errors.test.lua
+++ b/test/sql-tap/sql-errors.test.lua
@@ -1,6 +1,6 @@
 #!/usr/bin/env tarantool
 local test = require("sqltester")
-test:plan(72)
+test:plan(73)
 
 test:execsql([[
 	CREATE TABLE t0 (i INT PRIMARY KEY, a INT);
@@ -780,4 +780,12 @@ test:do_catchsql_test(
 		-- </sql-errors-2.9>
 	})
 
+test:do_execsql_test(
+	"sql-errors-3.0",
+	[[
+		SELECT RANDOMBLOB(5);
+	]], {
+		1
+	})
+
 test:finish_test()
```

2) Run test. For example:
```
./test-run.py sql-tap/sql-errors.test.lua -j1
```

Result:
```
Started ./test-run.py sql-tap/sql-errors.test.lua -j1
Running in parallel with 1 workers

Timeout options:
-------------------
REPLICATION_SYNC_TIMEOUT: 100
TEST_TIMEOUT:             110
NO_OUTPUT_TIMEOUT:        120

Collecting tests in 'app'            (Found 0   tests): application server tests.
Collecting tests in 'app-tap'        (Found 0   tests): application server tests (TAP).
Collecting tests in 'box'            (Found 0   tests): Database tests.
Collecting tests in 'box-py'         (Found 0   tests): legacy python tests.
Collecting tests in 'box-tap'        (Found 0   tests): Database tests with #! using TAP.
Collecting tests in 'engine'         (Found 0   tests): tarantool multiengine tests.
Collecting tests in 'engine_long'    (Found 0   tests): tarantool engine stress tests.
Collecting tests in 'long_run-py'    (Found 0   tests): long running tests.
Collecting tests in 'replication'    (Found 0   tests): tarantool/box, replication.
Collecting tests in 'replication-py' (Found 0   tests): tarantool/box, replication.
Collecting tests in 'small'          (Found 0   tests): libsmall unit tests.
Collecting tests in 'sql'            (Found 0   tests): sql tests.
[...]

PR #297, opened 2021-04-30
Add error log for files with non-unicode symbol 

VitaliyaIoffe:
If a temporary test result file contains a non-unicode symbol, it will raise
additional logging and will fail the test.

Closes: #293

### DIFF
--- a/lib/test.py
+++ b/lib/test.py
@@ -341,12 +341,12 @@ class Test(object):
 
     def check_tap_output(self):
         """ Returns is_tap, is_ok """
-        with open(self.tmp_result, 'r') as f:
-            content = f.read()
-        tap = pytap13.TAP13()
         try:
+            with open(self.tmp_result, 'r') as f:
+                content = f.read()
+            tap = pytap13.TAP13()
             tap.parse(content)
-        except ValueError as e:
+        except (ValueError, UnicodeDecodeError) as e:
             color_stdout('\nTAP13 parse failed (%s).\n' % str(e),
                          schema='error')
             color_stdout('\nNo result file (%s) found.\n' % self.result,


lib/pytap13.py
```diff
@@ -17,43 +17,37 @@
...
-    from StringIO import StringIO
-except ImportError:
-    # Python 3
-    from io import StringIO
+from io import BytesIO
 
 import yaml
-from lib.utils import string_types
 
+from lib.utils import string_types, binary_regex, str_to_bytes, bytes_to_str
```
LeonidVas, line 24: Use a separate line for each import (I don't remember where these rules come from, but we follow it).

lib/pytap13.py
```diff
@@ -17,43 +17,37 @@
...
 
-RE_VERSION = re.compile(r"^\s*TAP version 13\s*$")
-RE_PLAN = re.compile(
+RE_VERSION = binary_regex(r"^\s*TAP version 13\s*$")
```
LeonidVas, line 26: Why not just `re.compile(b."....")`?

lib/pytap13.py
```diff
@@ -102,8 +96,8 @@ def _parse(self, source):
                     in_yaml = True
                     continue
 
-            on_top_level = not line.startswith('    ')
-            raw_line = line.rstrip('\n')
+            on_top_level = not line.startswith(str_to_bytes('    '))
```
LeonidVas, line 99: Why not just `(b'....')`? Here and below.

lib/test.py
```diff
@@ -9,15 +8,17 @@
 from functools import partial
 from hashlib import md5
 
+import gevent
```
LeonidVas, line 11: From https://github.com/tarantool/tarantool/wiki/Code-review-procedure#general-coding-points-to-check : "Don't do a change if it is not related to the patch."
Shuffling the import order is not related to the problem solved in the patch. Here and below.

lib/utils.py
```diff
@@ -351,3 +355,8 @@ def str_to_bytes(s):
     if PY2:
         return s
     return s.encode('utf-8')
+
+
+def binary_regex(s, flags=0):
```
LeonidVas, line 360: I don't think this function is really useful for two reasons:
1) It adds 3 unnecessary function calls (`binary_regex`, `str_to_bytes` and `encode`) and doesn't reduce the amount of code.
2) It's so simple, just write `re.compile (b" ")`.

LeonidVas (changes requested):
Hi! Thank you for the patch.
About the commit title:
Maybe something like: "Switch to using binary data in tests" is better? And I propose to move the description of the problem with decoding to the commit message.
About the commit message:
- From https://github.com/tarantool/tarantool/wiki/Code-review-procedure#commit-message : "Commit message should fit 72 symbols line width."
- "Fixes: #293"

Have you check that it works fine in both python2 and python3?

Also see several comments bellow:

lib/pytap13.py
VitaliyaIoffe: Binary strings with the prefix "b" are allowed only in python3. We discussed this approach with additional function with @Totktonada

lib/utils.py
VitaliyaIoffe: As I said above it is not so easy due to the "rb" prefix. Our patterns are raw strings. Either we should avoid raw strings and use b-prefix, or use this method.

lib/test.py
VitaliyaIoffe: Got it
Thank you

lib/pytap13.py
VitaliyaIoffe: Maybe I didn't understand @Totktonada correctly. He said that there was a discussion about strings in the test-run and how to use them to support both python's versions. Consequently, you don't use literals associated with unicode and binary strings. So, I used str_to_bytes function. I was wrong?

lib/pytap13.py
VitaliyaIoffe: Ok. 
Thank you

lib/pytap13.py
VitaliyaIoffe: There is "br" literal in python 2.7. Fixed.

lib/pytap13.py
VitaliyaIoffe: Fixed

Totktonada:
(Related to several discussions above.) `str_to_bytes()` and `bytes_to_str()` are general purpose functions. If we can make the code easier for the byte array literal case and don't use those functions here, it looks okay for me, at least on the first glance.

VitaliyaIoffe:
Yes, I've checked it with python2 and python3

Totktonada:
Please, drop `broken_unicode.result`: so the change will be tested (see https://github.com/tarantool/test-run/issues/293#issuecomment-819601782). Or, better, add another similar test without a result file.

lib/pytap13.py
```diff
@@ -174,13 +169,13 @@ def _parse(self, source):
         if len(self.tests) != self.tests_planned:
             comment = 'DIAG: Expected %s tests, got %s' % \
                 (self.tests_planned, len(self.tests))
-            self.tests.append(Test('not ok', len(self.tests), comment=comment))
+            self.tests.append(Test(b'not ok', len(self.tests), comment=comment))
 
     def parse(self, source):
         if isinstance(source, string_types):
-            self._parse(StringIO(source))
+            self._parse(BytesIO(source))
```
Totktonada, line 176: (Not a real review comment, just note. I'll give more thoughs a bit later.)

`string_types` is `bytes` and `unicode` on Python 2 and just `str` (unicode string) on Python 3. Let's consider Python 3 case. BytesIO accepts only `bytes`, not `str`:

```python
>>> from io import BytesIO
>>> BytesIO('')
  File "<stdin>", line 1, in <module>
TypeError: a bytes-like object is required, not 'str'
```

So the if-condition (`isinstance(source, string_types)`) contradicts with the `BytesIO(source)` expression.

Totktonada:
(Sorry, I'm not good in concise explanations, but I'm always in attempts to do this better...)

There are several ideas that I keep in the mind here:

1. It is nice, when things just work, without artificial limitations. So, it would be good to support non-utf-8 data in TAP 13 results.
2. OTOH, complexity (maintainability) of the implementation matters. I think we should keep pieces of code that work with binary data quite dense and don't spread them much across APIs. (See 395edeb6b743c4479a62dd2183062124973d2b2a about the idea.)

To be honest, I don't know what we should choose here. I'll share several more thoughts around and will take timeout to think more.

* The `Test` class contains both binary and string data after the patch. It does not look nice. Moreover, the class is used outside of pytap13.py that descreases locality of the binary data.
* The pytap13.py code mixes byte literals and string literals after the patch and it looks error prone. It is not easy to verify correctness. Let's consider line 152 (157 on `master`): `Test('not ok', <...>)` — it'll raise `ValueError` from `str_to_bytes()`.
* If we'll choose to support only UTF-8 TAP 13 results, we should give a good error message instead of traceback and `UnicodeDecodeError`. It'll push a test author to either keep a result file or to make test cases description human readable. Not so bad.
* I had a thought to encode non-UTF-8 data in a test case description automatically in the `tap` built-in module, but it does not follow perl's Test::More behaviour.
* TAP 13 may contain YAML blocks and our YAML encoder uses `!!binary` tag with base64 encoding to represent non-printable or non-utf-8 characters. So the only piece of data that may break UTF-8 encoding is a test case description.

There are several ways to more forward:

* Decide to don't support non-UTF-8 data in a TAP 13 output and give a meaningful error message instead.
* Decide to support non-UTF-8 data in a TAP 13 output, but rearrange TAP 13 related code to don't spread byte arrays much.
* Proceed with a patch like current one, but elaborate agreemtn how the code should be refactored in a future.

lib/pytap13.py
Totktonada: (In fact, after the patch, we never going into this branch, so it becomes a dead code.)

LeonidVas:
> (Sorry, I'm not good in concise explanations, but I'm always in attempts to do this better...)
> 
> There are several ideas that I keep in the mind here:
> 
> 1. It is nice, when things just work, without artificial limitations. So, it would be good to support non-utf-8 data in TAP 13 results.
> 2. OTOH, complexity (maintainability) of the implementation matters. I think we should keep pieces of code that work with binary data quite dense and don't spread them much across APIs. (See [395edeb](https://github.com/tarantool/test-run/commit/395edeb6b743c4479a62dd2183062124973d2b2a) about the idea.)
> 
> To be honest, I don't know what we should choose here. I'll share several more thoughts around and will take timeout to think more.
> 
> * The `Test` class contains both binary and string data after the patch. It does not look nice. Moreover, the class is used outside of pytap13.py that descreases locality of the binary data.
> * The pytap13.py code mixes byte literals and string literals after the patch and it looks error prone. It is not easy to verify correctness. Let's consider line 152 (157 on `master`): `Test('not ok', <...>)` — it'll raise `ValueError` from `str_to_bytes()`.
> * If we'll choose to support only UTF-8 TAP 13 results, we should give a good error message instead of traceback and `UnicodeDecodeError`. It'll push a test author to either keep a result file or to make test cases description human readable. Not so bad.
> * I had a thought to encode non-UTF-8 data in a test case description automatically in the `tap` built-in module, but it does not follow perl's Test::More behaviour.
> * TAP 13 may contain YAML blocks and our YAML encoder uses `!!binary` tag with base64 encoding to represent non-printable or non-utf-8 characters. So the only piece of data that may break UTF-8 encoding is a test case description.
> 
> There are several ways to more forward:
> 
> * Decide to don't support non-UTF-8 data in a TAP 13 output and give a meaningful error message instead.
> * Decide to support non-UTF-8 data in a TAP 13 output, but rearrange TAP 13 related code to don't spread byte arrays much.
> * Proceed with a patch like current one, but elaborate agreemtn how the code should be refactored in a future.

Just raw thoughts. If we decide to use base64, it will hide where the error was in the raw data (which byte was wrong). Although it may not matter.

lib/test.py
```diff
@@ -341,8 +341,15 @@ def tap_parse_print_yaml(self, yml):
 
     def check_tap_output(self):
         """ Returns is_tap, is_ok """
-        with open(self.tmp_result, 'r') as f:
-            content = f.read()
+        try:
+            with open(self.tmp_result, 'r') as f:
+                content = f.read()
+        except UnicodeDecodeError as e:
+            color_stdout('\nCannot read result file (%s).\n' % str(e),
```
LeonidVas, line 348: It might be nice to write something about “maybe you are using binary” or “try to use base64” ... Up to you.

LeonidVas (commented):
Hi! Thank you for the patch.
LGTM.

LeonidVas (approved):

lib/test.py
Totktonada: The actual message looks so:

> Cannot read result file ('utf-8' codec can't decode byte 0xc2 in position 27: invalid continuation byte).

However I would highlight that we only unable to use it as TAP13: file-to-file comparison still work. The 'TAP13 parse failed' message below looks good here (and also suggest `--update-result`). So, maybe, just move open+read under the try-except block and catch UnicodeDecodeError in the except clause?

Totktonada: BTW, it would be good to highlight the fact that the change is only about parsing output as TAP13 somewhere in the commit message. Now it is misleading a bit.

VitaliyaIoffe: @Totktonada fixed

lib/pytap13.py
Totktonada: Since we decided to raise a good error instead of support of non UTF-8 data in TAP13 parser, this discussion is not actual anymore.

lib/test.py
Totktonada: Thanks!

Totktonada (approved):
LGTM.

Totktonada:
Updated the test-run submodule in tarantool in [2.9.0-104-g7fa70b661](https://github.com/tarantool/tarantool/commit/7fa70b6619be693db3831411e306c6a62c35e655), 2.8.1-75-g029d66355, 2.7.2-62-g6f9ccad83, 1.10.10-25-g9b4acec8f.

Issue #280, opened 2021-03-22
optimization: write debug logs directly from workers


Totktonada:
Cited from https://github.com/tarantool/test-run/issues/279#issuecomment-804411674:

> Obvious optimization point: write per-worker logs directly, without passing over the queue. There is no reason to write them from the main process, because each worker has its own log.

Issue #247, opened 2020-12-06
color_log() output don't shown on the terminal with --debug when called from the main process


Totktonada:
We should tune the behaviour here:

https://github.com/tarantool/test-run/blob/e84355283f117e91964fe84eac70f26bb32d7f06/lib/colorer.py#L218-L219

Depending on whether --debug was passed.

Totktonada:
The obvious fix:

```diff
diff --git a/lib/__init__.py b/lib/__init__.py
index dcfe149..3bf3fca 100644
--- a/lib/__init__.py
+++ b/lib/__init__.py
@@ -2,6 +2,7 @@ import os
 import sys
 import shutil
 
+from lib.colorer import colorer_cfg
 from lib.options import Options
 from lib.tarantool_server import TarantoolServer
 from lib.unittest_server import UnittestServer
@@ -26,6 +27,8 @@ def setenv():
 def module_init():
     """ Called at import """
     args = Options().args
+    colorer_cfg(copy_log_to_stdout=args.debug)
+
     # Change the current working directory to where all test
     # collections are supposed to reside
     # If script executed with (python test-run.py) dirname is ''
diff --git a/lib/colorer.py b/lib/colorer.py
index 48ab560..091151f 100644
--- a/lib/colorer.py
+++ b/lib/colorer.py
@@ -7,6 +7,19 @@ import sys
 color_stdout = None  # = Colorer(); below the class definition
 
 
+cfg = dict()
+cfg_opts = ['copy_log_to_stdout']
+
+
+def colorer_cfg(**kwargs):
+    global cfg
+    for k in cfg_opts:
+        if k in kwargs:
+            cfg[k] = kwargs[k]
+        else:
+            raise ValueError('Unknown colorer option {}'.format(k))
+
+
 def color_log(*args, **kwargs):
     """ Print the message only to log file, not on the screen. The intention is
     use this function only for regular, non-error output that appears every run
@@ -15,7 +28,7 @@ def color_log(*args, **kwargs):
     exceptional circumstances) from the screen, because such details especially
     useful with CI bots.
     """
-    kwargs['log_only'] = True
+    kwargs['log_only'] = not cfg['copy_log_to_stdout']
     color_stdout(*args, **kwargs)
```

The logs from the main process still not written to any log file. It should be fixed here or should be tracked in a separate issue.

Totktonada:
[...]

PR #298, opened 2021-05-03
Add RSS statistics and timings collecting

### LINKED ISSUE #277: Add memory usage profiling for running tests

### LINKED ISSUE #286: Collect tests timings

avtikhon:
Patch set consists of 3 commits:

1. sampler: add simple sampling infrastructure
    
Track tarantool and unit test executables that are run using test-run
with metainformation: worker, test, test configuration and server name.
    
Add a function that will be called each 0.1 second for each tracked
process.
    
The implementation tracks non-default servers and re-register default
servers that executes several tests ('core = tarantool' case).
    
Part of #277

2. Add RSS statistics collecting.

Found that some tests may fail due to lack of memory. Mostly it happens
in CI on remote hosts. To be able to collect memory used statistic
decided to add RSS memory status collecting routine get_proc_stat_rss()
which parses files:
```
      /proc/<worker pid>/status
```
for RSS value 'VmRSS' which is size of memory portions. It contains the
three following parts (VmRSS = RssAnon + RssFile + RssShmem) [1]:
```
      RssAnon - size of resident anonymous memory
      RssFile - size of resident file mappings
      RssShmem - size of resident shmem memory (includes SysV shm, mapping of
                 tmpfs and shared anonymous mappings)
```
Decided that the best way for CI not to run this RSS collecting routine
for each sent command from tests tasks, but to run it after the test
task started each 0.1 second delay, to collect its maximum RSS value
during task run. This delay used to run routines in 'SamplerWatcher'
listener. Also found that delay of 0.1 sec is completely enough to catch
RSS use increase, due to tested check:
```
      tarantool> require('clock').bench(function() local t = {} for i = 1, 1024^2 * 100 do t[i] = true end end)
```
Which checked that 100 Mb of data allocated in seconds:
```
      - on CI test host: 3.153877479
      - on local fast host: 0.54504489
```
The main idea is to check all test depend processes running at some
point in time and choose maximum RSS reached value by it. For it used
'_sample_process()' routine which gets RSS for each currently alive
process and '_sample()' routine which counts sum of each task alive
processes RSS and checks if this value is bigger than previously
saved for the current task. Both routines are in 'Sampler()' class
which is called by 'process_timeout()' routine from 'SamplerWatcher'
listener.

Also used print_statistics() routine in listener 'StatisticsWatcher'
which prints statistics to stdout after testing. It used to print
RSS usage for failed tasks and up to 10 most used it tasks. Created
new subdirectory 'statistics' in 'vardir' path to save statistics
files. The current patch uses it to save there 'rss.log' file with
RSS values per tested tasks in format:
```
      <test task name> <maximum RSS value>
```
Closes #277

[1]: https://www.kernel.org/doc/html/latest/filesystems/proc.html

3. Add tests duration collecting

Decided to collect tests run durations in standalone file and print to
stdout after testing finished. To stdout printing durations for failed
tasks and up to 10 most long tasks.

For durations collecting used listener 'StatisticsWatcher' which has
the following used routines:
```
      process_result()
        Using 'WorkerTaskResult' queue message to collect tasks that failed
        and final durations for finished tasks. Durations collecting routine
        added in the worker running tests. Its values using 'result_queue'
        passing through 'WorkerTaskResult' queue message into the listener.
    
      print_statistics() - statistics printing to stdout after testing.
        Prints durations for failed tasks and up to 10 most long tasks.
        We use standalone 'statistics' directory in 'vardir' path to save
        'durations.log' file with durations for each tested tasks in format:
    
          <test task name> <float duration in secs with 2 digits after point>
```
Closes #286

### DIFF
--- a/dispatcher.py
+++ b/dispatcher.py
@@ -167,8 +167,9 @@ class Dispatcher:
         self.statistics = StatisticsWatcher(log_output_watcher.get_logfile)
         self.artifacts = ArtifactsWatcher(log_output_watcher.get_logfile)
         output_watcher = OutputWatcher()
-        self.listeners = [self.statistics, log_output_watcher, output_watcher, self.artifacts,
-                          sampler.watcher]
+        self.listeners = [self.statistics, log_output_watcher, output_watcher, self.artifacts]
+        if sampler.is_enabled:
+            self.listeners.append(sampler.watcher)
         if watch_fail:
             self.fail_watcher = FailWatcher(self.terminate_all_workers)
             self.listeners.append(self.fail_watcher)
--- a/dispatcher.py
+++ b/dispatcher.py
@@ -34,6 +34,7 @@ except ImportError:
     from multiprocessing.queues import SimpleQueue
 
 from lib import Options
+from lib.sampler import sampler
 from lib.utils import set_fd_cloexec
 from lib.worker import WorkerTaskResult, WorkerDone
 from lib.colorer import color_stdout
@@ -121,7 +122,7 @@ class Dispatcher:
             self.result_queues.append(task_queue_disp.result_queue)
             self.task_queues.append(task_queue_disp.task_queue)
 
-        self.report_timeout = 1.0
+        self.report_timeout = 0.1
 
         self.statistics = None
         self.artifacts = None
@@ -166,7 +167,8 @@ class Dispatcher:
         self.statistics = StatisticsWatcher(log_output_watcher.get_logfile)
         self.artifacts = ArtifactsWatcher(log_output_watcher.get_logfile)
         output_watcher = OutputWatcher()
-        self.listeners = [self.statistics, log_output_watcher, output_watcher, self.artifacts]
+        self.listeners = [self.statistics, log_output_watcher, output_watcher, self.artifacts,
+                          sampler.watcher]
         if watch_fail:
             self.fail_watcher = FailWatcher(self.terminate_all_workers)
             self.listeners.append(self.fail_watcher)
@@ -416,6 +418,7 @@ class TaskQueueDispatcher:
         os.environ['TEST_RUN_TCP_PORT_END'] = str(tcp_port_range[1])
         color_stdout.queue = self.result_queue
         worker = self.gen_worker(worker_id)
+        sampler.set_queue(self.result_queue, worker_id, worker.name)
         worker.run_all(self.task_queue, self.result_queue)
 
     def add_worker(self, worker_id, tcp_port_range):
--- a/lib/sampler.py
+++ b/lib/sampler.py
@@ -1,0 +1,155 @@
+import os
+import sys
+import time
+
+from lib.colorer import color_log
+from lib.colorer import qa_notice
+from lib.utils import format_process
+
+
+if sys.version_info[0] == 2:
+    ProcessLookupError = OSError
+
+
+# Don't inherit BaseWorkerMessage to bypass cyclic import.
+class RegisterProcessMessage(object):
+    """Ask the sampler in the main test-run process to register
+       given process.
+    """
+    def __init__(self, worker_id, worker_name, pid, task_id, server_name):
+        self.worker_id = worker_id
+        self.worker_name = worker_name
+        self.pid = pid
+        self.task_id = task_id
+        self.server_name = server_name
+
+
+# Don't inherit BaseWatcher to bypass cyclic import.
+class SamplerWatcher(object):
+    def __init__(self, sampler):
+        self._sampler = sampler
+        self._last_sample = 0
+        self._sample_interval = 0.1  # seconds
+        self._warn_interval = self._sample_interval * 4
+
+    def process_result(self, obj):
+        if isinstance(obj, RegisterProcessMessage):
+            self._sampler.register_process(
+                obj.pid, obj.task_id, obj.server_name, obj.worker_id,
+                obj.worker_name)
+        self._wakeup()
+
+    def process_timeout(self, delta_seconds):
+        self._wakeup()
+
+    def _wakeup(self):
+        """Invoke Sampler.sample() if enough time elapsed since
+           the previous call.
+        """
+        now = time.time()
+        delta = now - self._last_sample
+        if self._last_sample > 0 and delta > self._warn_interval:
+            template = 'Low sampling resolution. The expected interval\n' + \
+                       'is {:.2f} seconds ({:.2f} seconds without warnings),\n' + \
+                       'but the last sample was collected {:.2f} seconds ago.'
+            qa_notice(template.format(self._sample_interval, self._warn_interval,
+                                      delta))
+        if delta > self._sample_interval:
+            self._sampler._sample()
+            self._last_sample = now
+
+
+class Sampler:
+    def __init__(self):
+        # The instance is created in the test-run main process.
+
+        # Field for an instance in a worker.
+        self._worker_id = None
+        self._worker_name = None
+        self._queue = None
+
+        # Field for an instance in the main process.
+        self._watcher = SamplerWatcher(self)
+
+        self._processes = dict()
+
+    def set_queue(self, queue, worker_id, worker_name):
+        # Called from a worker process (_run_worker()).
+        self._worker_id = worker_id
+        self._worker_name = worker_name
+        self._queue = queue
+        self._watcher = None
+
+    @property
+    def watcher(self):
+        if not self._watcher:
+            raise RuntimeError('sampler: watcher is available only in the ' +
+                               'main test-run process')
+        return self._watcher
+
+    def register_process(self, pid, task_id, server_name, worker_id=None,
+                         worker_name=None):
+        """Register a process to sampling.
+
+           Call it without worker_* arguments from a worker
+           process.
+        """
+        if not self._queue:
+            # In main test-run process.
+            self._processes[pid] = {
+                'task_id': task_id,
+                'server_name': server_name,
+                'worker_id': worker_id,
+                'worker_name': worker_name,
+            }
+            self._log('register', pid)
+            return
+
+        # Pass to the main test-run process.
+        self._queue.put(RegisterProcessMessage(
+            self._worker_id, self._worker_name, pid, task_id, server_name))
+
+    def unregister_process(self, pid):
+        if self._queue:
+            raise NotImplementedError('sampler: a process unregistration ' +
+                                      'from a test-run worker is not ' +
+                                      'implemented yet')
+        if pid not in self._processes:
+            return
+
+        self._log('unregister', pid)
+        del self._processes[pid]
+
+    def _log(self, event, pid):
+        # Those logs are not written due to gh-247.
+        process_def = self._processes[pid]
+        task_id = process_def['task_id']
+        test_name = task_id[0] + ((':' + task_id[1]) if task_id[1] else '')
+        worker_name = process_def['worker_name']
+        server_name = process_def['server_name']
+        color_log('DEBUG: sampler: {} {}\n'.format(
+                  event, format_process(pid)), schema='info')
+        color_log(' | worker: {}\n'.format(worker_name))
+        color_log(' | test: {}\n'.format(test_name))
+        color_log(' | server: {}\n'.format(str(server_name)))
+
+    def _sample(self):
+        for pid in list(self._processes.keys()):
+            # Unregister processes that're gone.
+            # Assume that PIDs are rarely reused.
+            try:
+                os.kill(pid, 0)
+            except ProcessLookupError:
+                self.unregister_process(pid)
+            else:
+                self._sample_process(pid)
+
+    def _sample_process(self, pid):
+        # Your sampling code here.
+        pass
+
+
+# The 'singleton' sampler instance: created in the main test-run
+# process, but then work differently in the main process and
+# workers.
+sampler = Sampler()
--- a/lib/sampler.py
+++ b/lib/sampler.py
@@ -5,6 +5,8 @@ import time
 from lib.colorer import color_log
 from lib.colorer import qa_notice
 from lib.utils import format_process
+from lib.utils import get_proc_stat_rss
+from lib.utils import proc_stat_rss_supported
 
 
 if sys.version_info[0] == 2:
@@ -42,6 +44,10 @@ class SamplerWatcher(object):
     def process_timeout(self, delta_seconds):
         self._wakeup()
 
+    @property
+    def sample_interval(self):
+        return self._sample_interval
+
     def _wakeup(self):
         """Invoke Sampler.sample() if enough time elapsed since
            the previous call.
@@ -72,6 +78,7 @@ class Sampler:
         self._watcher = SamplerWatcher(self)
 
         self._processes = dict()
+        self._rss_summary = dict()
 
     def set_queue(self, queue, worker_id, worker_name):
         # Called from a worker process (_run_worker()).
@@ -81,11 +88,24 @@ class Sampler:
         self._watcher = None
 
     @property
+    def rss_summary(self):
+        """Task ID to maximum RSS mapping."""
+        return self._rss_summary
+
+    @property
+    def sample_interval(self):
+        return self._watcher.sample_interval
+
+    @property
     def watcher(self):
         if not self._watcher:
             raise RuntimeError('sampler: watcher is available only in the ' +
                                'main test-run process')
         return self._watcher
+
+    @property
+    def is_enabled(self):
+        return proc_stat_rss_supported()
 
     def register_process(self, pid, task_id, server_name, worker_id=None,
                          worker_name=None):
@@ -134,6 +154,7 @@ class Sampler:
         color_log(' | server: {}\n'.format(str(server_name)))
 
     def _sample(self):
+        tasks_rss = dict()
         for pid in list(self._processes.keys()):
             # Unregister processes that're gone.
             # Assume that PIDs are rarely reused.
@@ -142,11 +163,17 @@ class Sampler:
             except ProcessLookupError:
                 self.unregister_process(pid)
             else:
-                self._sample_process(pid)
+                self._sample_process(pid, tasks_rss)
 
-    def _sample_process(self, pid):
-        # Your sampling code here.
-        pass
+        # Save current overall RSS value if it is bigger than saved.
+        for task_id in tasks_rss:
+            if self.rss_summary.get(task_id, 0) < tasks_rss[task_id]:
+                self.rss_summary[task_id] = tasks_rss[task_id]
+
+    def _sample_process(self, pid, tasks_rss):
+        task_id = self._processes[pid]['task_id']
+        # Count overall RSS per task.
+        tasks_rss[task_id] = get_proc_stat_rss(pid) + tasks_rss.get(task_id, 0)
 
 
 # The 'singleton' sampler instance: created in the main test-run
--- a/lib/test_suite.py
+++ b/lib/test_suite.py
@@ -289,3 +289,6 @@ class TestSuite:
 
     def show_reproduce_content(self):
         return self.ini['show_reproduce_content']
+
+    def test_is_long(self, task_id):
+        return os.path.basename(task_id[0]) in self.ini['long_run']
--- a/lib/test_suite.py
+++ b/lib/test_suite.py
@@ -9,6 +9,7 @@ import json
 import os
 import re
 import sys
+import time
 
 from lib import Options
 from lib.app_server import AppServer
@@ -268,18 +269,20 @@ class TestSuite:
             conf = test.conf_name
         color_stdout(just_and_trim(conf, 15) + ' ', schema='test_var')
 
+        start_time = time.time()
         if self.is_test_enabled(test, conf, server):
             short_status, result_checksum = test.run(server)
         else:
             color_stdout("[ disabled ]\n", schema='t_name')
             short_status = 'disabled'
             result_checksum = None
+        duration = time.time() - start_time
 
         # cleanup only if test passed or if --force mode enabled
         if Options().args.is_force or short_status == 'pass':
             inspector.cleanup_nondefault()
 
-        return short_status, result_checksum
+        return short_status, result_checksum, duration
 
     def is_parallel(self):
         return self.ini['is_parallel']
--- a/lib/worker.py
+++ b/lib/worker.py
@@ -146,15 +146,17 @@ class WorkerTaskResult(BaseWorkerMessage):
     than binary True/False. The result_checksum (string) field saves the results
     file checksum on test fail. The task_id (any hashable object) field hold ID of
     the processed task. The is_long (boolean) field shows if task is in long test
-    list in suite.ini. The show_reproduce_content configuration from suite.ini.
+    list in suite.ini. The duration (float) field saves the task run time. The
+    show_reproduce_content configuration from suite.ini.
     """
     def __init__(self, worker_id, worker_name, task_id,
-                 short_status, result_checksum, is_long, show_reproduce_content):
+                 short_status, result_checksum, is_long, duration, show_reproduce_content):
         super(WorkerTaskResult, self).__init__(worker_id, worker_name)
         self.short_status = short_status
         self.result_checksum = result_checksum
         self.task_id = task_id
         self.is_long = is_long
+        self.duration = duration
         self.show_reproduce_content = show_reproduce_content
 
 
@@ -220,10 +222,10 @@ class Worker:
         return WorkerCurrentTask(self.id, self.name, task_name, task_param,
                                  task_result, task_tmp_result)
 
-    def wrap_result(self, task_id, short_status, result_checksum):
+    def wrap_result(self, task_id, short_status, result_checksum, duration):
         return WorkerTaskResult(self.id, self.name, task_id, short_status,
                                 result_checksum,
-                                self.suite.test_is_long(task_id),
+                                self.suite.test_is_long(task_id), duration,
                                 self.suite.show_reproduce_content())
 
     def sigterm_handler(self, signum, frame):
@@ -315,7 +317,7 @@ class Worker:
             with open(self.reproduce_file, 'a') as f:
                 task_id_str = yaml.safe_dump(task.id, default_flow_style=True)
                 f.write('- ' + task_id_str)
-            short_status, result_checksum = self.suite.run_test(
+            short_status, result_checksum, duration = self.suite.run_test(
                 task, self.server, self.inspector)
         except KeyboardInterrupt:
             self.report_keyboard_interrupt()
@@ -325,7 +327,7 @@ class Worker:
                 '\nWorker "%s" received the following error; stopping...\n'
                 % self.name + traceback.format_exc() + '\n', schema='error')
             raise
-        return short_status, result_checksum
+        return short_status, result_checksum, duration
 
     def run_loop(self, task_queue, result_queue):
         """ called from 'run_all' """
@@ -341,6 +343,7 @@ class Worker:
 
             short_status = None
             result_checksum = None
+            duration = 0.0
             result_queue.put(self.current_task(task_id))
             testname = os.path.basename(task_id[0])
             fragile_checksums = self.suite.get_test_fragile_checksums(testname)
@@ -356,14 +359,14 @@ class Worker:
                         ' file checksum: "%s", rerunning with server restart ...\n'
                         % (task_id[0], task_id[1], result_checksum), schema='error')
                 # run task and save the result to short_status
-                short_status, result_checksum = self.run_task(task_id)
+                short_status, result_checksum, duration = self.run_task(task_id)
                 # check if the results file checksum set on fail and if
                 # the newly created results file is known by checksum
                 if not result_checksum or (result_checksum not in fragile_checksums):
                     break
                 retries_left = retries_left - 1
 
-            result_queue.put(self.wrap_result(task_id, short_status, result_checksum))
+            result_queue.put(self.wrap_result(task_id, short_status, result_checksum, duration))
             if short_status == 'fail':
                 if Options().args.is_force:
                     self.restart_server()
--- a/lib/worker.py
+++ b/lib/worker.py
@@ -145,14 +145,16 @@ class WorkerTaskResult(BaseWorkerMessage):
     the task processed successfully or not, but with little more flexibility
     than binary True/False. The result_checksum (string) field saves the results
     file checksum on test fail. The task_id (any hashable object) field hold ID of
-    the processed task. The show_reproduce_content configuration form suite.ini
+    the processed task. The is_long (boolean) field shows if task is in long test
+    list in suite.ini. The show_reproduce_content configuration from suite.ini.
     """
     def __init__(self, worker_id, worker_name, task_id,
-                 short_status, result_checksum, show_reproduce_content):
+                 short_status, result_checksum, is_long, show_reproduce_content):
         super(WorkerTaskResult, self).__init__(worker_id, worker_name)
         self.short_status = short_status
         self.result_checksum = result_checksum
         self.task_id = task_id
+        self.is_long = is_long
         self.show_reproduce_content = show_reproduce_content
 
 
@@ -221,6 +223,7 @@ class Worker:
     def wrap_result(self, task_id, short_status, result_checksum):
         return WorkerTaskResult(self.id, self.name, task_id, short_status,
                                 result_checksum,
+                                self.suite.test_is_long(task_id),
                                 self.suite.show_reproduce_content())
 
     def sigterm_handler(self, signum, frame):
--- a/listeners.py
+++ b/listeners.py
@@ -36,6 +36,7 @@ class StatisticsWatcher(BaseWatcher):
         self.stats = dict()
         self.field_size = 60
         self._sampler = sampler
+        self.duration_stats = dict()
         self.failed_tasks = []
         self.get_logfile = get_logfile
         self.long_tasks = set()
@@ -56,6 +57,8 @@ class StatisticsWatcher(BaseWatcher):
                                       obj.worker_name,
                                       obj.result_checksum,
                                       obj.show_reproduce_content))
+
+        self.duration_stats[obj.task_id] = obj.duration
 
     def get_long_mark(self, task):
         return '(long)' if task in self.long_tasks else ''
@@ -106,6 +109,42 @@ class StatisticsWatcher(BaseWatcher):
                                       rss_summary[task_id]))
         fd.close()
 
+    # Durations.
+    def print_duration(self, stats_dir):
+        top_durations = 10
+
+        # Print to stdout durations for all failed tasks.
+        if self.failed_tasks:
+            color_stdout('Duration of failed tests (seconds):\n',
+                         schema='info')
+            for task in self.failed_tasks:
+                task_id = task[0]
+                if task_id in self.duration_stats:
+                    color_stdout('* %6.2f %s %s\n' % (self.duration_stats[task_id],
+                                 self.prettify_task_name(task_id).ljust(self.field_size),
+                                 self.get_long_mark(task_id)),
+                                 schema='info')
+            color_stdout('\n')
+
+        # Print to stdout durations for some number of most long tasks.
+        color_stdout('Top {} longest tests (seconds):\n'.format(top_durations),
+                     schema='info')
+        results_sorted = sorted(self.duration_stats.items(), key=lambda x: x[1], reverse=True)
+        for task_id, duration in results_sorted[:top_durations]:
+            color_stdout('* %6.2f %s %s\n' % (duration,
+                         self.prettify_task_name(task_id).ljust(self.field_size),
+                         self.get_long_mark(task_id)), schema='info')
+
+        color_stdout('-' * 81, "\n", schema='separator')
+
+        # Print duration statistics to '<vardir>/statistics/duration.log' file.
+        filepath = os.path.join(stats_dir, 'duration.log')
+        fd = open(filepath, 'w')
+        for task_id in self.duration_stats:
+            fd.write("{} {}\n".format(self.prettify_task_name(task_id),
+                                      self.duration_stats[task_id]))
+        fd.close()
+
     def print_statistics(self):
         """Print statistics and results of testing."""
         # Prepare standalone subpath '<vardir>/statistics' for statistics files.
@@ -113,6 +152,7 @@ class StatisticsWatcher(BaseWatcher):
         safe_makedirs(stats_dir)
 
         self.print_rss_summary(stats_dir)
+        self.print_duration(stats_dir)
 
         if self.stats:
             color_stdout('Statistics:\n', schema='test_var')
--- a/listeners.py
+++ b/listeners.py
@@ -6,6 +6,7 @@ import shutil
 from lib import Options
 from lib.colorer import color_stdout
 from lib.colorer import decolor
+from lib.sampler import sampler
 from lib.worker import WorkerCurrentTask
 from lib.worker import WorkerDone
 from lib.worker import WorkerOutput
@@ -33,12 +34,18 @@ class BaseWatcher(object):
 class StatisticsWatcher(BaseWatcher):
     def __init__(self, get_logfile):
         self.stats = dict()
+        self.field_size = 60
+        self._sampler = sampler
         self.failed_tasks = []
         self.get_logfile = get_logfile
+        self.long_tasks = set()
 
     def process_result(self, obj):
         if not isinstance(obj, WorkerTaskResult):
             return
+
+        if obj.is_long:
+            self.long_tasks.add(obj.task_id)
 
         if obj.short_status not in self.stats:
             self.stats[obj.short_status] = 0
@@ -50,8 +57,63 @@ class StatisticsWatcher(BaseWatcher):
                                       obj.result_checksum,
                                       obj.show_reproduce_content))
 
+    def get_long_mark(self, task):
+        return '(long)' if task in self.long_tasks else ''
+
+    def prettify_task_name(self, task_id):
+        return task_id[0] + ((':' + task_id[1]) if task_id[1] else '')
+
+    # RSS.
+    def print_rss_summary(self, stats_dir):
+        if not self._sampler.is_enabled:
+            return
+
+        rss_summary = self._sampler.rss_summary
+        top_rss = 10
+
+        # Print to stdout RSS statistics for all failed tasks.
+        if self.failed_tasks:
+            color_stdout('Occupied memory in failed tests (RSS, Mb):\n', schema='info')
+            for task in self.failed_tasks:
+                task_id = task[0]
+                if task_id in rss_summary:
+                    color_stdout('* %6.1f %s %s\n' % (float(rss_summary[task_id]) / 1024,
+                                 self.prettify_task_name(task_id).ljust(self.field_size),
+                                 self.get_long_mark(task_id)),
+                                 schema='info')
+            color_stdout('\n')
+
+        # Print to stdout RSS statistics for some number of most it used tasks.
+        color_stdout('Top {} tests by occupied memory (RSS, Mb):\n'.format(
+                     top_rss), schema='info')
+        results_sorted = sorted(rss_summary.items(), key=lambda x: x[1], reverse=True)
+        for task_id, rss in results_sorted[:top_rss]:
+            color_stdout('* %6.1f %s %s\n' % (float(rss) / 1024,
+                         self.prettify_task_name(task_id).ljust(self.field_size),
+                         self.get_long_mark(task_id)), schema='info')
+        color_stdout('\n')
+
+        color_stdout('(Tests quicker than {} seconds may be missed.)\n'.format(
+                     self._sampler.sample_interval), schema='info')
+
+        color_stdout('-' * 81, "\n", schema='separator')
+
+        # Print RSS statistics to '<vardir>/statistics/rss.log' file.
+        filepath = os.path.join(stats_dir, 'rss.log')
+        fd = open(filepath, 'w')
+        for task_id in rss_summary:
+            fd.write("{} {}\n".format(self.prettify_task_name(task_id),
+                                      rss_summary[task_id]))
+        fd.close()
+
     def print_statistics(self):
-        """Returns are there failed tasks."""
+        """Print statistics and results of testing."""
+        # Prepare standalone subpath '<vardir>/statistics' for statistics files.
+        stats_dir = os.path.join(Options().args.vardir, 'statistics')
+        safe_makedirs(stats_dir)
+
+        self.print_rss_summary(stats_dir)
+
         if self.stats:
             color_stdout('Statistics:\n', schema='test_var')
         for short_status, cnt in self.stats.items():
@@ -121,6 +183,8 @@ class ArtifactsWatcher(BaseWatcher):
                             ignore=shutil.ignore_patterns(
                                 '*.socket-iproto', '*.socket-admin',
                                 '*.sock', '*.control'))
+        shutil.copytree(os.path.join(vardir, 'statistics'),
+                        os.path.join(artifacts_dir, 'statistics'))
 
 
 class LogOutputWatcher(BaseWatcher):
--- a/lib/app_server.py
+++ b/lib/app_server.py
@@ -13,6 +13,7 @@ from lib.colorer import color_log
 from lib.colorer import qa_notice
 from lib.options import Options
 from lib.preprocessor import TestState
+from lib.sampler import sampler
 from lib.server import Server
 from lib.server import DEFAULT_SNAPSHOT_NAME
 from lib.tarantool_server import Test
@@ -31,9 +32,10 @@ def timeout_handler(server_process, test_timeout):
     server_process.kill()
 
 
-def run_server(execs, cwd, server, logfile, retval):
+def run_server(execs, cwd, server, logfile, retval, test_id):
     os.putenv("LISTEN", server.iproto)
     server.process = Popen(execs, stdout=PIPE, stderr=PIPE, cwd=cwd)
+    sampler.register_process(server.process.pid, test_id, server.name)
     test_timeout = Options().args.test_timeout
     timer = Timer(test_timeout, timeout_handler, (server.process, test_timeout))
     timer.start()
@@ -57,7 +59,7 @@ class AppTest(Test):
         execs = server.prepare_args()
         retval = dict()
         tarantool = TestRunGreenlet(run_server, execs, server.vardir, server,
-                                    server.logfile, retval)
+                                    server.logfile, retval, self.id)
         self.current_test_greenlet = tarantool
 
         # Copy the snapshot right before starting the server.
--- a/lib/tarantool_server.py
+++ b/lib/tarantool_server.py
@@ -32,6 +32,7 @@ from lib.colorer import color_log
 from lib.colorer import qa_notice
 from lib.options import Options
 from lib.preprocessor import TestState
+from lib.sampler import sampler
 from lib.server import Server
 from lib.server import DEFAULT_SNAPSHOT_NAME
 from lib.test import Test
@@ -332,6 +333,10 @@ class LuaTest(Test):
 
     def execute(self, server):
         super(LuaTest, self).execute(server)
+
+        # Track the same process metrics as part of another test.
+        sampler.register_process(server.process.pid, self.id, server.name)
+
         cls_name = server.__class__.__name__.lower()
         if 'gdb' in cls_name or 'lldb' in cls_name or 'strace' in cls_name:
             # don't propagate gdb/lldb/strace mixin to non-default servers,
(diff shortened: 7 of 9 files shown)


listeners.py
```diff
@@ -178,6 +179,86 @@ def __del__(self):
...
+            return name
+        return name + ':' + conf
+
+    def process_result(self, obj):
+        # called only once on task run initiated
+        if isinstance(obj, WorkerCurrentTask):
+            task_name = self.get_task_name_conf(obj.task_name, obj.task_param)
+            self.worker_current_task[obj.worker_id] = obj
+            self.rss_results[task_name] = get_proc_stat_rss(
+                    self.worker_id_to_pid[obj.worker_id])
```
Totktonada, line 205: It is Python process, not tarantool. Don't forget about non-default servers as well.

listeners.py
```diff
@@ -178,6 +179,86 @@ def __del__(self):
                 pass
 
 
+class StatisticsMonitor(BaseWatcher):
```
Totktonada, line 183: For the sake of terminology consistency: all other listeners/monitors/watchers are called like FooWatcher. I would follow it here.

listeners.py
```diff
@@ -178,6 +179,86 @@ def __del__(self):
...
+            if count < top_rss:
+                color_stdout('* %6.1f %s\n' % (int(rss) / 1000, task_name), schema='test_var')
+            count = count + 1
+        color_stdout('-' * 41, "\n", schema='separator')
+
+        # print RSS statistics to '<vardir>/statistics/rss.log' file
+        filepath = os.path.join(stats_dir, 'rss.log')
+        fd = open(filepath, 'w')
+        for task_name in self.rss_results:
+            fd.write("{} {}\n" . format(task_name, self.rss_results[task_name]))
```
Totktonada, line 292: Please, don't surround a period in a method call with whitespaces.

listeners.py
```diff
@@ -178,6 +179,86 @@ def __del__(self):
...
+
+        # print to stdout RSS statistics for some number of most it used tasks
+        color_stdout('Up to {} most RSS used tasks in Mb:\n' . format(top_rss), schema='test_var')
+        results_sorted = sorted(self.rss_results.items(), key=lambda x: x[1], reverse=True)
+        count = 0
+        for task_name, rss in results_sorted:
+            if count < top_rss:
+                color_stdout('* %6.1f %s\n' % (int(rss) / 1000, task_name), schema='test_var')
+            count = count + 1
+        color_stdout('-' * 41, "\n", schema='separator')
```
Totktonada, line 286: You can loop over `results_sorted[:top_rss]` instead of the `if` block and the counter.

listeners.py
```diff
@@ -178,6 +179,86 @@ def __del__(self):
...
+                    self.worker_id_to_pid[obj.worker_id])
+
+        # called only once on task run finished
+        if isinstance(obj, WorkerTaskResult):
+            task_name = self.get_task_name_conf(obj.task_id[0], obj.task_id[1])
+            if obj.short_status == 'fail':
+                self.failed_tasks[task_name] = 1
+
+    # called on each 'delta_seconds' delay within task run
+    def process_timeout(self, delta_seconds):
```
Totktonada, line 216: It is not guaranteed to be called once per 0.1 seconds. We can receive results from the results queue quite often (especially due to #280) and never execute `process_timeout()`. So we should collect metrics in `process_result()` too, but ratelimit it: save when we collect them last time and collect again only after at least 0.1 seconds.

dispatcher.py
```diff
@@ -43,6 +43,7 @@
 from listeners import LogOutputWatcher
 from listeners import OutputWatcher
 from listeners import StatisticsWatcher
+from listeners import StatisticsMonitor
```
Totktonada, line 46: AFAIU we resolve #277 here, which is subtask of https://github.com/tarantool/tarantool-qa/issues/98.

listeners.py
```diff
@@ -197,13 +199,16 @@ def process_result(self, obj):
...
+            self.tasks_timings[task_name] = time.time()
             self.worker_current_task[obj.worker_id] = obj
             self.rss_results[task_name] = get_proc_stat_rss(
                     self.worker_id_to_pid[obj.worker_id])
 
         # called only once on task run finished
         if isinstance(obj, WorkerTaskResult):
             task_name = self.get_task_name_conf(obj.task_id[0], obj.task_id[1])
+            self.tasks_timings[task_name] = round(time.time() -
+                                                  self.tasks_timings[task_name], 2)
```
Totktonada, line 211: I would add test timing into `WorkerTaskResult`, because of several reasons:

1. Accuracy. We process many different results here and the moment of this `time.time()` is not strictly after the end of a test.
2. Don't sum timings re-runs of flaky tests.
3. Reliability. If we'll add some heavy work into the main process (intentionally or not), timings will be affected if we'll collect them from the main process. While tracking from the worker process will achieve the best possible precision.

The worst thing here is that even if everything works nice during testing of the patch, we can lose accuracy later in a specific environment or after some test-run changes and **we will not know about this**. That's why I insist on the approach that will ensure best possible accuracy. It is not a thing that either work or report an error.

Totktonada:
@avtikhon asked me a suggestion how to better track tarantool processes in test-run, so I created a simple sampling infrastructure for test-run as example: 6e73ac470322c458eed08f58f74fad1fdf8c9d99.

Not tested thoroughly, but at least I looked over logs for different types of tests and it seems that it tracks everything that I would expect.

Sadly, the register-unregister logs are not written to `test/var/log/` due to #247, but we can print them to the terminal for debugging purposes this way:

```diff
diff --git a/lib/sampler.py b/lib/sampler.py
index 8bcc6ca..95dcec0 100644
--- a/lib/sampler.py
+++ b/lib/sampler.py
@@ -2,7 +2,7 @@ import os
 import sys
 import time
 
-from lib.colorer import color_log
+from lib.colorer import color_stdout as color_log
 from lib.colorer import qa_notice
 from lib.utils import format_process
```

listeners.py
avtikhon: Right, corrected.

listeners.py
avtikhon: Right, changed to "ResourcesWatcher".

listeners.py
avtikhon: Sure, moved get finish time to the "WorkerTaskResult", also start time to the "WorkerCurrentTask".

Totktonada:
> Using 'WorkerCurrentTask' queue it saves the initial RSS value of
> the used worker when task started to run.

It is the result queue message, not the queue name.

listeners.py
```diff
@@ -178,6 +178,55 @@ def __del__(self):
...
+            for task_name in self.failed_tasks:
+                if task_name in self.rss_results:
+                    color_stdout('* %6.1f %s\n' % (self.rss_results[task_name] / 1000, task_name),
+                                 schema='test_var')
+
+        # print to stdout RSS statistics for some number of most it used tasks
+        color_stdout('Up to {} most RSS used tasks in Mb:\n'.format(top_rss), schema='test_var')
+        results_sorted = sorted(self.rss_results.items(), key=lambda x: x[1], reverse=True)
+        for task_name, rss in results_sorted[:top_rss]:
+            color_stdout('* %6.1f %s\n' % (int(rss) / 1000, task_name), schema='test_var')
```
Totktonada, line 247: 1. 1000 → 1024.
2. In Python 2 it is integer division.

avtikhon:
> It is the result queue message, not the queue name.

Right, corrected all the same places in commit messages.

listeners.py
avtikhon: Sure, with:
```
diff --git a/listeners.py b/listeners.py
index 3ab8fd4..5e5db70 100644
--- a/listeners.py
+++ b/listeners.py
@@ -237,14 +237,14 @@ class ResourcesWatcher(BaseWatcher):
             color_stdout('RSS of the failed tasks:\n', schema='test_var')
             for task_name in self.failed_tasks:
                 if task_name in self.rss_results:
-                    color_stdout('* %6.1f %s\n' % (self.rss_results[task_name] / 1000, task_name),
-                                 schema='test_var')
+                    color_stdout('* %6.1f %s\n' % (float(self.rss_results[task_name] / 1024),
+                                 task_name), schema='test_var')
 
         # print to stdout RSS statistics for some number of most it used tasks
         color_stdout('Up to {} most RSS used tasks in Mb:\n'.format(top_rss), schema='test_var')
         results_sorted = sorted(self.rss_results.items(), key=lambda x: x[1], reverse=True)
         for task_name, rss in results_sorted[:top_rss]:
-            color_stdout('* %6.1f %s\n' % (int(rss) / 1000, task_name), schema='test_var')
+            color_stdout('* %6.1f %s\n' % (float(int(rss) / 1024), task_name), schema='test_var')
         color_stdout('-' * 41, "\n", schema='separator')
 
         # print RSS statistics to '<vardir>/statistics/rss.log' file
```

Totktonada: Please, don't write code in a hurry. I'm sure you able to find mistakes if you'll look again at the code:

1. `float(self.rss_results[task_name] / 1024)`
2. `float(int(rss) / 1024)`

Totktonada:
## 2/3 Add RSS statistics collecting

I'll describe my complains briefly to unblock you. Sorry for the non-detailed comments.

First of all, the commit message does not correspond the implementation of ResourcesWatcher: at least it does not look at WorkerCurrentTask.

Next, I don't like how the code is organized. Sorry, but it either not logical or can be organized better:

1. ResourceWatcher does not watch for resources, just tracks failed tests, which are already tracked in StatisticsWatcher. In the next commit (3/3 'Add tests timings collecting') it also tracks tests duration, but it is not about resource consumption too: it seems logical to track it in the StatisticsWatcher too.

If we'll eliminate the ResourceWatcher, the collected resource consumption statistics will be accessible anyway: we can hold it in the sampler. The test duration statistics may be hold in the StatisticsWatcher.

2. Some sampling code and fields are in SamplerWatcher. It should not be so. The responsibility area of SamplerWatcher is to integrate Sampler into our event loop (which spins around [select()](https://github.com/tarantool/test-run/blob/e698240f6a2034138c102e5ce627a46343b6e5ad/dispatcher.py#L301-L302)): all those process_result() / process_timeout(), tracking of a last sampler wakeup time. Everything else should be performed by sampler.

3. There is the `_sample_process()` method, which is appropriate place to read `/proc/<pid>/status`. It is called for each process from `_sample()`, which is appropriate place to accumulate, transform and save metrics. The result of `_sample()` call is update of a sampler field with accumulated statistics in a form useful for reporting (say, `self._rss_summary` with per-test maximums). There is no need to track alive test list with corresponsing processes, all information is already available from `self.process`.

It is undesirable to depend on test-run internals in sampler.py, so I guess some external procedure (`StatisticsWatcher.print_statistics()`?) should call color_stdout() and write to appropriate log file. Sampler should just provide all information in a convenient form (say, `Sampler.rss_summary()`).

## 3/3 Add tests timings collecting 

> self.started_time = 0.0

Not sure it is good idea. If it is possible to reference this value, so the duration will be ~1.6 * 10^9 seconds.

How it works with reruns? Please, clarify.

I would use the term 'duration' rather than 'timing'.

> timing = round(time.time() - self.started_time, 2)

I would round it at printing, not when it is collected.

## Common

AFAIR, we agreed to mark long tests (long_run in suite.ini) in the statistics output.

Totktonada:
> 2. Some sampling code and fields are in SamplerWatcher. It should not be so. The responsibility area of SamplerWatcher is to integrate Sampler into our event loop (which spins around [select()](https://github.com/tarantool/test-run/blob/e698240f6a2034138c102e5ce627a46343b6e5ad/dispatcher.py#L301-L302)): all those process_result() / process_timeout(), tracking of a last sampler wakeup time. Everything else should be performed by sampler.

My bad, I wrongly read the diff and had thought that `collect_rss()` is in SimplerWatcher. Sorry.

However there is (unused?) `self.rss_procs_results` new field in SamplerWatcher. And `Sampler.collect_rss()` should be integrated into `_sample_process()` / `_sample()` (see also the 3rd point above).

avtikhon:
> ## 2/3 Add RSS statistics collecting
> I'll describe my complains briefly to unblock you. Sorry for the non-detailed comments.
> 
> First of all, the commit message does not correspond the implementation of ResourcesWatcher: at least it does not look at WorkerCurrentTask.
> 
> Next, I don't like how the code is organized. Sorry, but it either not logical or can be organized better:
> 
> 1. ResourceWatcher does not watch for resources, just tracks failed tests, which are already tracked in StatisticsWatcher. In the next commit (3/3 'Add tests timings collecting') it also tracks tests duration, but it is not about resource consumption too: it seems logical to track it in the StatisticsWatcher too.
> 
> If we'll eliminate the ResourceWatcher, the collected resource consumption statistics will be accessible anyway: we can hold it in the sampler. The test duration statistics may be hold in the StatisticsWatcher.
> 
> 1. Some sampling code and fields are in SamplerWatcher. It should not be so. The responsibility area of SamplerWatcher is to integrate Sampler into our event loop (which spins around [select()](https://github.com/tarantool/test-run/blob/e698240f6a2034138c102e5ce627a46343b6e5ad/dispatcher.py#L301-L302)): all those process_result() / process_timeout(), tracking of a last sampler wakeup time. Everything else should be performed by sampler.
> 2. There is the `_sample_process()` method, which is appropriate place to read `/proc/<pid>/status`. It is called for each process from `_sample()`, which is appropriate place to accumulate, transform and save metrics. The result of `_sample()` call is update of a sampler field with accumulated statistics in a form useful for reporting (say, `self._rss_summary` with per-test maximums). There is no need to track alive test list with corresponsing processes, all information is already available from `self.process`.
> 
> It is undesirable to depend on test-run internals in sampler.py, so I guess some external procedure (`StatisticsWatcher.print_statistics()`?) should call color_stdout() and write to appropriate log file. Sampler should just provide all information in a convenient form (say, `Sampler.rss_summary()`).
> 
> ## 3/3 Add tests timings collecting
> > self.started_time = 0.0
> 
> Not sure it is good idea. If it is possible to reference this value, so the duration will be ~1.6 * 10^9 seconds.
> 
> How it works with reruns? Please, clarify.
> 
> I would use the term 'duration' rather than 'timing'.
> 
> > timing = round(time.time() - self.started_time, 2)
> 
> I would round it at printing, not when it is collected.
> 
> ## Common
> AFAIR, we agreed to mark long tests (long_run in suite.ini) in the statistics output.

Alexander, thanks a lot for a deep explanation and offline help. I've made all of your suggestions. Please check the new updated version of the patchset.

listeners.py
avtikhon: Right, corrected.

dispatcher.py
```diff
@@ -122,7 +122,7 @@ def __init__(self, task_groups, max_workers_cnt, randomize):
             self.result_queues.append(task_queue_disp.result_queue)
             self.task_queues.append(task_queue_disp.task_queue)
 
-        self.report_timeout = 1.0
+        self.report_timeout = 0.1
```
Totktonada, line 125: It was in the 'sampler: add simple sampling infrastructure' commit and it would be good to leave it here. Otherwise the following statement in the commit message will lie:

> Add a function that will be called each 0.1 second for each tracked
process.

lib/sampler.py
```diff
@@ -0,0 +1,155 @@
...
+
+        # Field for an instance in a worker.
+        self._worker_id = None
+        self._worker_name = None
+        self._queue = None
+
+        # Field for an instance in the main process.
+        self._watcher = SamplerWatcher(self)
+
+        self.processes = dict()
```
Totktonada, line 75: I don't know why I don't underscored it. It is the private field. Can you fix it, please?

lib/sampler.py
```diff
@@ -80,6 +82,9 @@ def set_queue(self, queue, worker_id, worker_name):
         self._queue = queue
         self._watcher = None
 
+    def rss_summary(self):
```
Totktonada, line 85: It would be nice to add a doc comment with description of return format.

lib/sampler.py
```diff
@@ -142,11 +151,18 @@ def _sample(self):
...
+                self._sample_process(pid, tasks_rss)
+        # save current overall RSS value if it is bigger than saved
+        for task in tasks_rss:
+            if self._rss_summary[task] < tasks_rss[task]:
+                self._rss_summary[task] = tasks_rss[task]
+
+    def _sample_process(self, pid, tasks_rss):
+        self.processes[pid]['rss'] = get_proc_stat_rss(pid)
+        task_id = self.processes[pid]['task_id']
+        task = task_id[0] + ((':' + task_id[1]) if task_id[1] else '')
```
Totktonada, line 163: I would just use `task_id` as a key.

lib/sampler.py
```diff
@@ -101,7 +106,10 @@ def register_process(self, pid, task_id, server_name, worker_id=None,
                 'server_name': server_name,
                 'worker_id': worker_id,
                 'worker_name': worker_name,
+                'rss': get_proc_stat_rss(pid),
             }
+            task = task_id[0] + ((':' + task_id[1]) if task_id[1] else '')
+            self._rss_summary[task] = self.processes[pid]['rss']
```
Totktonada: I don't see much value in performing sampling at process registeration. We'll collect RSS at the next sampling point (within 0.1 second) and your measurements (in the commit message) show that it should be often enough to catch a problem. I would remove this code to make it clear how sampling is organized in the code: `_sample_process()` collects per-process metrics, `_sample()` folds them. That's all, ho hacks around.

lib/sampler.py
```diff
@@ -142,11 +151,18 @@ def _sample(self):
...
-        # Your sampling code here.
-        pass
+                self._sample_process(pid, tasks_rss)
+        # save current overall RSS value if it is bigger than saved
+        for task in tasks_rss:
+            if self._rss_summary[task] < tasks_rss[task]:
+                self._rss_summary[task] = tasks_rss[task]
+
+    def _sample_process(self, pid, tasks_rss):
+        self.processes[pid]['rss'] = get_proc_stat_rss(pid)
```
Totktonada, line 161: We don't need to store this value anywhere. We can just add it to `tasks_rss[task_id]`.

lib/sampler.py
```diff
@@ -142,11 +151,18 @@ def _sample(self):
             except ProcessLookupError:
                 self.unregister_process(pid)
             else:
-                self._sample_process(pid)
-
-    def _sample_process(self, pid):
-        # Your sampling code here.
-        pass
+                self._sample_process(pid, tasks_rss)
+        # save current overall RSS value if it is bigger than saved
```
Totktonada, line 155: Nits:

1. Please, add an empty line to separate this block from the previous one.
2. Please, follow tarantool's commenting style, when a comment is started from a capital letter and ends with a period.

lib/worker.py
```diff
@@ -148,12 +148,13 @@ class WorkerTaskResult(BaseWorkerMessage):
...
     """
     def __init__(self, worker_id, worker_name, task_id,
-                 short_status, result_checksum, show_reproduce_content):
+                 short_status, result_checksum, is_long, show_reproduce_content):
         super(WorkerTaskResult, self).__init__(worker_id, worker_name)
         self.short_status = short_status
         self.result_checksum = result_checksum
         self.task_id = task_id
         self.show_reproduce_content = show_reproduce_content
+        self.is_long = is_long
```
Totktonada, line 158: Nit: Please, follow the same order in parameters and those assignments. It eases reading.

listeners.py
```diff
@@ -33,13 +34,19 @@ def process_timeout(self, delta_seconds):
 class StatisticsWatcher(BaseWatcher):
     def __init__(self, get_logfile):
         self.stats = dict()
+        self.rss_results = sampler.rss_summary()
```
Totktonada, line 37: Nit: It is better to avoid usage of different terms for the same thing. `rss_summary` is good as for the method as well as for the variable.

Totktonada, line 37: Here we lean on the fact that the dictionary returned at `StatisticsWatcher` construction will never be replaced by another one. However it would be better to just pass a sampler object here, save it and call `self._sampler.rss_summary()` at the point, where we'll need the statistics (in `print_statistics()`).

listeners.py
```diff
@@ -50,7 +57,48 @@ def process_result(self, obj):
...
+        stats_dir = os.path.join(Options().args.vardir, 'statistics')
+        safe_makedirs(stats_dir)
+
+        # RSS
+
+        # print to stdout RSS statistics for all failed tasks
+        if self.failed_tasks:
+            color_stdout('RSS of the failed tasks:\n', schema='test_var')
+            for task in self.failed_tasks:
+                task_name = self.get_failed_task_name(task)
```
Totktonada, line 85: If you'll pass `task[0]` to the function, it'll be common enough to don't repeat the `task_id[0] + ((':' + task_id[1]) if task_id[1] else '')` expression. I would propose a name like `pretty_task_name()`.

listeners.py
```diff
@@ -50,7 +57,48 @@ def process_result(self, obj):
...
+
+        # prepare standalone subpath '<vardir>/statistics' for statistics files
+        stats_dir = os.path.join(Options().args.vardir, 'statistics')
+        safe_makedirs(stats_dir)
+
+        # RSS
+
+        # print to stdout RSS statistics for all failed tasks
+        if self.failed_tasks:
+            color_stdout('RSS of the failed tasks:\n', schema='test_var')
```
Totktonada, line 83: Nit: `test_var` becomes widely used for yellow text by a mistake, I guess. `info` fits better here.

listeners.py
```diff
@@ -50,7 +57,48 @@ def process_result(self, obj):
...
+            for task in self.failed_tasks:
+                task_name = self.get_failed_task_name(task)
+                if task_name in self.rss_results:
+                    color_stdout('* %6.1f %s %s\n' % (float(self.rss_results[task_name]) / 1024,
```
Totktonada, line 87: Can we show units (like `1.23M`)?

listeners.py
```diff
@@ -50,7 +57,48 @@ def process_result(self, obj):
...
+    def get_failed_task_name(self, task):
+        return task[0][0] + ((':' + task[0][1]) if task[0][1] else '')
+
+    def get_long_mark(self, task):
+        return '(long)' if task in self.long_tasks else ''
+
     def print_statistics(self):
+        field_size = 60
+        top_rss = 10
+        results_sorted = []
```
Totktonada, line 73: `results_sorted = []` is redundant here.

lib/worker.py
```diff
@@ -219,8 +220,9 @@ def current_task(self, task_id):
                                  task_result, task_tmp_result)
 
     def wrap_result(self, task_id, short_status, result_checksum):
+        is_long = self.suite.check_is_long(task_id[0].split('/', 1)[1])
```
Totktonada, line 225: Or `os.path.basename(task_id[0])`, it looks a bit simpler.

listeners.py
```diff
@@ -33,13 +34,19 @@ def process_timeout(self, delta_seconds):
 class StatisticsWatcher(BaseWatcher):
     def __init__(self, get_logfile):
         self.stats = dict()
+        self.rss_results = sampler.rss_summary()
         self.failed_tasks = []
         self.get_logfile = get_logfile
+        self.long_tasks = []
```
Totktonada, line 41: We don't need an order here, so it seems logical to define it as `set()`.

listeners.py
```diff
@@ -33,13 +34,19 @@ def process_timeout(self, delta_seconds):
...
 
     def process_result(self, obj):
         if not isinstance(obj, WorkerTaskResult):
             return
 
+        task_name = obj.task_id[0] + ((':' + obj.task_id[1]) if obj.task_id[1] else '')
+        if obj.is_long:
+            self.long_tasks.append(task_name)
```
Totktonada, line 49: Nit: I would prefer to use `task_id` (a tuple) in internal structures rather than the human readable test name.

lib/worker.py
Totktonada: And I would call it from inside the function, just like in `is_test_enabled()`.

lib/test_suite.py
```diff
@@ -289,3 +289,7 @@ def fragile_retries(self):
 
     def show_reproduce_content(self):
         return self.ini['show_reproduce_content']
+
+    def check_is_long(self, test):
```
Totktonada, line 294: 'Check' is the too common word in most contexts and means nothing. I would name the function like `test_is_long()`.

lib/test_suite.py
```diff
@@ -289,3 +289,7 @@ def fragile_retries(self):
...
+        if Options().args.long:
```
Totktonada, line 295: It is not necessary in fact.

listeners.py
```diff
@@ -50,7 +57,48 @@ def process_result(self, obj):
...
+
+        # print to stdout RSS statistics for some number of most it used tasks
+        color_stdout('Up to {} most RSS used tasks in Mb:\n'.format(top_rss), schema='test_var')
+        results_sorted = sorted(self.rss_results.items(), key=lambda x: x[1], reverse=True)
+        for task_name, rss in results_sorted[:top_rss]:
+            color_stdout('* %6.1f %s %s\n' % (float(rss) / 1024, task_name.ljust(field_size),
+                         self.get_long_mark(task_name)), schema='test_var')
+        color_stdout('-' * 41, "\n", schema='separator')
+
+        # print RSS statistics to '<vardir>/statistics/rss.log' file
```
Totktonada, line 99: We should copy this directory into artifacts in `ArtifactsWatcher` I think.

Totktonada: I mean the whole directory with statistics.

lib/test.py
```diff
@@ -5,6 +5,7 @@
 import re
 import shutil
 import sys
+import time
```
Totktonada, line 8: Re commit message of 'Add tests durations collecting':

> Using 'WorkerTaskResult' message queue

Again, it is not a queue.

listeners.py
```diff
@@ -99,6 +103,34 @@ def print_statistics(self):
...
+                    color_stdout('* %6.2f %s %s\n' % (self.tasks_durations[task_name],
+                                 task_name.ljust(field_size), self.get_long_mark(task_name)),
+                                 schema='test_var')
+
+        # print to stdout durations for some number of most long tasks
+        color_stdout('Up to {} most long tasks in seconds:\n'.format(top_durations),
+                     schema='test_var')
+        results_sorted = sorted(self.tasks_durations.items(), key=lambda x: x[1], reverse=True)
+        for task_name, duration in results_sorted[:top_durations]:
+            color_stdout('* %6.2f %s %s\n' % (float(duration), task_name.ljust(field_size),
```
Totktonada, line 123: The duration is already float.

lib/test_suite.py
```diff
@@ -269,17 +269,18 @@ def run_test(self, test, server, inspector):
         color_stdout(just_and_trim(conf, 15) + ' ', schema='test_var')
 
         if self.is_test_enabled(test, conf, server):
-            short_status, result_checksum = test.run(server)
+            short_status, result_checksum, duration = test.run(server)
```
Totktonada, line 272: I'll cite my past question:

> How it works with reruns? Please, clarify.

listeners.py
```diff
@@ -50,7 +57,48 @@ def process_result(self, obj):
...
     def print_statistics(self):
+        field_size = 60
+        top_rss = 10
+        results_sorted = []
+
+        # prepare standalone subpath '<vardir>/statistics' for statistics files
+        stats_dir = os.path.join(Options().args.vardir, 'statistics')
+        safe_makedirs(stats_dir)
+
+        # RSS
```
Totktonada, line 79: I think we should skip this output (to the terminal and to the file both) when `/proc` is not available at all (on Mac OS). And I would not add the sampler wather at all in the case.

Totktonada:
The patchset is in the good shape, but I would spin for a while around code readability. It is better to keep it good from scratch, rather than fix afterwards. I left my comments above.

lib/sampler.py
avtikhon: Sure, corrected.

lib/sampler.py
avtikhon: Right, initially I thought the same and tried to avoid of its use:
```
diff --git a/lib/sampler.py b/lib/sampler.py
index 11681e4..a47993c 100644
--- a/lib/sampler.py
+++ b/lib/sampler.py
@@ -108,8 +108,6 @@ class Sampler:
                 'worker_name': worker_name,
                 'rss': get_proc_stat_rss(pid),
             }
-            task = task_id[0] + ((':' + task_id[1]) if task_id[1] else '')
-            self._rss_summary[task] = self._processes[pid]['rss']
             self._log('register', pid)
             return
 
@@ -154,7 +152,7 @@ class Sampler:
                 self._sample_process(pid, tasks_rss)
         # save current overall RSS value if it is bigger than saved
         for task in tasks_rss:
-            if self._rss_summary[task] < tasks_rss[task]:
+            if self._rss_summary and task in self._rss_summary and self._rss_summary[task] < tasks_rss[task]:
                 self._rss_summary[task] = tasks_rss[task]
 
     def _sample_process(self, pid, tasks_rss):
```
but met the issue when there is no any RSS result due to test finished faster than 0.1 sec passed:
```
./test-run.py box/huge_field_map.test.lua 
...
[001] box/huge_field_map.test.lua                                     [ pass ]
Up to 10 most RSS used tasks in Mb:
-----------------------------------------
Up to 10 most long tasks in seconds:
*   0.03 box/huge_field_map.test.lua                                  
-----------------------------------------
Statistics:
* pass: 1
```
and only when I returned it back needed result was got:
```
./test-run.py box/huge_field_map.test.lua 
...
[001] box/huge_field_map.test.lua                                     [ pass ]
Up to 10 most RSS used tasks in Mb:
*   67.9 box/huge_field_map.test.lua                                  
-----------------------------------------
Up to 10 most long tasks in seconds:
*   0.03 box/huge_field_map.test.lua                                  
-----------------------------------------
Statistics:
* pass: 1
```

lib/sampler.py
avtikhon: Ok, sure, fixed in 2 places:
```
diff --git a/lib/sampler.py b/lib/sampler.py
index 11681e4..06541a4 100644
--- a/lib/sampler.py
+++ b/lib/sampler.py
@@ -106,10 +106,9 @@ class Sampler:
                 'server_name': server_name,
                 'worker_id': worker_id,
                 'worker_name': worker_name,
-                'rss': get_proc_stat_rss(pid),
             }
             task = task_id[0] + ((':' + task_id[1]) if task_id[1] else '')
-            self._rss_summary[task] = self._processes[pid]['rss']
+            self._rss_summary[task] = get_proc_stat_rss(pid)
             self._log('register', pid)
             return
 
@@ -158,11 +157,10 @@ class Sampler:
                 self._rss_summary[task] = tasks_rss[task]
 
     def _sample_process(self, pid, tasks_rss):
-        self._processes[pid]['rss'] = get_proc_stat_rss(pid)
         task_id = self._processes[pid]['task_id']
         task = task_id[0] + ((':' + task_id[1]) if task_id[1] else '')
         # count overall RSS per task
-        tasks_rss[task] = self._processes[pid]['rss'] + (tasks_rss[task] if task in tasks_rss else 0)
+        tasks_rss[task] = get_proc_stat_rss(pid) + (tasks_rss[task] if task in tasks_rss else 0)
 
 
 # The 'singleton' sampler instance: created in the main test-run
```

lib/sampler.py
avtikhon: Right, corrected all changes in all commits:
~/tarantool/test-run$ git show HEAD~0 | grep \#
```
    Closes #286
         # filled in execute() when a greenlet runs
         # cleanup only if test passed or if --force mode enabled
                 # run task and save the result to short_status
                 # check if the results file checksum set on fail and if
                 # the newly created results file is known by checksum
         # Prepare standalone subpath '<vardir>/statistics' for statistics files.
+        # Durations.
+        # Print to stdout durations for all failed tasks.
+        # Print to stdout durations for some number of most long tasks.
+        # Print durations statistics to '<vardir>/statistics/durations.log' file.
```
~/tarantool/test-run$ git show HEAD~1 | grep \#
```
    Closes #277
         # Called from a worker process (_run_worker()).
             # Unregister processes that're gone.
             # Assume that PIDs are rarely reused.
-        # Your sampling code here.
+        # Save current overall RSS value if it is bigger than saved.
+        # Count overall RSS per task.
 # The 'singleton' sampler instance: created in the main test-run
+        # Prepare standalone subpath '<vardir>/statistics' for statistics files.
+        # RSS.
+        # Print to stdout RSS statistics for all failed tasks.
+        # Print to stdout RSS statistics for some number of most it used tasks.
+        # Print RSS statistics to '<vardir>/statistics/rss.log' file.
```

listeners.py
avtikhon: Sure, removed.

listeners.py
avtikhon: Right, removed.

lib/test_suite.py
avtikhon: Ok, removed.

lib/test_suite.py
avtikhon: Right, that's better, changed.

lib/worker.py
avtikhon: Right, corrected.

listeners.py
avtikhon: Sure, changed rss_results -> rss_summary.

listeners.py
avtikhon: Sure, changed.

listeners.py
avtikhon: Ok, changed.

listeners.py
avtikhon: Ok, changed.

listeners.py
avtikhon: Ok, done.

lib/sampler.py
avtikhon: Right, changed.

lib/sampler.py
avtikhon: After complete offline discussion decided that tests running faster than minimal sample_interval of 0.1 are not interesting in RSS values. Added informational warning to statistics results part:
```
RSS of the failed tasks: (may skip tests run faster than 0.1 sec):
Up to 10 most RSS used tasks in Mb (may skip tests run faster than 0.1 sec):
```

lib/worker.py
avtikhon: Right, changed.

lib/test.py
avtikhon: Right, corrected.

listeners.py
avtikhon: Ok, changed.

lib/sampler.py
avtikhon: Added.

lib/test_suite.py
avtikhon: Running pre-broken box/cfg.test.lua test showed the following result statistics:
```
Durations of the failed tasks:
*   0.50 engine/misc.test.lua:memtx                                   
```
After added checksum to box/suite.ini to enable rerun for 10 times by default, got:
Durations of the failed tasks:
*   0.50 engine/misc.test.lua:memtx                                   
```
This result showed the latest run result from reruns.

File: listeners.py (hunk 20 above)
Comment (avtikhon) [reply listeners.py]: Do we really need it ? We have already printed it in the header of the column:
"Up to 10 most RSS used tasks in Mb (may skip tests run faster than 0.1 sec):"
also we collect results only in Mb and it will be constant prefix, so I don't think we need it.

File: listeners.py (hunk 27 above)
Comment (avtikhon) [reply listeners.py]: Ok, sure done.

File: listeners.py (hunk 20 above)
Comment (Totktonada) [reply listeners.py]: I'd missed it, okay.

File: lib/sampler.py [hunk 32]
```diff
@@ -142,11 +155,19 @@ def _sample(self):
...
-
-    def _sample_process(self, pid):
-        # Your sampling code here.
-        pass
+                self._sample_process(pid, tasks_rss)
+
+        # Save current overall RSS value if it is bigger than saved.
+        for task_id in tasks_rss:
+            if task_id not in self._rss_summary or \
+                    self._rss_summary[task_id] < tasks_rss[task_id]:
```
Totktonada, line 163: BTW, simpler variant: `self._rss_summary.get(task_id, 0) < tasks_rss[task_id]`.

lib/sampler.py
```diff
@@ -142,11 +155,19 @@ def _sample(self):
...
+                self._rss_summary[task_id] = tasks_rss[task_id]
+
+    def _sample_process(self, pid, tasks_rss):
+        task_id = self._processes[pid]['task_id']
+        # Count overall RSS per task.
+        tasks_rss[task_id] = get_proc_stat_rss(pid) + \
+            (tasks_rss[task_id] if task_id in tasks_rss else 0)
```
Totktonada, line 170: And here we can write it shorter too: `<...> + tasks_rss.get(task_id, 0)`.

listeners.py
```diff
@@ -121,6 +178,8 @@ def save_artifacts(self):
                             ignore=shutil.ignore_patterns(
                                 '*.socket-iproto', '*.socket-admin',
                                 '*.sock', '*.control'))
+        shutil.copytree(os.path.join(vardir, 'statistics'),
+                        os.path.join(artifacts_dir, 'statistics'))
```
Totktonada, line 217: NB: I guess you'll skip this directory creation, when there is no /proc (on Mac OS), so we should not forget to skip copying here in the case.

lib/sampler.py
```diff
@@ -80,6 +82,16 @@ def set_queue(self, queue, worker_id, worker_name):
...
 
+    def rss_summary(self):
+        """Return dictionary of collected maximum RSS values per each
+           run task, set with task_id keys.
+        """
+
+        return self._rss_summary
+
+    def sample_interval(self):
+        return self._watcher._sample_interval
```
Totktonada, line 93: The rule of thumb is to don't access external fields, which start from the underscore. I think we can apply the term 'external' either to code outside of the file or outside of the class. I would prefer the latter.

So, I would define `sample_interval` property in SamplerWather and would access the field using it.

BTW, it if we'll define `sample_interval` and `rss_summary` as properties, it would more in sync with the surrounding code.

listeners.py
```diff
@@ -50,7 +56,58 @@ def process_result(self, obj):
                                       obj.result_checksum,
                                       obj.show_reproduce_content))
 
+    def get_failed_task_name(self, task):
+        return task[0][0] + ((':' + task[0][1]) if task[0][1] else '')
```
Totktonada, line 63: Unused?

listeners.py
```diff
@@ -50,7 +56,58 @@ def process_result(self, obj):
...
+
+    def get_long_mark(self, task):
+        return '(long)' if task in self.long_tasks else ''
+
+    def pretty_task_name(self, task_id):
+        return task_id[0] + ((':' + task_id[1]) if task_id[1] else '')
```
Totktonada, line 69: Just got that 'pretty' is noun, sorry. It is better to use 'prettify' — verb.

lib/test.py
```diff
@@ -130,6 +131,7 @@ def __init__(self, name, args, suite_ini, params={}, conf_name=None):
         self.is_terminated = False
         self.run_params = params
         self.conf_name = conf_name
+        self.started_time = time.time()
```
Totktonada, line 134: Hm, it seems we don't need to store this value within the Test instance, right?

Totktonada: More I think how to better handle succeeded and failed tests both, more I find the Test.run() method too long already. How about moving time.time() calls outside: into TestSuite.run_test()?

listeners.py
Totktonada: However statistics will be here always after the patch with adding timings. I would add `if os.path.isdir()` anyway.

Totktonada: Ah, I see, the directory is created always. Okay, at least it simplifies the code.

listeners.py
avtikhon: Right, done.

listeners.py
avtikhon: Right, removed.

lib/sampler.py
avtikhon: Sure, changed.

lib/sampler.py
avtikhon: Sure, changed.

listeners.py
avtikhon: Sure, changed.

lib/sampler.py
avtikhon: Ok, done.

lib/test.py
avtikhon: Actually we have rerun inside Test.run() method and going outside of it will produce the sum of duration on rerun test instead of the latest duration as it is now.

Totktonada:
Pushed several fixups (mostly around naming and wording). Please, apply if you don't object against them. The fixup commits arranged to ease squashing.

<details>
<summary>Patches.</summary>

```diff
commit 38067df314b47d8c542f8ccf6453a128f3b2d6f2
Author: Alexander Turenko <alexander.turenko@tarantool.org>
Date:   Sun May 23 03:23:54 2021 +0300

    FIXUP: Actualize comment in print_statistics()

diff --git a/listeners.py b/listeners.py
index 619121e..88d5bb4 100644
--- a/listeners.py
+++ b/listeners.py
@@ -104,13 +104,13 @@ class StatisticsWatcher(BaseWatcher):
         fd.close()
 
     def print_statistics(self):
+        """Print statistics and results of testing."""
         # Prepare standalone subpath '<vardir>/statistics' for statistics files.
         stats_dir = os.path.join(Options().args.vardir, 'statistics')
         safe_makedirs(stats_dir)
 
         self.print_rss_summary(stats_dir)
 
-        """Returns are there failed tasks."""
         if self.stats:
             color_stdout('Statistics:\n', schema='test_var')
         for short_status, cnt in self.stats.items():
```

```diff
commit ef322a34fc1fc089e21c104504391135551cc578
Author: Alexander Turenko <alexander.turenko@tarantool.org>
Date:   Sun May 23 02:27:47 2021 +0300

    FIXUP: sampler: don't use private API

diff --git a/lib/sampler.py b/lib/sampler.py
index 92cbee3..ca977da 100644
--- a/lib/sampler.py
+++ b/lib/sampler.py
@@ -44,6 +44,10 @@ class SamplerWatcher(object):
     def process_timeout(self, delta_seconds):
         self._wakeup()
 
+    @property
+    def sample_interval(self):
+        return self._sample_interval
+
     def _wakeup(self):
         """Invoke Sampler.sample() if enough time elapsed since
            the previous call.
@@ -94,7 +98,7 @@ class Sampler:
 
     @property
     def sample_interval(self):
-        return self._watcher._sample_interval
+        return self._watcher.sample_interval
 
     @property
     def watcher(self):
```

```diff
commit db21cb9fffa19d36bc7f1758ea084c3605de4255
Author: Alexander Turenko <alexander.turenko@tarantool.org>
Date:   Sun May 23 02:31:53 2021 +0300

    FIXUP: sampler: fixups for is_enabled
    
    * Use proc_stat_rss_supported() directly: the access is not frequent, we
      don't need to cache the value in the instance's field.
    * Convert from a method to a property (it is already accessed as a
      property in dispatcher).

diff --git a/lib/sampler.py b/lib/sampler.py
index ca977da..753ef5b 100644
--- a/lib/sampler.py
+++ b/lib/sampler.py
@@ -79,7 +79,6 @@ class Sampler:
 
         self._processes = dict()
         self._rss_summary = dict()
-        self.proc_stat_rss_supported = proc_stat_rss_supported()
 
     def set_queue(self, queue, worker_id, worker_name):
         # Called from a worker process (_run_worker()).
@@ -107,8 +106,9 @@ class Sampler:
                                'main test-run process')
         return self._watcher
 
+    @property
     def is_enabled(self):
-        return self.proc_stat_rss_supported
+        return proc_stat_rss_supported()
 
     def register_process(self, pid, task_id, server_name, worker_id=None,
                          worker_name=None):
diff --git a/listeners.py b/listeners.py
index 88d5bb4..d5ab0df 100644
--- a/listeners.py
+++ b/listeners.py
@@ -65,7 +65,7 @@ class StatisticsWatcher(BaseWatcher):
 
     # RSS.
     def print_rss_summary(self, stats_dir):
-        if not self._sampler.is_enabled():
+        if not self._sampler.is_enabled:
             return
 
         rss_summary = self._sampler.rss_summary
```

```diff
commit 9dff43866ffff152119998046e0e4021f7583841
Author: Alexander Turenko <alexander.turenko@tarantool.org>
Date:   Sun May 23 03:13:02 2021 +0300

    FIXUP: Adjust wording in memory statistics report

diff --git a/listeners.py b/listeners.py
index d5ab0df..f4bd9b7 100644
--- a/listeners.py
+++ b/listeners.py
@@ -73,9 +73,7 @@ class StatisticsWatcher(BaseWatcher):
 
         # Print to stdout RSS statistics for all failed tasks.
         if self.failed_tasks:
-            color_stdout('RSS of the failed tasks in Mb:', schema='info')
-            color_stdout(' (may skip tests run faster than {} sec):\n'.format(
-                         self._sampler.sample_interval), schema='info')
+            color_stdout('Occupied memory in failed tests (RSS, Mb):\n', schema='info')
             for task in self.failed_tasks:
                 task_id = task[0]
                 if task_id in rss_summary:
@@ -83,17 +81,22 @@ class StatisticsWatcher(BaseWatcher):
                                  self.prettify_task_name(task_id).ljust(self.field_size),
                                  self.get_long_mark(task_id)),
                                  schema='info')
+            color_stdout('\n')
 
         # Print to stdout RSS statistics for some number of most it used tasks.
-        color_stdout('Up to {} most RSS used tasks in Mb'.format(top_rss), schema='info')
-        color_stdout(' (may skip tests run faster than {} sec):\n'.format(
-                     self._sampler.sample_interval), schema='info')
+        color_stdout('Top {} tests by occupied memory (RSS, Mb):\n'.format(
+                     top_rss), schema='info')
         results_sorted = sorted(rss_summary.items(), key=lambda x: x[1], reverse=True)
         for task_id, rss in results_sorted[:top_rss]:
             color_stdout('* %6.1f %s %s\n' % (float(rss) / 1024,
                          self.prettify_task_name(task_id).ljust(self.field_size),
                          self.get_long_mark(task_id)), schema='info')
-        color_stdout('-' * 41, "\n", schema='separator')
+        color_stdout('\n')
+
+        color_stdout('(Tests quicker than {} seconds may be missed.)\n'.format(
+                     self._sampler.sample_interval), schema='info')
+
+        color_stdout('-' * 81, "\n", schema='separator')
 
         # Print RSS statistics to '<vardir>/statistics/rss.log' file.
         filepath = os.path.join(stats_dir, 'rss.log')
```

```diff
commit 4eadef6f502d0df71e2e5c5afcaeb440d7b51908
Author: Alexander Turenko <alexander.turenko@tarantool.org>
Date:   Sun May 23 03:15:25 2021 +0300

    FIXUP: sampler: adjust wording in rss_summary comment

diff --git a/lib/sampler.py b/lib/sampler.py
index 753ef5b..eb83d89 100644
--- a/lib/sampler.py
+++ b/lib/sampler.py
@@ -89,10 +89,7 @@ class Sampler:
 
     @property
     def rss_summary(self):
-        """Return dictionary of collected maximum RSS values per each
-           run task, set with task_id keys.
-        """
-
+        """Task ID to maximum RSS mapping."""
         return self._rss_summary
 
     @property
```

```diff
commit 1df77515ec8fd2c75845236eb63dfa778256e1b2
Author: Alexander Turenko <alexander.turenko@tarantool.org>
Date:   Sun May 23 03:20:45 2021 +0300

    FIXUP: Adjust wording in duration statistics

diff --git a/listeners.py b/listeners.py
index 231ea5c..d90c070 100644
--- a/listeners.py
+++ b/listeners.py
@@ -115,7 +115,8 @@ class StatisticsWatcher(BaseWatcher):
 
         # Print to stdout durations for all failed tasks.
         if self.failed_tasks:
-            color_stdout('Duration of the failed tasks:\n', schema='info')
+            color_stdout('Duration of failed tests (seconds):\n',
+                         schema='info')
             for task in self.failed_tasks:
                 task_id = task[0]
                 if task_id in self.tasks_durations:
@@ -123,16 +124,18 @@ class StatisticsWatcher(BaseWatcher):
                                  self.prettify_task_name(task_id).ljust(self.field_size),
                                  self.get_long_mark(task_id)),
                                  schema='info')
+            color_stdout('\n')
 
         # Print to stdout durations for some number of most long tasks.
-        color_stdout('Up to {} most long tasks in seconds:\n'.format(top_durations),
+        color_stdout('Top {} longest tests (seconds):\n'.format(top_durations),
                      schema='info')
         results_sorted = sorted(self.tasks_durations.items(), key=lambda x: x[1], reverse=True)
         for task_id, duration in results_sorted[:top_durations]:
             color_stdout('* %6.2f %s %s\n' % (duration,
                          self.prettify_task_name(task_id).ljust(self.field_size),
                          self.get_long_mark(task_id)), schema='info')
-        color_stdout('-' * 41, "\n", schema='separator')
+
+        color_stdout('-' * 81, "\n", schema='separator')
 
         # Print durations statistics to '<vardir>/statistics/duration.log' file.
         filepath = os.path.join(stats_dir, 'duration.log')
```

```diff
commit 44322a8ff1615c530f227461835bd2639a4217b7
Author: Alexander Turenko <alexander.turenko@tarantool.org>
Date:   Sun May 23 03:29:47 2021 +0300

    FIXUP: Adjust wording: tasks_durations -> duration_stats

diff --git a/listeners.py b/listeners.py
index d90c070..107e57e 100644
--- a/listeners.py
+++ b/listeners.py
@@ -36,7 +36,7 @@ class StatisticsWatcher(BaseWatcher):
         self.stats = dict()
         self.field_size = 60
         self._sampler = sampler
-        self.tasks_durations = dict()
+        self.duration_stats = dict()
         self.failed_tasks = []
         self.get_logfile = get_logfile
         self.long_tasks = set()
@@ -58,7 +58,7 @@ class StatisticsWatcher(BaseWatcher):
                                       obj.result_checksum,
                                       obj.show_reproduce_content))
 
-        self.tasks_durations[obj.task_id] = obj.duration
+        self.duration_stats[obj.task_id] = obj.duration
 
     def get_long_mark(self, task):
         return '(long)' if task in self.long_tasks else ''
@@ -119,8 +119,8 @@ class StatisticsWatcher(BaseWatcher):
                          schema='info')
             for task in self.failed_tasks:
                 task_id = task[0]
-                if task_id in self.tasks_durations:
-                    color_stdout('* %6.2f %s %s\n' % (self.tasks_durations[task_id],
+                if task_id in self.duration_stats:
+                    color_stdout('* %6.2f %s %s\n' % (self.duration_stats[task_id],
                                  self.prettify_task_name(task_id).ljust(self.field_size),
                                  self.get_long_mark(task_id)),
                                  schema='info')
@@ -129,7 +129,7 @@ class StatisticsWatcher(BaseWatcher):
         # Print to stdout durations for some number of most long tasks.
         color_stdout('Top {} longest tests (seconds):\n'.format(top_durations),
                      schema='info')
-        results_sorted = sorted(self.tasks_durations.items(), key=lambda x: x[1], reverse=True)
+        results_sorted = sorted(self.duration_stats.items(), key=lambda x: x[1], reverse=True)
         for task_id, duration in results_sorted[:top_durations]:
             color_stdout('* %6.2f %s %s\n' % (duration,
                          self.prettify_task_name(task_id).ljust(self.field_size),
@@ -137,12 +137,12 @@ class StatisticsWatcher(BaseWatcher):
 
         color_stdout('-' * 81, "\n", schema='separator')
 
-        # Print durations statistics to '<vardir>/statistics/duration.log' file.
+        # Print duration statistics to '<vardir>/statistics/duration.log' file.
         filepath = os.path.join(stats_dir, 'duration.log')
         fd = open(filepath, 'w')
-        for task_id in self.tasks_durations:
+        for task_id in self.duration_stats:
             fd.write("{} {}\n".format(self.prettify_task_name(task_id),
-                                      self.tasks_durations[task_id]))
+                                      self.duration_stats[task_id]))
         fd.close()
 
     def print_statistics(self):
```
</details>

avtikhon:
> Pushed several fixups (mostly around naming and wording). Please, apply if you don't object against them. The fixup commits arranged to ease squashing.

All patches LGTM, squashed it, thank you.

Totktonada (approved):
LGTM.

Totktonada:
Updated the test-run submodule in tarantool in the following commits: [2.9.0-35-g213f480e7](https://github.com/tarantool/tarantool/commit/213f480e74e549ba337daacc56dc83c60958f485), 2.8.1-19-g67ff7cc5f, 2.7.2-15-gc35655113, 1.10.10-8-gfb804dcca.

Issue #199, opened 2020-02-06
Support 'release_disabled' for app and unittest suites


Totktonada:
It is supported only for 'tarantool' suites for now.

It is used in tarantool's app and unittest suites and so the corresponding tests are never run.

Totktonada:
See 'debug' field / property in tarantool_server.py, app_server,py, unittest_server.py.

Totktonada:
Estimated as 2sp assuming that we need to find a way to test that the 'release_disabled' app and unittest tests are run.

PR #299, opened 2021-05-11
Add release_disabled field to app- and unit- tests

VitaliyaIoffe:
Previously if suite.ini has a release_disabled field it works only for
tarantool tests. This commit allows to use it for app- and unittests
suites.

Despite unittest does not take tarantool binary, unittests get version
(release/debug) from the tarantool binary information.

For instance,

Add line to test-unit/suite.ini:
`release_disabled = broken_unicode.test`

On debug version:
`[001] test-unit/broken_unicode.test                                   [ pass ]`

On release version:
`[001] test-unit/broken_unicode.test                                   [ disabled ]`

Closes: #199

### DIFF
--- a/lib/app_server.py
+++ b/lib/app_server.py
@@ -1,6 +1,7 @@
 import errno
 import glob
 import os
+import re
 import shutil
 import signal
 import sys
@@ -107,11 +108,9 @@ class AppServer(Server):
         self.testdir = os.path.abspath(os.curdir)
         self.vardir = ini['vardir']
         self.builddir = ini['builddir']
-        self.debug = False
         self.lua_libs = ini['lua_libs']
         self.name = 'app_server'
         self.process = None
-        self.binary = TarantoolServer.binary
         self.use_unix_sockets_iproto = ini['use_unix_sockets_iproto']
 
     @property
@@ -222,6 +221,9 @@ class AppServer(Server):
     @classmethod
     def find_exe(cls, builddir):
         cls.builddir = builddir
+        cls.binary = TarantoolServer.binary
+        cls.debug = bool(re.findall(r'-Debug', str(cls.version()),
+                                    re.I))
 
     @staticmethod
     def find_tests(test_suite, suite_path):
--- a/lib/__init__.py
+++ b/lib/__init__.py
@@ -5,6 +5,7 @@ import shutil
 from lib.options import Options
 from lib.tarantool_server import TarantoolServer
 from lib.unittest_server import UnittestServer
+from lib.app_server import AppServer
 from lib.utils import warn_unix_sockets_at_start
 
 
@@ -62,6 +63,7 @@ def module_init():
 
     TarantoolServer.find_exe(args.builddir)
     UnittestServer.find_exe(args.builddir)
+    AppServer.find_exe(args.builddir)
 
     Options().check_schema_upgrade_option(TarantoolServer.debug)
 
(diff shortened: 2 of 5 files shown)


lib/app_server.py
```diff
@@ -17,7 +19,7 @@
 from lib.tarantool_server import Test
 from lib.tarantool_server import TarantoolServer
 from lib.tarantool_server import TarantoolStartError
-from lib.utils import find_port
+from lib.utils import find_port, bytes_to_str
```
Totktonada, line 22: Please, keep the style: one import object per line.

lib/app_server.py
```diff
@@ -219,9 +219,19 @@ def kill():
         self.process.wait()
         timer.cancel()
 
+    @classmethod
+    def version(cls):
+        p = subprocess.Popen([cls.binary, "--version"], stdout=subprocess.PIPE)
+        version = bytes_to_str(p.stdout.read()).rstrip()
+        p.wait()
+        return version
```
Totktonada, line 227: Can we reuse existing `version()` classmethod from TarantoolServer in some way (directly or by moving it to Server or even to utils)?

lib/app_server.py
VitaliyaIoffe: Ok, I'll change it.
FYI, I relied on this document
https://www.tarantool.io/en/doc/latest/dev_guide/python_style_guide/#imports

lib/app_server.py
VitaliyaIoffe: Moved to Server class

Totktonada:
First of all, sql-tap/debug_mode_only.test.lua fails now. I looked around a bit and there are several facts about it:

* The test leans on the `sql_current_time` variable, which is not exported from the executable since [2.0.4-838-g32cb1ad29](https://github.com/tarantool/tarantool/commit/32cb1ad298b2b55d8536a85bdfb3827c8c8739e1).
* The test is about date/time types in SQL. Support of those types should be restored in the scope of https://github.com/tarantool/tarantool/issues/3694.

Since @tsafin assigned to this task I would ask him to decide what to do with the test: should we disable it now (to fix in the scope of his task) or remove at all?

tsafin:
> * The test is about date/time types in SQL. Support of those types should be restored in the scope of https://github.com/tarantool/tarantool/issues/3694.
> 
> Since @tsafin assigned to this task I would ask him to decide what to do with the test: should we disable it now (to fix in the scope of his task) or remove at all?

Yes, please disable it - we will return here quite soon, and reinstate the code

Totktonada:
> Yes, please disable it - we will return here quite soon, and reinstate the code

Okay, I'll proceed with that, when I'll update the test-run submodule.

Totktonada (approved):
This change looks minimalistic and solves the problem. I would refactor `find_exe()` and eliminate AppServer / UnittestServer dependency on TarantoolServer, but I'll do it as a follow up. The patch per se LGTM.

avtikhon (approved):
Patch LGTM

Totktonada:
Updated the test-run submodule in tarantool in the following commits: [2.9.0-30-g7bee61531](https://github.com/tarantool/tarantool/commit/7bee61531e19685c6715835cdbcd4215fb7747bd), 2.8.1-16-gb54bfb86b, 2.7.2-12-gbaa8c8212, 1.10.10-5-g047b910d5.

Issue #276, opened 2021-03-19
test-run worker hangs, when an instance does not enter to the event loop


ligurio:
### How to reproduce

test reproducer (test path `xlog/hang.test.lua`):
```
test_run = require('test_run').new()
test_run:cmd('create server replica with rpl_master=default, script="xlog/replica.lua"')
test_run:cmd('start server replica')
test_run:cmd('cleanup server replica')
test_run:cmd('delete server replica')
```

test hang due to absence read access (see log `var/001_xlog/replica.log`):
```
...
2021-03-19 18:11:44.865 [1987357] main/103/replica I> connected to 1 replicas
2021-03-19 18:11:44.865 [1987357] main/103/replica I> bootstrapping replica from 8a89e9f8-7364-4d6a-96cd-d2c8fe5b93bb at unix/:/home/sergeyb/sources/MRG/tarantool/build/test/var/001_xlog/xlog.socket-iproto
2021-03-19 18:11:44.865 [1987357] main/112/applier/unix/:/home/sergeyb/sources/MRG/tarantool/build/test/var/001_xlog/xlog.socket-iproto I> can't read row
2021-03-19 18:11:44.865 [1987357] main/112/applier/unix/:/home/sergeyb/sources/MRG/tarantool/build/test/var/001_xlog/xlog.socket-iproto session.cc:332 E> ER_ACCESS_DENIED: Read access to universe '' is denied for user 'guest'
2021-03-19 18:11:44.865 [1987357] main/112/applier/unix/:/home/sergeyb/sources/MRG/tarantool/build/test/var/001_xlog/xlog.socket-iproto I> will retry every 1.00 second
```	

test is passed if add line `box.schema.user.grant('guest', 'replication')` on top of test

### Versions

test-run 59417412699c46d6e498301794cc3c3d6298d255

`tarantool --version`:
```
Tarantool 2.8.0-134-g81c663335
Target: Linux-x86_64-Debug
Build options: cmake . -DCMAKE_INSTALL_PREFIX=/usr/local -DENABLE_BACKTRACE=ON
Compiler: /usr/bin/cc /usr/bin/c++
[...]

Issue #274, opened 2021-03-16
Convert reproducers from PR comments to tests


Totktonada:
To be done after PR #273.

This one for example: https://github.com/tarantool/test-run/pull/269#issuecomment-789373803

I'll find several more ones in PR comments (and maybe commit comments) and will update the issue.

Totktonada:
One more: https://github.com/tarantool/test-run/pull/281#issuecomment-808248408

Totktonada:
More: https://github.com/tarantool/test-run/pull/244#issuecomment-739628746

ylobankov:
One more: https://github.com/tarantool/test-run/pull/302#discussion_r639084286

Issue #252, opened 2020-12-15
Crash detector seems being broken


Totktonada:
### System information

Linux.
test-run: e84355283f117e91964fe84eac70f26bb32d7f06.
tarantool: 2.7.0-111-g28f3b2f1e.
python2: 2.7.17.
gevent: 1.5_alpha2 and 20.6.2.
greenlet: 0.4.15-r1 and 0.4.16.

### How to observe the problem

1. Pull and build recent tarantool (2.7.0-111-g28f3b2f1e in my case).
2. Copy `test/app-tap/test-timeout.test.lua` from https://github.com/tarantool/test-run/pull/244#issuecomment-739628746 (don't forget to set the executable bit: `chmod a+x test/app-tap/test-timeout.test.lua`).
3. Mangle `test/app-tap/debug/server.lua` to fail after 1 second. Place this code at the end of the file:

   ```lua
   local fiber = require('fiber')
   fiber.create(function()
       fiber.sleep(1)
       os.exit(1)
   end)
   ```
4. Run the test: `./test/test-run.py -j1 app-tap/test-timeout.test.lua`.

Expected: the fail of the non-default server detected and the testing fails with appropriate message after ~1 second.

Got: fail after 120 seconds (default --no-output-timeout value), no report about the fail of the non-default server.

### Investigation

I observed that `self.process.returncode` in `TarantoolServer.crash_detect()` is 0, while the process returns 1.

After any of the following two patches the exit code becomes correct.

Variant 1:

```diff
diff --git a/lib/tarantool_server.py b/lib/tarantool_server.py
index 481b08f..6624ccf 100644
--- a/lib/tarantool_server.py
+++ b/lib/tarantool_server.py
@@ -8,7 +8,7 @@ import re
 import shlex
 import shutil
 import signal
-import subprocess
+from gevent import subprocess
 import sys
 import time
 import yaml
```

Variant 2:

```diff
diff --git a/lib/tarantool_server.py b/lib/tarantool_server.py
index 481b08f..a26e7c6 100644
--- a/lib/tarantool_server.py
+++ b/lib/tarantool_server.py
@@ -928,7 +928,7 @@ class TarantoolServer(Server):
         while self.process.returncode is None:
             self.process.poll()
             if self.process.returncode is None:
-                gevent.sleep(0.1)
+                time.sleep(0.1)
 
         if self.process.returncode in [0, -signal.SIGKILL, -signal.SIGTERM]:
             return
```

[...]

PR #302, opened 2021-05-17
Add timeout for starting tarantool server

### LINKED ISSUE #256: test-run does not kill server that hangs in an instance script

VitaliyaIoffe:
This patch includes 4 commits:

**Use gevent sleep for switching context**

`time.sleep` was changed to `gevent.sleep` to allow current greenlet to
sleep and others to run. When `time.sleep` is used, greenlet's context
is not changed from the main process to test's greenlet.

As a result of this, there is no data received by the main process while
hanging the tarantool server process and the test is fallen down by the
common timeout (NO_OUTPUT_TIMEOUT). Even worse, the tarantool server
process is not killed by test-run.

Using `gevent.sleep` makes the test fail by the test timeout and kill
the farantool server process.

Part of #276

**Add timeout for starting tarantool server**

When a tarantool server starts, it waits for a special pattern in the
log file to proceed. If there is no pattern present, the server hangs.
After the test timeout (TEST_TIMEOUT) runs out, the test fails.

This patch adds the `--server-start-timeout` option to test-run (by
default it equals to 90 seconds). Now when the server hangs and the
time (SERVER_START_TIMEOUT) runs out, a comprehensible exception is
raised with the message that the server failed to start within the
timeout.

Fixes: #276

**Kill processes of non-started tarantool servers**

It was found that processes of non-started tarantool servers are not
killed by test-run and leave to hang. This situation can be reproduced
by creating the main server, then creating a replica server, but the
replica server is unable to join the master, for example, due to lack
of user permissions. In this case, the test fails by the server start
timeout and test-run kills the main server process only.

Fixes #256
Follows up #276

**Add unit test checking no hanging procs present**

This patch adds a simple unit test checking that if a tarantool server
failed to start within a certain amount of seconds, test-tun raises a
comprehensible exception and kills the server process.

Follows up #256
Follows up #276

### DIFF
--- a/.luacheckrc
+++ b/.luacheckrc
@@ -1,4 +1,4 @@
-globals = {"box", "_TARANTOOL", "tonumber64", "os"}
+globals = {"box", "_TARANTOOL", "tonumber64", "os", "test_run"}
 ignore = {
     -- Accessing an undefined field of a global variable <debug>.
     "143/debug",
--- a/lib/options.py
+++ b/lib/options.py
@@ -214,6 +214,14 @@ class Options(object):
                 Note: The option works now only with parallel testing.""")
 
         parser.add_argument(
+                "--server-start-timeout",
+                dest="server_start_timeout",
+                default=env_int('SERVER_START_TIMEOUT', 90),
+                type=int,
+                help="""Stop the server process if the server starts longer than
+                this amount of seconds. Default: 90 [seconds].""")
+
+        parser.add_argument(
                 "--test-timeout",
                 dest="test_timeout",
                 default=env_int('TEST_TIMEOUT', 110),
--- a/lib/tarantool_server.py
+++ b/lib/tarantool_server.py
@@ -1016,16 +1016,6 @@ class TarantoolServer(Server):
             color_log('Server [%s] start against running ...\n',
                       schema='test_var')
             return
-        if self.status != 'started':
-            if not silent:
-                raise Exception('Server is not started')
-            else:
-                color_log(
-                    'Server [{0.name}] is not started '
-                    '(status:{0.status}) ...\n'.format(self),
-                    schema='test_var'
-                )
-            return
         if not silent:
             color_stdout('[Instance {}] Stopping the server...\n'.format(
                 self.name), schema='info')
--- a/lib/tarantool_server.py
+++ b/lib/tarantool_server.py
@@ -414,8 +414,15 @@ CON_SWITCH = {
 
 
 class TarantoolStartError(OSError):
-    def __init__(self, name=None):
+    def __init__(self, name=None, timeout=None):
         self.name = name
+        self.timeout = timeout
+
+    def __str__(self):
+        message = '[Instance "{}"] Failed to start'.format(self.name)
+        if self.timeout:
+            return "\n{} within {} seconds\n".format(message, self.timeout)
+        return "\n{}\n".format(message)
 
 
 class TarantoolLog(object):
@@ -444,16 +451,18 @@ class TarantoolLog(object):
                 if pos != -1:
                     return pos
 
-    def seek_wait(self, msg, proc=None, name=None):
-        while True:
+    def seek_wait(self, msg, proc=None, name=None, deadline=None):
+        while not deadline or time.time() < deadline:
             if os.path.exists(self.path):
                 break
             gevent.sleep(0.001)
+        else:
+            return False
 
         with open(self.path, 'r') as f:
             f.seek(self.log_begin, os.SEEK_SET)
             cur_pos = self.log_begin
-            while True:
+            while not deadline or time.time() < deadline:
                 if not (proc is None):
                     if not (proc.poll() is None):
                         raise TarantoolStartError(name)
@@ -463,8 +472,9 @@ class TarantoolLog(object):
                     f.seek(cur_pos, os.SEEK_SET)
                     continue
                 if re.findall(msg, log_str):
-                    return
+                    return True
                 cur_pos = f.tell()
+        return False
 
 
 class TarantoolServer(Server):
@@ -873,9 +883,10 @@ class TarantoolServer(Server):
         self.crash_detector.start()
 
         if wait:
+            deadline = time.time() + Options().args.server_start_timeout
             try:
-                self.wait_until_started(wait_load)
-            except TarantoolStartError:
+                self.wait_until_started(wait_load, deadline)
+            except TarantoolStartError as err:
                 # Python tests expect we raise an exception when non-default
                 # server fails
                 if self.crash_expected:
@@ -884,9 +895,7 @@ class TarantoolServer(Server):
                         self.current_test.is_crash_reported):
                     if self.current_test:
                         self.current_test.is_crash_reported = True
-                    color_stdout('\n[Instance "{0.name}"] Tarantool server '
-                                 'failed to start\n'.format(self),
-                                 schema='error')
+                    color_stdout(err, schema='error')
                     self.print_log(15)
                 # Raise exception when caller ask for it (e.g. in case of
                 # non-default servers)
@@ -1084,7 +1093,19 @@ class TarantoolServer(Server):
         self.wait_until_stopped(pid)
         return True
 
-    def wait_until_started(self, wait_load=True):
+    def wait_load(self, deadline):
+        """Wait until the server log file is matched the entry pattern
+
+        If the entry pattern couldn't be found in a log file until a timeout
+        is up, it will raise a TarantoolStartError exception.
+        """
+        msg = 'entering the event loop|will retry binding|hot standby mode'
+        p = self.process if not self.gdb and not self.lldb else None
+        if not self.logfile_pos.seek_wait(msg, p, self.name, deadline):
+            raise TarantoolStartError(
+                self.name, Options().args.server_start_timeout)
+
+    def wait_until_started(self, wait_load=True, deadline=None):
         """ Wait until server is started.
 
         Server consists of two parts:
@@ -1095,12 +1116,9 @@ class TarantoolServer(Server):
         color_log('DEBUG: [Instance {}] Waiting until started '
                   '(wait_load={})\n'.format(self.name, str(wait_load)),
                   schema='info')
-
         if wait_load:
-            msg = 'entering the event loop|will retry binding|hot standby mode'
-            p = self.process if not self.gdb and not self.lldb else None
-            self.logfile_pos.seek_wait(msg, p, self.name)
-        while True:
+            self.wait_load(deadline)
+        while not deadline or time.time() < deadline:
             try:
                 temp = AdminConnection('localhost', self.admin.port)
                 if not wait_load:
@@ -1126,6 +1144,9 @@ class TarantoolServer(Server):
                     gevent.sleep(0.1)
                     continue
                 raise
+        else:
+            raise TarantoolStartError(
+                self.name, Options().args.server_start_timeout)
 
     def wait_until_stopped(self, pid):
         while True:
--- a/lib/tarantool_server.py
+++ b/lib/tarantool_server.py
@@ -448,7 +448,7 @@ class TarantoolLog(object):
         while True:
             if os.path.exists(self.path):
                 break
-            time.sleep(0.001)
+            gevent.sleep(0.001)
 
         with open(self.path, 'r') as f:
             f.seek(self.log_begin, os.SEEK_SET)
@@ -459,7 +459,7 @@ class TarantoolLog(object):
                         raise TarantoolStartError(name)
                 log_str = f.readline()
                 if not log_str:
-                    time.sleep(0.001)
+                    gevent.sleep(0.001)
                     f.seek(cur_pos, os.SEEK_SET)
                     continue
                 if re.findall(msg, log_str):
@@ -1123,14 +1123,14 @@ class TarantoolServer(Server):
                 if e.errno == errno.ECONNREFUSED:
                     color_log(' | Connection refused; will retry every 0.1 '
                               'seconds...')
-                    time.sleep(0.1)
+                    gevent.sleep(0.1)
                     continue
                 raise
 
     def wait_until_stopped(self, pid):
         while True:
             try:
-                time.sleep(0.01)
+                gevent.sleep(0.01)
                 os.kill(pid, 0)
                 continue
             except OSError:
@@ -1221,7 +1221,7 @@ class TarantoolServer(Server):
     def wait_lsn(self, node_id, lsn):
         while (self.get_lsn(node_id) < lsn):
             # print("wait_lsn", node_id, lsn, self.get_param("vclock"))
-            time.sleep(0.01)
+            gevent.sleep(0.01)
 
     def get_log(self):
         return TarantoolLog(self.logfile).positioning()
--- a/test-run.py
+++ b/test-run.py
@@ -91,6 +91,8 @@ def main_loop_parallel():
 
     color_stdout("Timeout options:\n", schema='tr_text')
     color_stdout('-' * 19, "\n",       schema='separator')
+    color_stdout("SERVER_START_TIMEOUT:" . ljust(26) + "{}\n" .
+                 format(args.server_start_timeout), schema='tr_text')
     color_stdout("REPLICATION_SYNC_TIMEOUT:" . ljust(26) + "{}\n" .
                  format(args.replication_sync_timeout), schema='tr_text')
     color_stdout("TEST_TIMEOUT:" . ljust(26) + "{}\n" .
--- a/test/unittest/box-cc0544b6afd1.lua
+++ b/test/unittest/box-cc0544b6afd1.lua
@@ -1,0 +1,7 @@
+#!/usr/bin/env tarantool
+
+box.cfg {
+    listen = os.getenv('LISTEN'),
+}
+
+require('console').listen(os.getenv('ADMIN'))
--- a/test/unittest/hang.test.lua
+++ b/test/unittest/hang.test.lua
@@ -1,0 +1,6 @@
+test_run = require('test_run').new()
+
+-- This test should hang: we are unable to bootstrap the replica, because it is
+-- unable to join the master because of lack of granting user permissions.
+test_run:cmd('create server replica with rpl_master=default, script="unittest/replica-7f4d4895ff58.lua"')
+test_run:cmd('start server replica')
--- a/test/unittest/replica-7f4d4895ff58.lua
+++ b/test/unittest/replica-7f4d4895ff58.lua
@@ -1,0 +1,8 @@
+#!/usr/bin/env tarantool
+
+box.cfg {
+    listen = os.getenv('LISTEN'),
+    replication = os.getenv('MASTER'),
+}
+
+require('console').listen(os.getenv('ADMIN'))
--- a/test/unittest/suite.ini
+++ b/test/unittest/suite.ini
@@ -1,0 +1,6 @@
+[default]
+core = tarantool
+description = unit tests
+script = box-cc0544b6afd1.lua
+use_unix_sockets = True
+use_unix_sockets_iproto = True
--- a/test/unittest/test_tarantool_server.py
+++ b/test/unittest/test_tarantool_server.py
@@ -1,0 +1,47 @@
+import os
+import subprocess
+import sys
+import unittest
+
+
+class TestTarantoolServer(unittest.TestCase):
+    def test_tarantool_server_not_hanging(self):
+        env = os.environ.copy()
+        env['SERVER_START_TIMEOUT'] = '5'
+
+        cmd = [sys.executable, 'test/test-run.py', 'unittest/hang.test.lua']
+
+        # File names intentionally have hashes to find exactly these processes
+        # using 'ps' command.
+        box_instance = 'box-cc0544b6afd1'
+        replica_instance = 'replica-7f4d4895ff58'
+
+        err_msg_1 = ('[Instance "%s"] Failed to start tarantool '
+                     'instance "%s"' % (box_instance, replica_instance))
+        err_msg_2 = ('[Instance "%s"] Failed to start within %s seconds'
+                     % (replica_instance, env['SERVER_START_TIMEOUT']))
+
+        try:
+            subprocess.check_output(cmd, env=env, universal_newlines=True)
+            self.fail("Command `%s` did not return non-zero exit code"
+                      % ' '.join(cmd))
+        except subprocess.CalledProcessError as exc:
+            err_obj = exc
+
+        self.assertIn(err_msg_1, err_obj.output)
+        self.assertIn(err_msg_2, err_obj.output)
+
+        ps_lines = subprocess.check_output(
+            ['ps', '-o', 'command'], universal_newlines=True
+        ).splitlines()
+        proc_lines = [line.strip() for line in ps_lines
+                      if 'tarantool %s.lua' % box_instance in line or
+                      'tarantool %s.lua' % replica_instance in line]
+
+        self.assertFalse(
+            proc_lines, 'There are some hanging tarantool processes!'
+        )
+
+
+if __name__ == '__main__':
+    unittest.main()
--- a/Makefile
+++ b/Makefile
@@ -15,7 +15,7 @@ luacheck:
 	luacheck --config .luacheckrc .
 
 test_integration:
-	PYTHONPATH=$(PROJECT_DIR) $(PYTHON) test/test-run.py --force $(TEST_RUN_EXTRA_PARAMS)
+	PYTHONPATH=$(PROJECT_DIR) $(PYTHON) test/test-run.py --force --exclude unittest $(TEST_RUN_EXTRA_PARAMS)
 
 test_unittest:
 	$(PYTHON) -m unittest discover test/unittest/


lib/tarantool_server.py
```diff
@@ -449,7 +449,7 @@ def seek_wait(self, msg, proc=None, name=None):
         while True:
             if os.path.exists(self.path):
                 break
-            time.sleep(0.001)
+            gevent.sleep(0.001)
```
Totktonada, line 452: Regarding commit message:

> Use gevent sleep

The title should give an idea about the change for someone, who didn't read the patch, in some concise way. The most descriptive way is often to describe what was changed from a user point of view.

> Was changed `time.sleep` to `gevent.sleep` to allow current greenlet to
> sleep and others to run.

This information is more or less obvious from the patch (however, sure, good to mention). However the commit message lacks non-obvious details: what is the greenlet we should unblock, what is the problem (and what are circumstances when it appears), what will be changed for a user?

Also, please, mark it as part of #276.

A commit message is written once, but will be read many times. It is worthful to maximize its usefulness. Especially at the point, when you remember every detail about the problem, about the patch, used tricks and so on.

I spin a lot around different code and I'm under a strict impression that development is mostly about expressing ideas rather than about tweaking code to make it work. At least in long living projects.

lib/tarantool_server.py
```diff
@@ -464,8 +465,9 @@ def seek_wait(self, msg, proc=None, name=None):
                     f.seek(cur_pos, os.SEEK_SET)
                     continue
                 if re.findall(msg, log_str):
-                    return
+                    return True
                 cur_pos = f.tell()
+        return False
```
Totktonada, line 470: The timeout situation does not look much different from just startup error. Can we raise `TarantoolStartError` with a descriptive error message in the case (it seems, we should add a parameter into its constructor and `__str__` method)?

lib/tarantool_server.py
```diff
@@ -454,7 +454,8 @@ def seek_wait(self, msg, proc=None, name=None):
         with open(self.path, 'r') as f:
             f.seek(self.log_begin, os.SEEK_SET)
             cur_pos = self.log_begin
-            while True:
+            current_time = time.time()
```
Totktonada, line 457: I would name it `start_time`, current naming is consufing. Or, as it is often written:

```python
deadline = time.time() + timeout
while time.time() < deadline:
    # <..workload..>
```

lib/tarantool_server.py
```diff
@@ -920,6 +924,7 @@ def start(self, silent=True, wait=True, wait_load=True, rais=True, args=[],
                                                    actual_version),
                              schema='error')
                 raise TarantoolStartError(self.name)
+        return True
```
Totktonada, line 927: What if we'll stuck in attempts to establish a connection in `wait_until_started()`? I think the whole `wait_until_started()` call should have one deadline, not just `wait_seek()`. Sketchy code to illustrate the idea:

```python
def start(<...>):
    deadline = time.time() + Options.args.server_start_timeout
    <..start tarantool..>
    wait_until_started(<...>, deadline)

def wait_until_started(<...>, deadline):
    wait_load(<...>, deadline)
    while time.time() < deadline:
        <..check connection..>

def wait_load(<...>, deadline):
    while time.time() < deadline:
        <..check logs..>
```

lib/preprocessor.py
```diff
@@ -206,8 +208,15 @@ def server_start(self, ctype, sname, opts):
             if crash_expected:
                 # disable crash detector
                 self.servers[sname].crash_expected = True
-            self.servers[sname].start(silent=True, rais=True, wait=wait,
-                                      wait_load=wait_load, args=args)
+            is_started = self.servers[sname].start(
+                    silent=True, rais=True, wait=wait,
+                    wait_load=wait_load, args=args,
+                    server_start_timeout=Options().args.server_start_timeout)
```
Totktonada, line 214: Are we actually need to pass the timeout from here? I would grab it from options somewhere deeper by the call stack: say, from inside TarantoolServer.start(). If we'll need to pass different startup timeouts, we'll implement it later. It is how the test timeout is implemented and I like the simplicity that the approach brings into the code.

lib/tarantool_server.py
```diff
@@ -829,14 +831,14 @@ def cleanup(self, *args, **kwargs):
                                              *args, **kwargs)
 
     def start(self, silent=True, wait=True, wait_load=True, rais=True, args=[],
-              **kwargs):
+              server_start_timeout=90, **kwargs):
```
Totktonada, line 834: If you'll agree with my proposal about calculating a deadline inside the `start()` method and passing it down to `wait_until_started()` and `wait_seek()`, this point will not be actual. However I'll note that the `server_start_timeout` name looks redundant in the `start()` of a `Server` subclass: just `timeout` would be enough. Now different names are used for the same timeout and it is a bit confusing I think.

lib/tarantool_server.py
Totktonada, line 452: There are other `time.sleep()` usages across the code. Can them lead to problems of the same kind? Say, `time.sleep(0.1)` in `wait_until_started()` — I guess it is easy to verify by removing a `console.listen()` call from the instance file.

I don't push you to reproduce and/or investigate all possible problems at once, but, please, look around and if you are not sure that there are no problems anymore, file a follow up issue. If you'll do, leave a link to #276 and back to this discussion.

Totktonada, line 452: To be honest, lack of negative test cases disquiets me more and more. We unlikely will break the general case, when all tests are passed, becase every test-run push is verified against tarantool's test suites (manually, by me). I'm not so sure about a negative and tricky cases like this one. Maybe it is good time to start collecting negative test cases somewhere in the repository? Even if it'll require some manual work to run and verify, it'll be a good start.

I would add a directory with negative test cases and would add some simple runner (Makefile / shell script / pytest — does not matter), which would do (sketchy code, I didn't run it):

```sh
#!/bin/sh

set -eu

cd negative_testing
rm -rf output
mkdir output
for test in */*.test.lua; do
    printf 'Starting %s\n' "${test}"

    suite="${test%/*}"
    output_file="output/${test%.test.lua}.out.txt"
    validate_script="${test%.test.lua}/.validate.sh"

    mkdir -p "output/${suite}"
    ./test-run.py "${test}" > "${output_file}" 2>&1 || true
    "./${validate_script}" < "${output_file}"

    printf 'The test %s passed\n' "${test}"
done
```

negative_testing/foo/bar.validate.sh would contain a script that verifies an output of test-run. Just grep in the simplest case.

Totktonada, line 452: Another point that disquiets me: whether a Popen handle gives a correct result, when we spin around `<Popen_handle>.pool()` and `gevent.sleep()`? I saw some strange effects and described them in #252. I would make an attempt to reveal a root of the problem and understand, whether a problem of this kind can appear here too. At least try and if'll not become clear, leave a note in #252 to revisit the server startup waiting code too when we'll investigate the issue.

Totktonada: (Or at least track it with an issue.)

lib/tarantool_server.py
VitaliyaIoffe: Do you want to raise the exception here? I'm not confident due to this function is about seeking a pattern. Of course, for a proc, which does not exist, is important. But in case of a pattern absence, I suggest returning the bool value. I may add the raise condition wait_until_started.

lib/tarantool_server.py
VitaliyaIoffe: I like this idea to prevent a connection stuck too, but in my opinion, `wait_load, wait_until_started` should take a timeout, not a deadline

lib/tarantool_server.py
VitaliyaIoffe: I hope I fixed it well.

lib/tarantool_server.py
VitaliyaIoffe: You're right. Thank you!

lib/preprocessor.py
VitaliyaIoffe: Fixed

lib/tarantool_server.py
VitaliyaIoffe: Fixed

lib/tarantool_server.py
VitaliyaIoffe: I was investigating it. Tbh, I tried your method but didn't understand clearly the situation. Can we return to this later?

VitaliyaIoffe: Fixed your script and added pytest script for comparing gold log and test-run log. Does it seem like an instrument which you want?

VitaliyaIoffe: As a comment about time/gevent sleep, I spent some time clarify it, but haven't received any good results. I suggest shifting it for the next time.

Totktonada: The title does not explain changes from a user perspective, but it is more or less descriptive anyway. Okay.

Please, mark the first commit as `Part of #276`, not `Fixes #276`. Just from the common sense: we cannot fix an issue several times in row.

Generally speaking a patch series should be marked this way (or something equivalent):

* Part of #xxx.
* Part of #xxx.
* Fixes #xxx.
* Follows up #xxx.
* Follows up #xxx.

I would mark the last commit as `Follows up #276`, because we come into the need of negative testing during work on the issue.

lib/tarantool_server.py
Totktonada: Okay.

lib/tarantool_server.py
Totktonada: So, the real timeout will be `timeout * 2`: each step can take `timeout` seconds. If we'll have more steps in a future, it'll be `timeout * N`, and, what is worse, it'll not obvious from the code.

In fact, with current code, we have even more than `timeout * 2` at max: the first loop in `seek_wait()` has no check for a deadline and the deadline is calculated after it.

lib/tarantool_server.py
Totktonada: It is definitely something toward the negative testing I want. But I don't see any test case. Can you add a test case for the issue of the topic? Otherwise I'll unable to experiment around the instrument and give you a feedback.

While we are here, several preliminary points:

1. Are we really need two runners: a shell script and pytest? If we use pytest here, I would start test-run.py from pytest directly (as a subprocess).
2. I'm under heavy impression that the approach with comparison of all output against a gold result leads to many problems. I would check the output for presence of an expected hunk (certain error message or so).

Totktonada: So, please, leave a note in the issue: we should revisit those time.sleep() -> gevent.sleep() changes in context of the described subprocess module problem.

lib/tarantool_server.py
```diff
@@ -886,6 +895,11 @@ def start(self, silent=True, wait=True, wait_load=True, rais=True, args=[],
                     color_stdout('\n[Instance "{0.name}"] Tarantool server '
                                  'failed to start\n'.format(self),
                                  schema='error')
+                    if err.timeout:
+                        color_stdout(
+                            '\n[Instance "{0.name}"] Tarantool server start '
+                            'timeout {1} was reached.\n'.format(
+                                self, err.timeout), schema='error')
```
Totktonada, line 902: Can we tweak the line above to reuse `TarantoolStartError.__str__()` and, so, don't duplicate the error reporting code?

lib/preprocessor.py
```diff
@@ -206,8 +207,14 @@ def server_start(self, ctype, sname, opts):
...
-            self.servers[sname].start(silent=True, rais=True, wait=wait,
-                                      wait_load=wait_load, args=args)
+            is_started = self.servers[sname].start(
+                    silent=True, rais=True, wait=wait,
+                    wait_load=wait_load, args=args)
+            if not is_started and wait and wait_load:
+                color_stdout('\n[Instance "{0}"] Failed to start tarantool '
+                             'instance\n'.format(sname),
+                             schema='error')
+                self.kill_current_test()
```
Totktonada, line 217: To be honest, I don't get, why we should do something aside catching an exception and re-raising it as `LuaPreprocessorException`. In my head, the plan was to raise TarantoolStartError, when we reach the timeout, and everything else should work.

ligurio:
@VitaliyaIoffe could you rebase to master? Latest changes includes code coverage measurement and it would be interesting how code coverage numbers are changed with your tests.

test/negative_tests/test_reconcile_logs.py
```diff
@@ -0,0 +1,17 @@
+import pytest
```
ligurio, line 1: What is a sense to place tests, helpers and static file to a separate directory? We already have a directory for unit tests - `unittest` and I think we can place your pytest tests there too. If you want to run negative tests separately we can use [pytest marks](https://docs.pytest.org/en/6.2.x/mark.html) for this.

ligurio (changes requested):

test/negative_tests/test_reconcile_logs.py
Totktonada: If we're about my initial idea, those are not unit tests: here we want to run test-run with a test, which is known to fail, and check that it fails as expected: look at the error message, verify that there are no orphan tarantool instances after test-run termination.

Totktonada: xfail is the different thing in my understanding: it is when a test should pass, but the code is not ready at the moment. It is the temporary mark, when we expect that the code will be fixed in a future.

My idea is different: add tests that will verify test-run behaviour, when a test under test-run fails. Here we have another success critetia: not just whether a test-run's test passed. But everything else is the same: correct test-run behaviour is not 'expected fail' — it is success.

lib/tarantool_server.py
VitaliyaIoffe: Got it

lib/tarantool_server.py
VitaliyaIoffe: You are right. It is unexpected behavior.

lib/tarantool_server.py
VitaliyaIoffe: Thank you for this reproduce way.

VitaliyaIoffe: I update a new variant without reconciliation between logs.  Now we wait for expected lines, which can explain the current state test test-run log.

lib/tarantool_server.py
VitaliyaIoffe: Fixed

lib/preprocessor.py
VitaliyaIoffe: Fixed

test/negative_tests/test_reconcile_logs.py
VitaliyaIoffe: I agree with @Totktonada. `xfail `mark is used in another way. A test can be failed in different situations, for example, unexpected exception, negative testing is a type of testing when we know the behavior of our product.

lib/tarantool_server.py
Totktonada: I'll reopen this discussion to remind about the first loop in `seek_wait()`:

```python
while True:
    if os.path.exists(self.path):
        break
    gevent.sleep(0.001)
```

There is no deadline check here.

lib/preprocessor.py
```diff
@@ -206,8 +206,9 @@ def server_start(self, ctype, sname, opts):
             if crash_expected:
                 # disable crash detector
                 self.servers[sname].crash_expected = True
-            self.servers[sname].start(silent=True, rais=True, wait=wait,
-                                      wait_load=wait_load, args=args)
+            self.servers[sname].start(
+                    silent=True, rais=True, wait=wait,
+                    wait_load=wait_load, args=args)
```
Totktonada, line 211: Nit: Unnecessary change.

lib/tarantool_server.py
Totktonada: I [did](https://github.com/tarantool/test-run/issues/252#issuecomment-855380882) it.

test/negative_tests/hang.result
```diff
@@ -0,0 +1,25 @@
+-- test-run result file version 2
+--  box.schema.user.grant('guest', 'replication')replication
+local test_run = require('test_run').new()
```
Totktonada, line 3: Commands here are different from the test itself. Anyway, is there any sense in placing a result file for this test? We expect a failure before comparing output with it.

Totktonada:
I would synchronize voicely, but I'll highlight several general points.

* We unable to add (non-disabled) negative test case into the `test/` directory. But nobody obligates us to do so. We can add another top level directory: say, `ntest/` (negative testing) or `ftest/` (failing tests).
* In fact, in the scope of this issue, I want just to keep those test cases and, ideally, add them together with the corresponding fixes. Just to illustrate, what was wrong. It is okay if I'll need to run them manually (like `./ntest/test-run.py foo/bar.test.lua`) and postpone all scripting around.
* We fixed two problems here (see the [discussion](https://github.com/tarantool/test-run/pull/302#discussion_r634745313) above), so there should be two test cases.
* There are xlogs (write ahead logs of the database), tarantool log files, test-run output. The TarantoolLog class is used to search over test-run output, it looks hacky. And I'm against of using test-run's code to test test-run. Especially, when we need to modify it for that.
* There are a lot of completely irrelevant Lua code and names around the new test case. `xlog.lua` — we don't test anything around xlogs, custom values for `memtx_memory`, `force_recovery`, `wal_max_size` and other box.cfg() options — defaults will work good for given test case. The same is applicable for `suite.ini` — there are options, which are not necessary here. (Mental experiment: if someone will ask you about arbitrary line of code you added or modified, will you able to explain why it should be written exactly how it is written now?)
* The new test case code spread across different files. So, if we use pytest, maybe, generate the content of the `ntest/` directory dynamically? Now it is just hard to understand how all those files are relevant to each other and what exactly we want to test.
* (However, if we'll not add the pytest harness here, several other questions will appear: the test case should be named properly (not just 'hang') and have a comment, what a negative situation it reproduces.)
* Unused `box.lua`.

I don't insist on adding the testing harness within this PR (however problems we fixing here look as good candidates to play with it). Either way you'll choose (with or without harness), I'll ask to arrange the testing code in some readable way. (Mental experiment: I'm just a stranger, which don't know much about tarantool and test-run. Will I understand idea of the testing code by looking here and there?)

lib/tarantool_server.py
```diff
@@ -833,12 +843,12 @@ def cleanup(self, *args, **kwargs):
     def start(self, silent=True, wait=True, wait_load=True, rais=True, args=[],
               **kwargs):
         if self._start_against_running:
-            return
+            return True
```
Totktonada, line 853: Is it some artifact from the old version of the patch? AFAIU, the return value is not used anywhere.

lib/tarantool_server.py
```diff
@@ -1109,12 +1138,9 @@ def wait_until_started(self, wait_load=True):
...
                   '(wait_load={})\n'.format(self.name, str(wait_load)),
                   schema='info')
-
         if wait_load:
-            msg = 'entering the event loop|will retry binding|hot standby mode'
-            p = self.process if not self.gdb and not self.lldb else None
-            self.logfile_pos.seek_wait(msg, p, self.name)
-        while True:
+            self.wait_load(deadline)
+        while time.time() < deadline:
```
avtikhon, line 1143: I've tried your patches and once got the following:
```
[036] replication-py/init_storage.test.py                             
[036] Test.run() received the following error:
[036] Traceback (most recent call last):
[036]   File "/home/ubuntu/tarantool/test-run/lib/test.py", line 190, in run
[036]     self.execute(server)
[036]   File "/home/ubuntu/tarantool/test-run/lib/tarantool_server.py", line 412, in execute
[036]     exec(code, new_globals)
[036]   File "replication-py/init_storage.test.py", line 76, in <module>
[036]     replica.wait_until_started()
[036]   File "/home/ubuntu/tarantool/test-run/lib/tarantool_server.py", line 1137, in wait_until_started
[036]     while time.time() < deadline:
[036] TypeError: '<' not supported between instances of 'float' and 'NoneType'
[036] 
[036] [ fail ]
[036] Test failed! Output from reject file var/rejects/replication-py/init_storage.reject:
[036] box.schema.user.grant('guest', 'replication')
```

avtikhon:
Luacheck fails with:
https://github.com/tarantool/tarantool/runs/2810348170#step:4:392
```
Checking test-run/test/negative_tests/hang.test.lua 5 warnings

    test-run/test/negative_tests/hang.test.lua:2:1: (W111) setting non-standard global variable test_run
    test-run/test/negative_tests/hang.test.lua:3:1: (W113) accessing undefined variable test_run
    test-run/test/negative_tests/hang.test.lua:4:1: (W113) accessing undefined variable test_run
    test-run/test/negative_tests/hang.test.lua:5:1: (W113) accessing undefined variable test_run
    test-run/test/negative_tests/hang.test.lua:6:1: (W113) accessing undefined variable test_run
```

lib/tarantool_server.py
```diff
@@ -928,6 +944,7 @@ def start(self, silent=True, wait=True, wait_load=True, rais=True, args=[],
                                                    actual_version),
                              schema='error')
                 raise TarantoolStartError(self.name)
+        return
```
Totktonada, line 947: Nit: Unnecessary change.

.luacheckrc
```diff
@@ -27,4 +27,5 @@ include_files = {
 exclude_files = {
     "lib/tarantool-python",
     "test/test-tarantool/*.test.lua",
+    "test/negative_tests/*.test.lua",
```
Totktonada, line 30: The 'negative_tests' directory now on the top level, right? And it is named 'negative_test' (without 's').

Totktonada:
CI is still red.

lib/preprocessor.py
Totktonada: It seems, it was fixed.

test/negative_tests/test_reconcile_logs.py
Totktonada: I'll close the discussion: it seems, there is nothing more to discuss. Sergey, if you have something else to say here, please, reopen.

lib/tarantool_server.py
Totktonada: It is still actual.

lib/options.py
```diff
@@ -211,6 +211,14 @@ def __init__(self):
                 help="""Break the test process with kill signal if the test runs
                 longer than this amount of seconds. Default: 110 [seconds].""")
 
+        parser.add_argument(
+            "--server-start-timeout",
+            dest="server_start_timeout",
+            default=env_int('SERVER_START_TIMEOUT', 90),
+            type=int,
+            help="""Break the test process with kill signal if the test starts
```
Totktonada, line 219: s/test/server/ (or instance)?

lib/tarantool_server.py
```diff
@@ -439,6 +439,7 @@ class TarantoolLog(object):
     def __init__(self, path):
         self.path = path
         self.log_begin = 0
+        self.last_position = 0
```
Totktonada, line 442: Those changes looks irrelevant.

lib/log_helper.py
```diff
@@ -0,0 +1,78 @@
+import os
```
Totktonada, line 1: Why don't place the test helper into the directory with tests?

negative_test/test_hanging_xlog.py
```diff
@@ -0,0 +1,39 @@
+from pathlib import Path
+
+from lib import Options
+from lib.log_helper import Log
+
+
+class TestHangingXLog:
```
Totktonada, line 7: There is nothing about xlog.

negative_test/test_hanging_xlog.py
```diff
@@ -0,0 +1,39 @@
...
+    def test_hanging_xlog(self, test_run_log: Log) -> None:
+        """Check if tarantool is hanging it will be stopped by start_timeout.
+
+        Check next patterns in the test-run log:
+        There are (timeout // 10 - 1) lines with 'No output' warnings, which
+        are going sequentially.
+        There are no (timeout // 10)th line with same warning.
+        Next, find log line with failing on start server timeout.
+        Finally, check log contains '[ fail ]'.
+        """
```
Totktonada, line 18: My expectation was that you'll describe what is going on in the test. Let's imagine I'm a stranger: what I see here? hang.test.lua starts replica.lua (which should use hang.test.lua as master). I asking myself: 'what is wrong here?', but there is no answer. Why the test should hang?

IOW, I want the story being persisted somewhere. Like so:

* hang.test.lua starts replica.lua.
* replica.lua calls box.cfg{replication = {<..hang.test.lua listen port..>}}.
* replica.lua unable to bootstrap, so test-run will never see 'entering the event loop' log entry.
* Before the fix: test-run spins in a kind of 'busy loop' around grepping tarantool's log and test timeout have no chance to occur.
* After the fix: test-run spins around grepping the log, but gently pass execution flow for other greenlets. So, test timeout will work as expected.
* After introducing start server timeout: we'll hit this new timeout before the test timeout, but everything else is the same.

lib/tarantool_server.py
Totktonada: How about adding this test case? It is about the {time -> gevent}.sleep() around ECONNREFUSE check.

test-run.py
```diff
@@ -94,6 +94,8 @@ def main_loop_parallel():
                  format(args.test_timeout), schema='tr_text')
     color_stdout("NO_OUTPUT_TIMEOUT:" . ljust(26) + "{}\n" .
                  format(args.no_output_timeout), schema='tr_text')
+    color_stdout("START_TIMEOUT:" . ljust(26) + "{}\n" .
+                 format(args.server_start_timeout), schema='tr_text')
```
Totktonada, line 98: I think, it would be better to report it as SERVER_START_TIMEOUT, not just START_TIMEOUT.

negative_test/test_hanging_xlog.py
Totktonada: In fact, any other way to describe what is going on is okay too. We should not miss this knowledge, that is my point.

lib/tarantool_server.py
ylobankov: Just resolving. Feel free to reopen.

lib/tarantool_server.py
ylobankov: Fixed.

lib/tarantool_server.py
ylobankov: Removed.

lib/log_helper.py
ylobankov: Not relevant anymore.

negative_test/test_hanging_xlog.py
ylobankov: Not relevant anymore.

negative_test/test_hanging_xlog.py
ylobankov: Not relevant anymore.

lib/tarantool_server.py
ylobankov: Fixed.

test/negative_tests/hang.result
ylobankov: Not relevant anymore.

lib/tarantool_server.py
ylobankov: Now fixed.

lib/tarantool_server.py
```diff
@@ -448,7 +448,7 @@ def seek_wait(self, msg, proc=None, name=None):
         while True:
             if os.path.exists(self.path):
                 break
-            time.sleep(0.001)
```
ligurio, line 451: Two locations with `time.sleep()` is left in `./lib/tarantool_server.py`. Why?

lib/tarantool_server.py
```diff
@@ -1016,16 +1016,6 @@ def stop(self, silent=True, signal=signal.SIGTERM):
             color_log('Server [%s] start against running ...\n',
                       schema='test_var')
             return
```
ligurio, line 1018: Could you cover a case by test?

ligurio (changes requested):

lib/options.py
```diff
@@ -213,6 +213,15 @@ def __init__(self):
                 Such files created by workers in the "var/reproduce" directory.
                 Note: The option works now only with parallel testing.""")
 
+        parser.add_argument(
```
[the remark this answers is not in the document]
ylobankov: Yes, we need it. `SERVER_START_TIMEOUT ` is used while starting a server, `TEST_TIMEOUT` is used while running a test.

lib/tarantool_server.py
ylobankov: If you mean a case covering the issue with hanging, I am going to do it in the next commit.

lib/tarantool_server.py
ylobankov: It looks like they were just missed.

ylobankov: Fixed.

lib/tarantool_server.py
ylobankov: Test is added.

lib/tarantool_server.py
```diff
@@ -1016,16 +1016,6 @@ def stop(self, silent=True, signal=signal.SIGTERM):
...
-        if self.status != 'started':
-            if not silent:
-                raise Exception('Server is not started')
-            else:
-                color_log(
-                    'Server [{0.name}] is not started '
-                    '(status:{0.status}) ...\n'.format(self),
-                    schema='test_var'
-                )
-            return
```
Totktonada, line 1019: Looks as a fix of (or tightly related to) #256.

ylobankov: Yep, looks like a fix. Mentioned in the commit message.

.github/workflows/test.yml
```diff
@@ -1,70 +1,69 @@
...
+        run: |
+          python -m pip install coveralls==3.*
+          make coverage
+
+      - name: upload coverage data to coveralls.io
+        if: ${{ matrix.python-version == '3.8' &&
+          matrix.tarantool-version == '2.7' }}
+        run: coveralls --service=github
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
ligurio, line 69: Why do you need to fix indentation here? Commit message describes changes as "refactoring", however changes includes adding spaces and changing indentation. I would fix only a problem `${{ matrix.python-version }} == 2.7` and revert other changes as useless.

.luacheckrc
```diff
@@ -27,6 +27,7 @@ include_files = {
 exclude_files = {
     "lib/tarantool-python",
     "test/test-tarantool/*.test.lua",
+    "test/unittest/*.test.lua",
```
ligurio, line 30: Why?

test/unittest/test_tarantool_server.py
```diff
@@ -0,0 +1,43 @@
+import os
```
ligurio, line 1: Please add a commit with fix of a problem covered by test to commit where test has been added or at least add a tag for  bug (`Follows up #256`).

test/unittest/box-cc0544b6afd1.lua
```diff
@@ -0,0 +1,12 @@
...
+
+box.cfg {
+    listen = os.getenv('LISTEN'),
+    memtx_memory = 107374182,
+    pid_file = 'tarantool.pid',
+    force_recovery = true,
+    wal_max_size = 500
+}
+
+require('console').listen(os.getenv('ADMIN'))
```
ligurio, line 12: Why do we need a console here?

test/unittest/box-cc0544b6afd1.lua
```diff
@@ -0,0 +1,12 @@
+#!/usr/bin/env tarantool
```
ligurio, line 1: Commit hash used in a filename looks non-existent.

test/unittest/suite.ini
```diff
@@ -0,0 +1,6 @@
+[default]
+core = tarantool
+description = ''
```
ligurio, line 3: Add a description or remove a field at all.

ligurio (changes requested):
Thanks for fixes!

See my comments inline and one more note: please put commit "Add unit test checking no hanging procs present" right after commit "Add timeout for starting tarantool server" with fix or squash them to a single commit.

lib/options.py
```diff
@@ -213,6 +213,15 @@ def __init__(self):
...
+            "--server-start-timeout",
+            dest="server_start_timeout",
+            default=env_int('SERVER_START_TIMEOUT', 90),
+            type=int,
+            help="""Break the server process with kill signal if the server
+            starts longer than this amount of seconds. Default: 90 [seconds]."""
```
Totktonada, line 222: An instance stopping process is a bit more complex than just sending SIGKILL.

Regular stopping of a default and a non-default instance is the following:

1. Send SIGTERM.
2. Start a watchdog, which will send SIGKILL after 5 seconds.
3. Wait termination of the process.
4. Stop the watchdog.

There is a process of stopping non-default servers, which remain alive after executing a test (disregarding whether it was succussful or failed). Here we use SIGKILL without any graceful period. See the `LuaTest.execute()` for advocation of this approach.

(If you think that it leads to problems I don't mind revisiting the approach if the problems it now solves will be taken into account. I guess the team will look at it in the scope of #318, but now this issue looks invalid or at least lacks necessary information.)

Back to the option description. I think that we should not clarify how exactly processes will be stopped, because it is beyond information, which is necessary to use the option.

Totktonada: Also a server process can failed to start at all: so there is no process to send a signal.

Totktonada: Important thing here is that we fail the test after this timeout if the server is unable to start (and if the server is not marked as expected to fail).

.luacheckrc
Totktonada: Those are `core = tarantool` tests (an input for a console). They poorly compatible with luacheck.

test/unittest/test_tarantool_server.py
```diff
@@ -0,0 +1,43 @@
...
+import subprocess
+import unittest
+
+
+class TestTarantoolServer(unittest.TestCase):
+    def test_tarantool_server_not_hanging(self):
+        env = os.environ.copy()
+        env['SERVER_START_TIMEOUT'] = '5'
+
+        cmd = ['python', 'test/test-run.py', 'unittest/hang.test.lua']
```
Totktonada, line 11: We use python3 in shebangs, because there are distros (say, CentOS 7), where python is alias for python2. Here we can use `sys.executable` to run test-run using exactly same interpreter as one, which runs the testing.

test/unittest/hang.test.lua
```diff
@@ -0,0 +1,6 @@
+test_run = require('test_run').new()
+
+test_run:cmd('create server replica with rpl_master=default, script="unittest/replica-7f4d4895ff58.lua"')
+test_run:cmd('start server replica')
+test_run:cmd('cleanup server replica')
+test_run:cmd('delete server replica')
```
Totktonada, line 6: Do you really think that it is obvious, why this code should fail?

test/unittest/replica-7f4d4895ff58.lua
```diff
@@ -0,0 +1,9 @@
+#!/usr/bin/env tarantool
+
+box.cfg {
+    listen = os.getenv('LISTEN'),
+    replication = os.getenv('MASTER'),
+    memtx_memory = 107374182
```
Totktonada, line 6: Default 256MiB should be okay. Don't clog box.cfg with options you don't need (here and in other instance files).

test/unittest/box-cc0544b6afd1.lua
Totktonada: Admin connection is used by test-run itself (say, to check `box.info.status` or for commands after `test_run:cmd("switch replica")`). Even if we got our expected fail without started console I would leave it as is. It is more or less standard instance file for test-run and it is harder to think about test-run's behaviour when it is non-standard.

lib/tarantool_server.py
Totktonada: Let's add the TCP binary port case to #274 and resolve it again.

test/unittest/box-cc0544b6afd1.lua
Totktonada: Non-existent?

I guess the hash is needed to find a process using `ps`. However, yep, it is not very obvious and it would be good to comment it here.

.github/workflows/test.yml
Totktonada: Formatting changes should not make it hard to read changes of a visible behaviour. Please, discard or move to its own commit.

.github/workflows/test.yml
```diff
@@ -1,70 +1,69 @@
 name: test-run
```
Totktonada, line 1: > previously, all jobs worked only against Python 2.7

I printed python version from test-run and [pushed](https://github.com/tarantool/test-run/commit/32aba8a686659193f816069beed9a7de41ab3861) it to trigger CI. The patch:

```diff
diff --git a/test-run.py b/test-run.py
index f706d4b..5e836be 100755
--- a/test-run.py
+++ b/test-run.py
@@ -291,6 +291,9 @@ if __name__ == "__main__":
 
     status = 0
 
+    color_stdout('sys.version: {}\n'.format(sys.version))
+    color_stdout('sys.executable: {}\n'.format(sys.executable))
+
     if Options().args.show_tags:
         show_tags()
         exit(status)
```

The output from the `build (2.7, 2.7)` job:

```
sys.version: 2.7.18 (default, Oct 29 2020, 08:03:56) 
[GCC 9.3.0]
sys.executable: /opt/hostedtoolcache/Python/2.7.18/x64/bin/python
```

The output from the `build (3.5, 2.7)` job:

```
sys.version: 3.5.10 (default, Nov 25 2020, 13:18:55) 
[GCC 9.3.0]
sys.executable: /opt/hostedtoolcache/Python/3.5.10/x64/bin/python
```

I see that that the step, which installs dependencies for Python 2, is not skipped on Python 3. However the statement that we test only Python 2 is not correct.

Do you want more strange things to look into?

* It seems, in python 3 jobs we install dependencies for python 2, but everything works. I guess the installed python bundle already have necessary dependencies.

Please, if you'll dig into all those things, fix them step by step, not everything in one commit. I hope it is not so hard work.

----

Wow, fresh gevent [supports](http://www.gevent.org/install.html#supported-platforms) latest python 2.7. I had a guess that out `requirements.txt` works only for Python 3, but it works for both. Nice.

Totktonada:
The patchset is mostly ok, but there are several things I want to be fixed. Please, acquire re-review after fixes.

Please, don't close comments opened by me (if I don't asked explicitly to do so after a fix), but respond to them. So next time I'll look over things that were important for me and will not re-review the entire patchset. It will save some time.

test/unittest/replica-7f4d4895ff58.lua
ylobankov: Fixed.

test/unittest/suite.ini
ylobankov: Fixed.

test/unittest/test_tarantool_server.py
ylobankov: Fixed.

test/unittest/box-cc0544b6afd1.lua
ylobankov: I added a comment about such names inside the test.

test/unittest/hang.test.lua
ylobankov: Fixed.

lib/tarantool_server.py
ylobankov: Added.

ylobankov:
> Thanks for fixes!
> 
> See my comments inline and one more note: please put commit "Add unit test checking no hanging procs present" right after commit "Add timeout for starting tarantool server" with fix or squash them to a single commit.

It is not possible. The test checks that a process of a non-started tarantool server will not hang and will be killed after server start timeout runs out. The problem with hanging processes of non-started tarantool servers was resolved in the `Kill processes of non-started tarantool servers` commit and the test can be added only after this commit.

.luacheckrc
ligurio: Anyway, I would fix luacheck warning with global in config than excluding a whole dir.

```
--- a/.luacheckrc
+++ b/.luacheckrc
@@ -1,4 +1,4 @@
-globals = {"box", "_TARANTOOL", "tonumber64", "os"}
+globals = {"box", "_TARANTOOL", "tonumber64", "os", "test_run"}
 ignore = {
     -- Accessing an undefined field of a global variable <debug>.
     "143/debug",
@@ -27,7 +27,6 @@ include_files = {
 exclude_files = {
     "lib/tarantool-python",
     "test/test-tarantool/*.test.lua",
-    "test/unittest/*.test.lua",
     "lib/luatest/**",
     "lib/checks/**",
 }
```

lib/options.py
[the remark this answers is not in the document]
ligurio: I'm ok. @Totktonada, are you satisfied by answer?

lib/tarantool_server.py
ligurio: Are you added it in a separate commit intentionally? Sometimes we add a fix and a test for it in a single commit.

test-run.py
```diff
@@ -91,6 +91,8 @@ def main_loop_parallel():
 
     color_stdout("Timeout options:\n", schema='tr_text')
     color_stdout('-' * 19, "\n",       schema='separator')
+    color_stdout("SERVER_START_TIMEOUT:" . ljust(26) + "{}\n" .
```
ligurio, line 94: Test for a new timeout is missed. Could you add it?

ligurio (changes requested):

test-run.py
ylobankov: Added unit test is about this timeout as well.

lib/tarantool_server.py
ylobankov: Yes. Actually, I don't see any issue that the test is in a separate commit. I have two commits (Add timeout for starting tarantool server and Kill processes of non-started tarantool servers) that fix two issues. And I have the commit with the test (Add unit test checking no hanging procs present) covering mentioned commits. And IMHO it is OK.

.luacheckrc
ylobankov: Ok, thanks! Fixed.

lib/options.py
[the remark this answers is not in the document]
Totktonada: I'm ok.

.github/workflows/test.yml
Totktonada: Now the PR has no github workflow changes, so I'll resolve this discussion.

Totktonada (approved):
LGTM.

test/unittest/box-cc0544b6afd1.lua
ligurio: > Non-existent?

Yep. I thought the idea was to add a hash with commit to filename, like we do with tests for tarantool (like `gh-6028-replica.lua`).

ligurio (approved):
LGTM

Issue #301, opened 2021-05-17
Wrong condition in wait_until_started


VitaliyaIoffe:
Have to refactor tarantool_server.py code. 

There is next in wait_until_started:
`             ....  

               elif ans in ('loading'): 

                    continue

                else:
                    raise Exception(
                        "Strange output for `box.info.status`: %s" % (ans)
                    ) 
               ....
`

If `ans = yaml.safe_load(temp.execute('box.info.status'))[0]` returns any substring of `loading` it should raise the exception. The condition must be `ans in ('loading',)` or `ans  == 'loading'`. 

It seems like a possible way for unexpected behavior.

PR #303, opened 2021-05-17
Fix wrong condition in wait_until_started

VitaliyaIoffe:
Was fixed check that the answer is equal to `'loading'`. There was a
wrong condition and any substring of `'loading'` seemed like a correct
and equal to it.

One element tuple in python is defined as `(element,)`, without
a comma is a string.

Closes: #301


avtikhon (approved):
Right, patch LGTM.

Totktonada (approved):
LGTM.

Issue #304, opened 2021-05-18
Consider integration with luatest


Totktonada:
[luatest](https://github.com/tarantool/luatest/) is the testing framework for tarantool written in Lua. The nice property of luatest is how tests are organized: a test case, setup code, teardown code are all in its own functions. Aside of this, there are a bunch of asserts and helpers, including helpers for spawning new tarantool instances.

The idea is to provide ability to write tests for tarantool using luatest API and run those tests together with existing ones. So we'll able to experiment around different ways to write a test and, if we'll dedice so, perform migration to luatest piece-by-piece.

The rough idea is to teach test-run to work as luatest harness: collect `*_test.lua` files and run them though luatest. test-run should report results for luatest tests in the same way as for all other ones.

There are other points, where we can integrate luatest and test-run. At least, we can reimplement server.lua to track started instances though test-run (so test-run will kill them at termination -- I don't know whether there is a problem with this in luatest, and will collect RSS usage -- after #277). It is not mandatory for an initial implementation.

kyukhin:
Let's refactor a couple of more or less complicated tests from the test suite. I'd consider something in `replication/` and `vinyl/`.

VitaliyaIoffe:
Implement parameterization in luatest tarantool/luatest#116

PR #310, opened 2021-06-15
Make it possible to execute luatest suites

VitaliyaIoffe:
1) Update paths for using luatest

Was added LUATEST_DIR and updated LUA_PATH for finding luatest binary
for execution `*_test.lua` inside the test-run.

Part of: #304 

2) Add luatest server

Provide luatest server, which can find tests by luatest pattern *_test.lua
and execute them `luatest -v <name>_test.lua -o tap --shuffle none`.

Was changed `get_filename_by_test` for right working not only with `*.test.*`,
but also with `*_test.*`.

Part of: #304


lib/test.py
```diff
@@ -94,7 +94,8 @@ def flush(self):
 
 
 def get_filename_by_test(postfix, test_name):
-    rg = re.compile(r'\.test.*')
+    """For <..>/<name>_test.* or <..>/<name>.test.* return <name> + postfix"""
```
Totktonada, line 97: I would just give a couple of examples:

- .result, foo/bar.test.lua -> bar.result
- .reject, bar_test.lua -> bar.reject

lib/luatest_server.py
```diff
@@ -0,0 +1,110 @@
+import glob
+import os
+import re
+import sys
+from subprocess import STDOUT
+from gevent.subprocess import Popen, PIPE
```
Totktonada, line 6: Why do you import Popen() and constants from different modules (subprocess and gevent.subprocess)?

lib/luatest_server.py
```diff
@@ -0,0 +1,110 @@
...
+        """
+        server.current_test = self
+        ts = TestState(self.suite_ini, None, TarantoolServer,
+                       self.run_params,
+                       default_server_no_connect=server)
+        self.inspector.set_parser(ts)
+        command = [os.environ['LUATEST_BIN'], '-v', '-c', self.name,
+                   '-o', 'tap',
+                   '--shuffle', 'none']
+        proc = Popen(command, cwd=server.vardir, stdout=PIPE, stderr=STDOUT)
```
Totktonada, line 41: I'm not very fun about mixing stdout in TAP13 format and stderr in arbitrary format. It is possible that the mix will anyway be parseable as TAP13 (or at least be accepted by given TAP13 parser). Say, if stderr is formatted in YAML, it may be interpreted as a [TAP13 YAML block](https://testanything.org/tap-version-13-specification.html#yaml-blocks). Depending of certain parser there may be other kinds of output that will be accepted. For example, our TAP13 parser [is not very strict](https://github.com/tarantool/test-run/blob/ab453e33960a02dc77e59cb71257486005d7891f/lib/pytap13.py#L167-L168) (`self.strict` is always `False`). AFAIR, I had intention to make the parser strictier, but our tests were not ready for that, so I hid this behaviour under the flag. It would be good to re-check, whether we can enable it now (out of this issue). There were fixes around incorrect test plans.

Back to mixing of stderr into stdout. In the past, AppTest had this behaviour, but I [changed](https://github.com/tarantool/test-run/commit/dac6b2b4640a7dd3d130ddde13bb4b25a730a4ad) it. I think it would be good to don't introduce it again for LuatestTest.

lib/luatest_server.py
```diff
@@ -0,0 +1,110 @@
...
+
+
+class LuatestTest(Test):
+    """ Handle *_test.lua.
+
+    Provide method for executing luatest <name>_test.lua test.
+    """
+
+    def __init__(self, *args, **kwargs):
+        Test.__init__(self, *args, **kwargs)
```
Totktonada, line 22: Nit: the rule of thumb is to use `super()`.

lib/luatest_server.py
```diff
@@ -0,0 +1,110 @@
...
+        Execute 'luatest -v <name>_test.lua -o tap --shuffle none'
+        Provide a verbose output in the tap format.
+        Use shuffle option in none mode for avoiding mixing tests.
+        Use capture mode.
+        """
+        server.current_test = self
+        ts = TestState(self.suite_ini, None, TarantoolServer,
+                       self.run_params,
+                       default_server_no_connect=server)
+        self.inspector.set_parser(ts)
```
Totktonada, line 37: I'm not sure that we want to use `require('test_run').new():cmd(<...>)` functions in tests written for luatest:

1. It'll add hard dependecy on test-run for tests and block transition to pure luatest (I don't know, whether we'll actually want it, but I want to have this ability).
2. If it is necessary and works, I want to see a test in the test-run repository.

If you found that it is necessary in our tests, please, point me an example.

lib/luatest_server.py
```diff
@@ -0,0 +1,110 @@
...
+                    answer.append(test)
+            return answer
+
+        test_suite.ini['suite'] = suite_path
+        tests = glob.glob(os.path.join(suite_path, "*_test.lua"))
+
+        if not tests:
+            executable_path_glob = os.path.join(test_suite.args.builddir,
+                                                'test', suite_path, '*_test')
+            tests = glob.glob(executable_path_glob)
```
Totktonada, line 101: I doubt that we should find tests in the binary dir. It is actual for unit tests, because we need to compile them. However it is not the case for Lua files.

lib/luatest_server.py
VitaliyaIoffe: As I remember, for now, I erased all test_run dependencies

lib/__init__.py
```diff
@@ -53,11 +53,16 @@ def module_init():
...
     os.environ["BUILDDIR"] = BUILDDIR
     soext = sys.platform == 'darwin' and 'dylib' or 'so'
-    os.environ["LUA_PATH"] = SOURCEDIR+"/?.lua;"+SOURCEDIR+"/?/init.lua;;"
-    os.environ["LUA_CPATH"] = BUILDDIR+"/?."+soext+";;"
+    os.environ["LUA_PATH"] = (LUATEST_DIR + "/?.lua;"
+                              + SOURCEDIR + "/test/?.lua;"
+                              + SOURCEDIR + "/?.lua;"
+                              + SOURCEDIR + "/?/init.lua;;")
+    os.environ['LUATEST_BIN'] = LUATEST_DIR + "/bin/luatest"
+    os.environ["LUA_CPATH"] = BUILDDIR + "/?." + soext + ";;"
```
Totktonada, line 66: To be honest, I completely missed the point of this commit. Please, explain which problems are solved here in the commit message.

I guess that one of relevant problems is where to find luatest executable. There are two cases:

1. luatest is installed into the system (rare, but possible; personally I use this variant).
2. luatest is installed using `tarantoolctl rocks install luatest` into `.rocks` directory within current directory (source, build, cwd?) or somewhere upward.

If we'll add `.rocks/bin` (or `../../<...>/../.rocks/bin`) into PATH when it is found, we'll support both cases. I implemented this approach [here](https://github.com/tarantool/conf/blob/e1da1194fd247405e909a76127e31e0866efc179/Makefile#L17-L33). I hope it'll be easier in Python comparing to GNU Make.

About LUA_PATH changes: I need a problem description to don't repeat your effort.

Totktonada: If luatest is installed by `tarantoolctl rocks install luatest` (case 2 above), `require('luatest')` should work without any LUA_PATH changes. But maybe I miss some cases: say, `--vardir` usage. I propose to describe your layout, so we can spot a problem and discuss how to better resolve it.

Totktonada:
I glanced over the code (without actual testing — will do on the next review iteration). Aside of comments above I would want to see a luatest test in the test-run repository.

I think this PR should close the issue: further acivities should go under separate tasks.

lib/luatest_server.py
VitaliyaIoffe: Have to change `-c` option in luatest
https://github.com/tarantool/luatest/issues/147

lib/__init__.py
VitaliyaIoffe: As we discussed, now the script is searching for the `.rocks` directory in source and build directories.

.github/workflows/test.yml
```diff
@@ -35,6 +35,9 @@ jobs:
        sudo apt update -y
        sudo apt-get -y install lua5.1 luarocks
        sudo luarocks install luacheck
+       cd ..
+       sudo tarantoolctl rocks install luatest
+       cd -
```
VitaliyaIoffe, line 40: Now we are searching for rocks in the source and build dirs. But for the workflow, both of them are equal to `/home/runner/work/test-run/`. Without this magic with paths `.rocks` would be installed to `home/runner/work/test-run/test-run`. Do we want to add conditions search in `source, ../source, source/*`?

lib/luatest_server.py
Totktonada: But the code in test-run that mixing stdout and stderr still there. It should not.

lib/__init__.py
Totktonada: It is not exactly what I wanted:

1. I think that it should be either build or source directory. Otherwise the behaviour will be less obvious.
2. It should traverse directories upward recursively till the root directory, just like tarantool does for rocks.
3. I would want to keep ability to use /usr/bin/luatest + /usr/share/tarantool/luatest.

Aside of this, as stated [here](https://github.com/tarantool/tarantool/pull/6181/files#r680340435), we'll need to bundle luatest code into tarantool's source tarball. The easiest way to do so is to add luatest as submodule to test-run.

After enabling the submodule there is no much point to search luatest somewhere else. So my points above look non actual.

If you don't mind, I'll propose a set of small related patches.

Totktonada: Ouch, I see now, it is recursive. Sorry.

lib/luatest_server.py
Totktonada: I'll leave it as is for now.

test/luatest_helpers.lua
```diff
@@ -0,0 +1,59 @@
...
+    local random = digest.urandom(9)
+    local id = digest.base64_encode(random, {urlsafe = true})
+    if name ~= nil then
+        name = 'tmp_workdir'
+    end
+    local workdir = ('%s/%s-%s'):format(vardir, name, id)
+    fio.mktree(workdir)
+
+    return workdir
+end
```
Totktonada, line 57: > It repeats server's initialization code, but slightly changes naming and misses rmdir. And not used anywhere.

https://github.com/tarantool/tarantool/pull/6181/files#r734163750

Totktonada: So I doubt that it'll land to tarantool without changes and would remove it from test-run (at least for now).

Totktonada: Removed.

Totktonada:
> I think this PR should close the issue: further acivities should go under separate tasks.

I changed it in the commit message and in the PR description.

PR #313, opened 2021-09-23
Fix false positive behavior of test-run.py in case of using '-j -1' opt

ylobankov:
The problem is that the test-run.py script returns exit code 0 in case of using
'-j -1' option even though some tests were failed during test run. This patch
fixes this false positive behavior by checking test status in a proper way.

### DIFF
--- a/test-run.py
+++ b/test-run.py
@@ -172,7 +172,10 @@ def main_loop_consistent(failed_test_ids):
         worker_id = 1
         worker = task_group['gen_worker'](worker_id)
         for task_id in task_ids:
-            short_status = worker.run_task(task_id)
+            # The 'run_task' method returns a tuple of three items:
+            # (short_status, result_checksum, duration). So taking the first
+            # item of this tuple for failure check.
+            short_status = worker.run_task(task_id)[0]
             if short_status == 'fail':
                 reproduce_file_path = \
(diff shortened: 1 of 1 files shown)


Totktonada:
NB: It is the regression from ec8c9911fbb15bdcf4d5e5f29f53bf12d6b3c0ba.

Totktonada:
We follow the 50/72 rule (more or less, at least we attempt to do so in most cases). See the [developer guidelines](https://www.tarantool.io/en/doc/latest/dev_guide/developer_guidelines/#how-to-write-a-commit-message). I polished it there to don't play ping-pong without a real need (I hope you don't mind).

Re-verified that the problem was exist and that it was fixed by the change.

Re-verified the new test-run version on full tarantool test suite, no regressions observed.

Totktonada (approved):
LGTM. I'll wait for CI and push.

Totktonada:
Updated the test-run submodule in tarantool in the following commits: [2.10.0-beta1-54-g9f208f0d3](https://github.com/tarantool/tarantool/commit/9f208f0d3dc32ac3941090670c23d95f230ebd1a), 2.8.2-16-g88458ba1c, 1.10.11-7-gdb5406eaa.

PR #315, opened 2021-11-03
checks: fix license authors

DifferentialOrange:
This patch checks out to up-to-date tarantool/checks master branch.
The only commit that has been added contains license authors fix.
See more in tarantool/checks#28.


Totktonada (approved):
Thanks!

Issue #316, opened 2022-01-14
Enable parallel run for fragile files


kyukhin:
We improve the testing of fragile files:
- [x] Looks like for now it is more or less safe to run in parallel
- [ ] Also, looks like `checksum`-s machinery is dead, let's remove it as well

PR #317, opened 2022-01-17
Enable parallel invocation of fragile tests

kyukhin:
This will speed up tarantool/tarantool testing about 3 times.


Totktonada:
NB: It is part of #316.

Totktonada (approved):
LGTM.

It left some dead code, but does the job.

Just in case I verified that a failed fragile test still restarted.

ylobankov (approved):
Straightforward enough.

PR #321, opened 2022-02-09
Drop 'checksum' machinery

### LINKED ISSUE #316: Enable parallel run for fragile files

ylobankov:
It looks like the 'checksum' machinery is dead, so let's remove it.

Closes #316

### DIFF
--- a/lib/test.py
+++ b/lib/test.py
@@ -7,7 +7,6 @@ import shutil
 import sys
 import traceback
 from functools import partial
-from hashlib import md5
 
 from lib import Options
 from lib.colorer import color_stdout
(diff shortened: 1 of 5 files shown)


Totktonada:
1. The comment states that we're going to drop checksums, but the code drops fragile lists at all. Whiel I support the second option, it should be cleanly described.
2. In either case we drop some existing functionality. Let's at least add `qa_notice()` call if test-run detects that a user has an ignored directive in the configuration.

ylobankov:
> 1. The comment states that we're going to drop checksums, but the code drops fragile lists at all. Whiel I support the second option, it should be cleanly described.
> 2. In either case we drop some existing functionality. Let's at least add `qa_notice()` call if test-run detects that a user has an ignored directive in the configuration.

You are right, I cut out a little bit more than it's needed :) Now fixed, thanks!

Totktonada (approved):
Looks generally good.

I glanced on the code to spot the idea and didn't performed a decent review and testing of the changes. If it is required in your team, please, find a second reviewer within your team. If not, feel free to go ahead and proceed.

I still think that the proper functionality deletion procedure should at least issue a QA warning if checksums are present. I'll not insist here.

Issue #122, opened 2018-09-03
Add command line option to disable reproducing trace


Korablev77:
Proposal is to add an option to be able to turn off reproducing trace: sometimes it consists of quite long list of tests and makes output unclear and complicated to read.

sergw:
Temporarily, this command can be used as a workaround:
```
find . -name suite.ini | xargs sed -i '/^show_reproduce_content =/{h;s/=.*/= False/};${x;/^$/{s//show_reproduce_content = False/;H};x}'
```

Totktonada:
Is any problem remains? We [implemented](https://github.com/tarantool/test-run/pull/116) `show_reproduce_content` option and [enabled](https://github.com/tarantool/tarantool/commit/e96cd176030b8c8282458b7bf9ef1817d05c0ccd) it for sql-tap test suite.

Totktonada:
Feel free to reopen if you still experience any problems regarding this kind of test-run's output.

PR #323, opened 2022-02-14
Update and refine the readme

NickVolynkin:
Preview: https://github.com/NickVolynkin/test-run/blob/update-readme/README.md

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,51 +2,185 @@
 
 [![Coverage Status](https://coveralls.io/repos/github/tarantool/test-run/badge.svg)](https://coveralls.io/github/tarantool/test-run)
 
-### Test Suite
 
-Bunch of tests, that lay down in the subfolder (recursively) with `suite.ini`
-file. `suite.ini` is basic ini-file, that consists of one section `default`,
-and a number of fields:
+<!-- START doctoc generated TOC please keep comment here to allow auto update -->
+<!-- DON'T EDIT THIS SECTION, INSTEAD RE-RUN doctoc TO UPDATE -->
+## Table of Contents
 
-* `core`
-* `description` - Test Suite description
-* `script` - shebang file to start tarantool with
-* disables:
-    * `disabled` - tests that must be skipped
-    * `release_disabled` - tests that must be skipped when Tarantool has been
-      builded with `Release`
-    * `valgrind_disabled` - tests that must be skipped when Valgrind is enabled
-* `lua_libs` - paths for lua files, that should be copied into the folder,
-  where server is started (delimited with the space, e.g. `lua_libs=lua/1.lua
-  lua/2.lua`)
-* `long_run` - mark tests as long, enabled only with `--long` option (delimited
-  with the space, e.g. `long_run=t1.test.lua t2.test.lua`)
-* `config` - test configuration file name
+- [Test suite configuration file](#test-suite-configuration-file)
+  - [General configuration values](#general-configuration-values)
+  - [Disabling and skipping tests](#disabling-and-skipping-tests)
+  - [Other parameters](#other-parameters)
+  - [Fragile tests](#fragile-tests)
+- [Test composition](#test-composition)
+- [Test execution](#test-execution)
+- [Test configuration](#test-configuration)
+- [Writing various types of tests](#writing-various-types-of-tests)
+  - [Python tests](#python-tests)
+  - [Lua](#lua)
+  - [SQL](#sql)
+- [Interaction with the test environment](#interaction-with-the-test-environment)
+- [pretest_clean()](#pretest_clean)
+- [Tags](#tags)
+- [Projects that use test-run](#projects-that-use-test-run)
 
-Field `core` must be one of:
+<!-- END doctoc generated TOC please keep comment here to allow auto update -->
 
-* `tarantool` - Test-Suite for Functional Testing
-* `app` - Another functional Test-Suite
-* `unittest` - Unit-Testing Test Suite
+## Test suite configuration file
 
-### Test
+Test suite is a bunch of tests located in a directory with a `suite.ini`
+file. The `suite.ini` is a basic ini-file with one section `[default]`.
 
-Each test consists of files `*.test(.lua|.sql|.py)?`, `*.result`, and may have
-skip condition file `*.skipcond`.  On first run (without `.result`) `.result`
-is generated from output.  Each run, in the beggining, `.skipcond` file is
-executed. In the local env there's object `self`, that's `Test` object. If test
-must be skipped - you must put `self.skip = 1` in this file. Next,
-`.test(.lua|.py)?` is executed and file `.reject` is created, then `.reject` is
-compared with `.result`. If something differs, then 15 last string of this diff
-file are printed and `.reject` file is saving in the `<vardir>/rejects/<suite>`
-subfolder given in options or set localy as `var/rejects/<suite>` by default.
-If not, then `.reject` file is deleted.
+```ini
+[default]
+core = luatest
+description = Example test suite using luatest
+...
+```
 
-### Test configuration
+There's a syntax for single- and multiline lists. (TODO see ./sql-tap/suite.ini)
+For example:
 
-Test configuration file contains config for multiple run. For each test section
-system runs separated test and compares result with common `.result` file. For
-example we need to run one test for different db engines("*" means default
+```ini
+single_line = first.test.lua second.test.lua third.test.lua
+multiline = first.test.lua ;
+            second.test.lua ;
+            third.test.lua ;
+            ...
+```
+
+Below is a list of configuration values (fields) in the `suite.ini` file.
+
+### General configuration values
+
+* `core` — major type of tests in this suite.
+  Should have one of the following values:
+
+  * `tarantool` — diff-test, a test that reads a file with commands, 
+    feeds it line by line to the tarantool console,
+    writes requests and responses to an output file,
+    and then compares that file with a reference file.
+  * `app` — a Lua script test.
+    Most of such tests produce
+    [TAP13 output](http://testanything.org/tap-version-13-specification.html)
+    using the built-in `tap` module.
+    Test-run validates such output without a reference file.
+    Some tests produce non-TAP13 output, which is compared to a reference file.
+  * `luatest` — test suite using the [luatest](https://github.com/tarantool/luatest/) library.
+  * `unittest` — an executable test file.
+
+* `description`
+
+* `script` — A file with Tarantool commands.
+  It is used to start the default server using `tarantoolctl`.
+  The value should be a file in the same directory as `suite.ini`, like `box.lua` or `master.lua`.
+  This setting is mandatory for suites with diff-tests (`core = tarantool`)
+  and ignored with other types of tests.
+  
+* `config` — name of a test configuration file, for example, `engine.cfg`.
+  For details, see the [test configuration](#test-configuration) section below.
+
+* `lua_libs` — paths to Lua files that should be copied to the directory,
+  where a server is started. For example:
+
+  ```ini
+  lua_libs = lua/require_mod.lua lua/serializer_test.lua lua/process_timeout.lua
+  ```
+
+### Disabling and skipping tests
+
+A number of fields are used to disable certain tests:
+
+* `disabled` — list of tests that should be skipped.
+
+* `release_disabled` — list of tests that should only run when Tarantool is
+   built in the debug mode (with `-DCMAKE_BUILD_TYPE=Debug` and not `=RelWithDebInfo`).
+   Tests that use error injections, which are only available in debug builds,
+   should always be `release_disabled`.
+
+* `valgrind_disabled` — list of tests that should be skipped when Valgrind is enabled
+
+* `long_run` — mark tests as long. Such tests will run only with `--long` option.
+
+  ```ini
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -2,51 +2,126 @@
...
+```
+
+There's a syntax for single- and multiline lists. (TODO see ./sql-tap/suite.ini)
+For example:
+
+```ini
+single_line = first.test.lua second.test.lua third.test.lua
+multiline = first.test.lua ;
+            second.test.lua ;
+            third.test.lua ;
```
Totktonada, line 24: AFAIR, semicolons are just comments. They don't affect parsing in this case.

We investigated the questions around precise syntax (multiline options, inline comments) within the patchset for support of Python 3. I can find necessary info, if you want.

README.md
```diff
@@ -2,51 +2,126 @@
...
+
+### General configuration
+
+* `core` — major testing dependency or method
+  Should have one of the following values:
+
+  * `tarantool` — test suite used for functional testing
+  * `app` — TAP-tests, another functional test suite
+  * `unittest` — unit testing test suite
+  * `luatest` — test suite using luatest library
```
Totktonada, line 38: If you're interested I have a description of all test types [here](https://github.com/tarantool/luatest/issues/191) (under the cut). It reveals one or two main characteristics of a test type (taking care to real usage and informal terminology we have). However it is balanced to be as concise as possible. As for me, current description gives no useful information (sorry).

README.md
```diff
@@ -2,51 +2,126 @@
...
+
+* `description` — test suite description
```
Totktonada, line 40: AFAIR we have a rule in our developer guidelines to don't write useless comments (which just repeats a variable name, for example). If you want to give a suggestion what to write here (just slug, a few words sumary, verbose description), let's give it. Otherwise it is better to don't comment at all.

README.md
```diff
@@ -2,51 +2,126 @@
...
+* `script` (optional) — shebang file to start tarantool with.
+  Most often, the value is a file in the same directory as `suite.ini`, like `box.lua` or `master.lua`.
```
Totktonada, line 42: First of all, it is only for `core = tarantool` test suites. It is not optional for them and ignored for all other ones. 'Optional' word is barely applicable here.

'Most often'? I don't know exceptions.

I don't like 'shebang file' term, it sounds a bit strange. How about just `an instance file to start the default server from it (using tarantoolctl); applicable for `core = tarantool` test suites'.

README.md
```diff
@@ -2,51 +2,126 @@
...
+  
+* `config` — name of a test configuration file, for example, `engine.cfg`.
+  TODO: what's a test configuration file?
```
Totktonada, line 45: Parametrization. Ability to run a test in several configurations. Say, for memtx and vinyl.

README.md
```diff
@@ -2,51 +2,126 @@
...
-executed. In the local env there's object `self`, that's `Test` object. If test
-must be skipped - you must put `self.skip = 1` in this file. Next,
-`.test(.lua|.py)?` is executed and file `.reject` is created, then `.reject` is
-compared with `.result`. If something differs, then 15 last string of this diff
-file are printed and `.reject` file is saving in the `<vardir>/rejects/<suite>`
-subfolder given in options or set localy as `var/rejects/<suite>` by default.
-If not, then `.reject` file is deleted.
+## Test suite configuration
+
+Test suite is a bunch of tests located in a subdirectory (recursively) with a `suite.ini`
```
Totktonada, line 7: TBH, the recursive search was never used (at least now at last four years) and I doubt that it works (even if it initially did). I would not mention that here.

README.md
```diff
@@ -2,51 +2,126 @@
...
+  TODO: what's a test configuration file?
+
+* `lua_libs` — paths to Lua files that should be copied to the directory,
+  where a server is started. For example:
+
+  ```ini
+  lua_libs = lua/require_mod.lua lua/serializer_test.lua lua/process_timeout.lua
+  ```
+
+### Skipping tests
```
Totktonada, line 54: I suggest to avoid using two terms for the same entity. It is confusing per se, but also it clashes with skipping of tests using `*.skipcond` files. Disabled and skipped tests are reported differently by test-run. (We can argue, whether it is good or bad, but it is how test-run present information of the screen. So readme should reflect it to don't introduce a confusion.)

README.md
```diff
@@ -2,51 +2,126 @@
...
+
+A number of fields are used to disable (skip) certain tests:
+
+* `disabled` — list of tests that should be skipped
+* `release_disabled` — list of tests that should be skipped when Tarantool is
```
Totktonada, line 59: Worth to add: usually tests, which use error injections (which are available only in debug tarantool builds).

README.md
```diff
@@ -2,51 +2,126 @@
...
+    built with `Release`
```
Totktonada, line 60: Strictly speaking, tarantool is built either with `Debug` or with `RelWithDebInfo` build type.

README.md
```diff
@@ -2,51 +2,126 @@
...
+* `valgrind_disabled` — list of tests that should be skipped when Valgrind is enabled
+* `long_run` — mark tests as long, such tests will run only with `--long` option.
```
Totktonada, line 62: It would be good to give a suggestions, what is considered 'short' and what 'long'. A long time ago the following informal rule was spoken: a test should usually pass within 1 second. Now it is far away of truth. Maybe not a question for this PR.

README.md
```diff
@@ -2,51 +2,126 @@
...
+
+  ```ini
+  long_run = t1.test.lua t2.test.lua
+  ```
+
+### Other parameters
+
+
+* `show_reproduce_content` (optional, `True` by default) — TODO: what's this?
```
Totktonada, line 71: https://github.com/tarantool/test-run/issues/122

README.md
```diff
@@ -2,51 +2,126 @@
...
+* `is_parallel` (optional, `False` by default) — whether the tests in the suite can run in parallel
+
+* `use_unix_sockets` (optional, `False` by default) — use hard-coded UNIX sockets
```
Totktonada, line 74: For console, AFAIR.

'Hard-coded' is a bit too hard words. It depends of server's script name.

README.md
```diff
@@ -2,51 +2,126 @@
...
+and an optional skip condition file `*.skipcond`.
+
+In the beginning of each test run, the `.skipcond` file is executed as a Python script.
+In the local Python environment there's a `self` object, which is an instance of the [`Test` class](./lib/test.py).
+To skip the test under some condition, set `self.skip = 1`.
+
+Next, the test in `.test(.lua|.py)?` is executed. 
+If this is a first test run and there's no `.result` file yet, it is generated from the output.
+If there is a `.result` file, output is written to a `.reject` file,
+which is then compared to `.result`.
```
Totktonada, line 110: 1. `.test.sql` as well.
2. And `_test.lua` in luatest test suites.
3. The autowritting `.result` file was removed together with introducing the `--update-result` option.

README.md
```diff
@@ -2,51 +2,126 @@
...
+
+If there's a difference between `.reject` and `.result`, the following happens:
+
+1. The last 15 lines of diff are printed to output.
+2. The `.reject` file is saved in the `<vardir>/rejects/<suite>`
+   subfolder given in options or set locally as `var/rejects/<suite>` by default.
+
+If there is no difference between `.reject` and `.result`, the `.reject` file is deleted.
```
Totktonada, line 118: Don't know, whether it is useful, to be honest.

README.md
Totktonada: 4. There is no a word about TAP13 compatible output (it does not need a result file).

README.md
NickVolynkin: thanks, I've added this info to the document

README.md
```diff
@@ -2,51 +2,185 @@
...
+}
+```
+
+## Test composition
+
+Each test consists of the following files:
+
+* Test file: `<name>.test.lua`, `<name>_test.lua`, `<name>.test.py`, or `<name>.test.sql`.
+* Reference file: `<name>.result`.
+* Skip condition file: `<name>.skipcond`.
```
Totktonada, line 150: The altter two bullets are optional, but the text looks like they're mandatory.

README.md
```diff
@@ -62,15 +196,15 @@ configuration):
 }
 ```
 
-In test case we can get configuration from inspector:
+In the test case we can get configuration from the inspector:
 
 ```lua
 engine = test_run:get_cfg('engine')
 -- first run engine is 'memtx'
 -- second run engine is 'sophia'
```
Totktonada, line 204: Let's replace sophia with vinyl.

README.md
```diff
@@ -62,15 +196,15 @@ configuration):
...
 ```
 
-"engine" value has a special meaning for *.test.sql files: if it is "memtx" or
+"engine" value has a special meaning for `*.test.sql` files: if it is "memtx" or
 "vinyl", then the corresponding default engine will be set before executing
 commands from a test file. An engine is set with the following commands:
```
Totktonada, line 209: Regarding commands below: maybe it is too deep details for a user guide.

README.md
```diff
@@ -79,11 +213,13 @@ UPDATE "_session_settings" SET "value" = 'memtx|vinyl' WHERE "name" = 'sql_defau
...
+
+## Writing various types of tests
 
-#### Python
+### Python tests
 
-Files: `<name>.test.py`, `<name>.result` and `<name>.skipcond`(optionaly).
+Files: `<name>.test.py`, `<name>.result` and `<name>.skipcond` (optionally).
 
 Environment:
```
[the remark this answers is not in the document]
Totktonada: While `admin` below still actual.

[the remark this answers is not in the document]
Totktonada: I also suggest to re-verify that the example below still work.

README.md
```diff
@@ -166,14 +302,14 @@ box.info.lsn
 ...
 ```
 
-#### Lua
+### Lua
 
 Files: `<name>.test.lua`, `<name>.result` and `<name>.skipcond`(optionaly).
 Tests interact only with `AdminConnection`. Supports some preprocessor functions (eg `delimiter`)
```
Totktonada, line 308: > Tests interact only with `AdminConnection`.

Don't know what it means, but it looks irrelevant now.

README.md
```diff
@@ -166,14 +302,14 @@ box.info.lsn
 ...
 ```
 
-#### Lua
+### Lua
```
Totktonada, line 305: `core = tarantool` and `core = app` tests are really different. It worth to split its description.

README.md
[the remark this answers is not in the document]
NickVolynkin: I don't want to touch this part of readme in this PR. Will review it later.

README.md
```diff
@@ -2,51 +2,185 @@
...
+Whenever a test fails, the `.reject` file is saved in the `<vardir>/rejects/<suite>`
+subdirectory given in options or set locally as `var/rejects/<suite>` by default.
+If the test is considered successful, the `.reject` file is deleted.
+
+## Test configuration
+
+Test configuration file contains configuration for multiple runs.
+For each test section, system runs a separate test and compares the result to the common `.result` file.
+For  example, we need to run a test with different DB engines (`"*"` means the default
 configuration):
```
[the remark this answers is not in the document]
NickVolynkin: In this example

```json
{
    "first.test.lua": {
      "memtx": {"engine": "memtx"}
    },
    "second.test.lua": {},
    "*": {
        "memtx": {"engine": "memtx"},
        "vinyl": {"engine": "vinyl"}
    }
}
```

what's the difference between first and second tests?

Totktonada:
My comments are mostly not answered. It is in the responsibility of a developer to update all discussions before asking for a next review. Otherwise a reviewer will be obligated to re-check every point, which (s)he highlighted. It is time consuming and not profitable.

I see a temporary commit in this pull request. I expect ready-to-go (in your opinion) pull request, when a review is acquired. Please, elaborate.

After this I'll look at your answers and, if nothing will confuse me, can skip re-check of the changes. You'll wait less time for review so.

README.md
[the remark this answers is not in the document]
Totktonada: Both are run once, but the first on will receive `'memtx'` from a `test_run:get_cfg('engine')` call. A test can be writtten in any way, the difference is mostly how the configuration information is presented for a reader.

README.md
NickVolynkin: removed this part

README.md
NickVolynkin: I've found a single use case: `./sql-tap/suite.ini:31:show_reproduce_content = False`. So, for all other tests this value is `True` (by default), which is useless — is that really so?

README.md
NickVolynkin: I've removed enumeration of test file extensions from this line, it resolves (1) and (2).  Also, I won't mention writing `.result` files at all. Moved it to a separate issue: https://github.com/tarantool/test-run/issues/327

Added a few words about TAP13

README.md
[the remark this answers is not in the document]
Totktonada: How I'll differentiate parts, where problems are left deliberately and where it is unintentional? No, I think your point does not work here: your changes are not so dense.

Totktonada:
> My comments are mostly not answered.

Still. Please, find another reviewer.

Totktonada:
Seems like nobody is eager to work on it.

PR #324, opened 2022-02-18
Refactor luatest test related stuff

ylobankov:
- Remove test/luatest_helpers.lua
- Move `default_cfg`, `env_cfg`, `box_cfg` functions from
  test/luatest_helpers.lua to test/instances/default.lua
- Move `instance_uri` function from test/luatest_helpers.lua
  to test/luatest_helpers/server.lua and rename it into
  `build_instance_uri`
- Break down one test into four smaller ones


Totktonada:
> Move test/instances/default.lua to test/test-luatest/instances/default.lua

A project may contain several `core = luatest` test suites. The default instance should work for all.

> Move test/luatest_helpers/server.lua to test/test-luatest/helpers/server.lua
> with small refactoring

The same here. Those helpers are not for just one test suite.

----

The idea to place those helpers into the testing code of test-run had the following motivation:

* Speed up initial development. Not relevant anymore.
* Provide an example how to setup luatest based testing in a project that already uses test-run.
* Verify `luatest_server.py` in autotesting and preferably do this similarly to how we actually want to use luatest in tarantool.

So I had idea to keep layout here the same as in tarantool. At least to don't give a non-viable example.

Since we anyway think on moving the helpers to luatest itself, all those helpers should gone from tarantool and test-run. Until that I would keep the layout the same.

The alternative is to re-implement the test as pure luatest one. However I would not go this way, because of 1) less powerful testing (or a lot of boilerplate code) and 2) future moving the helpers to luatest.

ylobankov:
> So I had idea to keep layout here the same as in tarantool. At least to don't give a non-viable example.

Done.

test/test-luatest/smoke_check_test.lua
```diff
@@ -1,10 +1,11 @@
 local t = require('luatest')
-local g = t.group()
 
-local luatest_helpers = require('test.luatest_helpers')
+local server = require('test.luatest_helpers.server')
```
Totktonada, line 3: I'm not a fun of OOP on Lua, but luatest name classes (modules that acts like classes) from the capital letter and I would keep the naming style uniform.

PR #326, opened 2022-02-19
Add an option to show test-run's environment variables

### LINKED ISSUE #325: Option to show environment variables set by test-run
It would be useful to source in a shell and use the `luatest` executable, which is provided by test-run, directly.

It is convenient to run `luatest` directly to pass additional options to it. Most often `-p` to run a particular test case.

Totktonada:
Usage:

```shell
$ . <(./test/test-run.py --env)
```

(Or copy-paste `./test/test-run.py --env` output into a shell.)

test-run ships luatest sources, but it is tricky to use it directly, because it requires to set appropriate `LUA_PATH`.

The new option prints all environment variables, which test-run sets at initialization, including `PATH` and `LUA_PATH`. The variables are printed in the format suitable for sourcing into a shell.

So now it is easy to call something like `luatest -v -p test_case_name`.

Fixes #325.

### DIFF
--- a/lib/options.py
+++ b/lib/options.py
@@ -297,6 +297,21 @@ class Options(object):
                 last position), test-run will show a list of tags
                 and stop.
                 """)
+        parser.add_argument(
+                '--env',
+                dest='show_env',
+                action='store_true',
+                default=False,
+                help="""Print environment variables, which are set by test-run.
+
+                Useful for, say, running built-in luatest executable.
+
+                Usage: just source it into a current environment:
+
+                ```
+                . <(./test/test-run.py --env)
+                ```
+                """)
 
         # XXX: We can use parser.parse_intermixed_args() on
         # Python 3.7 to understand commands like
--- a/lib/utils.py
+++ b/lib/utils.py
@@ -410,3 +410,12 @@ def find_tags(filename):
             else:
                 break
     return tags
+
+
+def prepend_path(p):
+    """ Add an absolute path into PATH (at start) if it is not already there.
+    """
+    p = os.path.abspath(p)
+    if p in os.environ['PATH'].split(os.pathsep):
+        return
+    os.environ['PATH'] = os.pathsep.join((p, os.environ['PATH']))
--- a/lib/utils.py
+++ b/lib/utils.py
@@ -11,12 +11,20 @@ import json
 import subprocess
 from gevent import socket
 from lib.colorer import color_stdout
+
 try:
     # Python3.5 or above
     from signal import Signals
 except ImportError:
     # Python2
     Signals = None
+
+try:
+    # Python 3.3+.
+    from shlex import quote as _shlex_quote
+except ImportError:
+    # Python 2.7.
+    from pipes import quote as _shlex_quote
 
 
 UNIX_SOCKET_LEN_LIMIT = 107
@@ -419,3 +427,7 @@ def prepend_path(p):
     if p in os.environ['PATH'].split(os.pathsep):
         return
     os.environ['PATH'] = os.pathsep.join((p, os.environ['PATH']))
+
+
+def shlex_quote(s):
+    return _shlex_quote(s)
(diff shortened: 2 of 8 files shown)


Totktonada:
Update: added a recipe how to run luatest directly into README.md.

lib/utils.py
```diff
@@ -410,3 +410,12 @@ def find_tags(filename):
             else:
                 break
     return tags
+
+
+def prepend_path(p):
+    """ Add an absolute path into PATH (at start) if it is not already there.
+    """
+    p = os.path.abspath(p)
+    if p in os.environ['PATH'].split(os.pathsep):
```
ylobankov, line 427: Looks like we should check that the provided path is the first item on the list. If I had the `PATH` variable already modified manually, say, I added `/path/to/test-run/lib/luatest/bin` after `/usr/bin`, then in a few days I installed luatest into the system (now it is in `/usr/bin`), this wouldn’t work as expected.

Totktonada: I need the following information from you to ensure that we understood each other.

1. Whether you mean the case of testing of tarantool (so locally built tarantool will be added to the PATH) or testing of a module (so system tarantool will be used)?
2. Whether you mean using `./test/test-run.py`, using `./test/test-run.py --env` + `luatest` or both?
3. What is expected result and what is actual result in given conditions?

ylobankov: Imagine the situation that I want to use `test-run` to run some tests via its `luatest` executable. I don't have `luatest` installed in my system (`luatest` executable is not present in `/usr/bin/`) and I want to use just filename `luatest` without specifying full executable path. To do it I change my `PATH` so that `PATH=/usr/bin:/path/to/test-run/luatest:<…>`. As you can see `/usr/bin` goes first. So I use `luatest` as usual and everything is fine. 

Later for some reason I install `luatest` into my system (now `luatest` is in `/usr/bin/`). Thus `luatest` from `/usr/bin/` is used for my purposes and this is fine to me.

Then I want to run `test-run` to execute some luatest tests. When I do it, the `luatest` executable form `/usr/bin/` will be used because `/usr/bin` goes first in `PATH`. You just check that test-run's luatest executable path is on the list and return. So it is not enough.

lib/options.py
```diff
@@ -297,6 +297,21 @@ def __init__(self):
                 last position), test-run will show a list of tags
                 and stop.
                 """)
+        parser.add_argument(
+                '--env',
+                dest='show_env',
+                action='store_true',
+                default=False,
+                help="""Print environment variables, which are set by test-run.
+
```
ylobankov, line 306: Please delete all empty lines, it is useless. It will be one line anyway.

lib/options.py
```diff
@@ -297,6 +297,21 @@ def __init__(self):
...
+                Useful for, say, running built-in luatest executable.
+
+                Usage: just source it into a current environment:
+
+                ```
```
ylobankov, line 311: Three backticks will look bad because it will be one line. Please use one backtick.

lib/utils.py
ylobankov: Actually, this patch will be very helpful for users who are going to use built-in test-run's luatest. So I am not going to block this PR and ask solve my case. Let's resolve this in the scope of a separate issue.

ylobankov (approved):
LGTM, please fix help message.

lib/options.py
Totktonada: In my opinion argparse mangle my nicely formatted help messages :)

I found a solution. Please, check it out: PR #337.

Totktonada:
Rebased on top of current master.

lib/options.py
ylobankov: Ok, let's proceed in #337.

lib/options.py
ylobankov: Ok, let's proceed in https://github.com/tarantool/test-run/pull/337.

PR #329, opened 2022-03-15
Rerun all failed tests, not only marked as fragile

### LINKED ISSUE #328: Retry tests on failure
When a test fails, retry it up to N times, expecting a successful run. Effectively, this means treating all tests as flaky (fragile).

Report a retried test in the logs. Use different log messages for tests in the "fragile" list and not in the list.

ylobankov:
test-run supports functionality to rerun failed tests in place, but
these tests have to be on so called fragile list. To add a test to the
fragile list we need to add a special configuration to the suite.ini
file of a test suite. Configuration example:

    fragile = {
        "retries": 5,
        "tests": {
            "tarantoolctl.test.lua": {
                "issues": [ "gh-5059", "gh-5346" ]
            },
            "debug.test.lua": {
                "issues": [ "gh-5346" ]
            },
            ...
        }
    }

Rerunning failed tests in place is quite convenient because it allows us
to avoid rerunning all tests again and thus save time.

But to make it work as expected we should keep the list of fragile tests
always up-to-date. Flaky tests may be introduced every day and keeping
the list of fragile tests always up-to-date becomes extremely difficult
to do.

So our solusion is quite simple: just rerun all failed tests.
By default, the number of retries for regular and fragile tests is 3.
But for fragile tests this number can be overriden in the suite.ini
file.

Closes #328

### DIFF
--- a/lib/test_suite.py
+++ b/lib/test_suite.py
@@ -45,6 +45,9 @@ class TestSuite:
     server for this suite, the client program to execute individual
     tests and other suite properties. The server is started once per
     suite."""
+
+    RETRIES_COUNT = 3
+
     def get_multirun_conf(self, suite_path):
         conf_name = self.ini.get('config', None)
         if conf_name is None:
@@ -91,7 +94,7 @@ class TestSuite:
         self.args = args
         self.tests = []
         self.ini = {}
-        self.fragile = {'retries': 0, 'tests': {}}
+        self.fragile = {'retries': self.RETRIES_COUNT, 'tests': {}}
         self.suite_path = suite_path
         self.ini["core"] = "tarantool"
 
@@ -128,7 +131,7 @@ class TestSuite:
         if config.has_option("default", "fragile"):
             fragiles = config.get("default", "fragile")
             try:
-                self.fragile = json.loads(fragiles)
+                self.fragile.update(json.loads(fragiles))
                 if 'tests' not in self.fragile:
                     raise RuntimeError(
                         "Key 'tests' absent in 'fragile' json: {}"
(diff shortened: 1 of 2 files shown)


lib/test_suite.py
```diff
@@ -45,6 +45,9 @@ class TestSuite:
     server for this suite, the client program to execute individual
     tests and other suite properties. The server is started once per
     suite."""
+
+    DEFAULT_TEST_RETRIES_COUNT = 3
```
NickVolynkin, line 49: Can it be useful to override this with an env variable?

artembo (approved):
LGTM, but upvoting NickVolynkin question

lib/test_suite.py
ylobankov: I think yes but if we added a special option to test-run (--retry-count). IMHO, at this moment the option is redundant.

lib/test_suite.py
```diff
@@ -45,6 +45,9 @@ class TestSuite:
     server for this suite, the client program to execute individual
     tests and other suite properties. The server is started once per
     suite."""
+
+    RETRIES_COUNT = 3
```
Totktonada, line 49: It means 4 runs (3 retries). Is it expected or you supposed to set the overall amount of runs to 3?

Totktonada (approved):
LGTM visually.

lib/test_suite.py
Totktonada: I would enter to Yaroslav here: I think that it is premature at least and doubt that we'll need it at all.

NickVolynkin (approved):

lib/test_suite.py
ylobankov: It is expected. 4 runs (3 retries).

PR #331, opened 2022-03-16
Implement ability to use lua_libs variable in luatests.

EvgenyMekhanik:
For all kind of tests lua_libs variable in suite.ini file
used to determine custom lua modules used for this tests.
This patch implement ability to use this variable for luatests
also.

### DIFF
--- a/lib/luatest_server.py
+++ b/lib/luatest_server.py
@@ -1,6 +1,8 @@
+import errno
 import glob
 import os
 import re
+import shutil
 import sys
 
 from subprocess import Popen, PIPE
@@ -71,6 +73,7 @@ class LuatestServer(Server):
         self.testdir = os.path.abspath(os.curdir)
         self.vardir = ini['vardir']
         self.builddir = ini['builddir']
+        self.lua_libs = ini['lua_libs']
         self.name = 'luatest_server'
 
     @property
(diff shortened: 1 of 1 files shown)


lib/luatest_server.py
```diff
@@ -71,6 +73,7 @@ def __init__(self, _ini=None, test_suite=None):
         self.testdir = os.path.abspath(os.curdir)
         self.vardir = ini['vardir']
         self.builddir = ini['builddir']
+        self.lua_libs = ini['lua_libs']
```
Totktonada, line 76: Am I understand right that we'll miss ability to run the test using `luatest` directly?

EvgenyMekhanik: From tarantool test directory we always run luatests using test-run.py (as i know). I add this option for this case. If someone want to run tests which use custom lua modules using luatest directly he should install this modules in the directory where `require` can find it.

lib/luatest_server.py
```diff
@@ -85,6 +88,21 @@ def deploy(self, vardir=None, silent=True, wait=True):
...
+                source = os.path.join(self.testdir, i)
+                try:
+                    if os.path.isdir(source):
+                        shutil.copytree(source,
+                                        os.path.join(self.vardir,
+                                                     os.path.basename(source)))
+                    else:
+                        shutil.copy(source, self.vardir)
+                except IOError as e:
+                    if (e.errno == errno.ENOENT):
```
ylobankov, line 102: Parentheses can be removed.

lib/luatest_server.py
```diff
@@ -85,6 +88,21 @@ def deploy(self, vardir=None, silent=True, wait=True):
...
+                        continue
+                    raise
+            os.environ["LUA_PATH"] = self.vardir + '/?.lua;' + os.environ.get("LUA_PATH", ";;")
```
ylobankov, line 105: If `LUA_PATH` is not defined, you will get `;;;` at the end of the path. Looks like it should be `os.environ.get("LUA_PATH", ";")`.

lib/luatest_server.py
EvgenyMekhanik: fixed

lib/luatest_server.py
EvgenyMekhanik: fixed

lib/luatest_server.py
Totktonada: > From tarantool test directory we always run luatests using test-run.py (as i know).

I often run `luatest -v -p foo` directly to run a particular use case. Another reason to keep tests luatest compatible is motivation behind implementing luatest support in test-run: sooner or later we want to fully shift to luatest.

> If someone want to run tests which use custom lua modules using luatest directly he should install this modules in the directory where `require` can find it.

Just place the Lua file somewhere in `test/` and require as `require('test.<...>')`. It works both with luatest under test-run and pure luatest.

PR #336, opened 2022-04-05
Add option to tune test run retries after failure

### LINKED ISSUE #334: Option to disable test auto-rerun
It complicates investigation of failure artifacts - rerun just creates more of them making it hard to find needed info.

ylobankov:
This change adds the --retries (-r) option for having a facility to tune
the number of test run retries after a failure. Besides regular tests,
it is also the default value for 'fragile' tests unless the `retries`
option is set in the suite.ini config file.

Also, it introduces the `TEST_RUN_RETRIES` env variable to tune retries
count via shell (useful for CI systems).

Closes #334

### DIFF
--- a/lib/options.py
+++ b/lib/options.py
@@ -208,10 +208,10 @@ class Options(object):
         parser.add_argument(
                 "-r", "--retries",
                 dest='retries',
-                default=env_int('TEST_RUN_RETRIES', 3),
+                default=env_int('TEST_RUN_RETRIES', 0),
                 type=int,
                 help="""The number of test run retries after a failure.
-                Default: ${TEST_RUN_RETRIES} or 3. It is also the default value
+                Default: ${TEST_RUN_RETRIES} or 0. It is also the default value
                 for 'fragile' tests unless the `retries` option is set in the
                 suite.ini config file.""")
 
--- a/lib/options.py
+++ b/lib/options.py
@@ -206,6 +206,16 @@ class Options(object):
                 (single process). """)
 
         parser.add_argument(
+                "-r", "--retries",
+                dest='retries',
+                default=env_int('TEST_RUN_RETRIES', 3),
+                type=int,
+                help="""The number of test run retries after a failure.
+                Default: ${TEST_RUN_RETRIES} or 3. It is also the default value
+                for 'fragile' tests unless the `retries` option is set in the
+                suite.ini config file.""")
+
+        parser.add_argument(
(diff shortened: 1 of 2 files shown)


NickVolynkin (approved):

lib/options.py
```diff
@@ -205,6 +205,16 @@ def __init__(self):
                 2 x CPU count). -1 means everything running consistently
                 (single process). """)
 
+        parser.add_argument(
+                "-r", "--retries",
+                dest='retries_count',
+                default=env_int('TEST_RUN_RETRIES_COUNT', 3),
```
locker, line 211: IMO the default value should be 0 and we should explicitly set the retry count for CI runs. Rationale: when a developer writes a new test and runs test-run manually, he expects failures so he has to set the retry counter to 0.

ylobankov: To be honest, I also thought about this in such a way. I like your idea and it is close to my feeling how should be right. But I had a small conversation about it with @Totktonada, maybe a month ago, and AFAIR he was sceptical about the default value == 0. AFAIR, @Totktonada was worry about the number of places where we need to do changes to enable auto-rerun. But it shouldn't be a hard task because at least in tarantool we have `environment` action where we can set the corresponding env variable `TEST_RUN_RETRIES_COUNT` and it automatically propagates through all our workflows because all of them use `environment` action. So I am for the default value == 0.

locker (approved):

lib/options.py
Totktonada: We discussed it with Yaroslav. There are points for and against, but is seems that we should just solve the problem, which is more painful at the moment and look what will going on. IOW, I'm okay to set the default to "don't retry".

----

For the sake of completeness, points for the "retry 3 times" default:

* Infra: we don't need to proxy the variable to all environments in CI (to packpack, which in turn passes it to rpmbuild/debuild; to a docker container; to a virtual machine). It is implemented now, but may cost a time for support.
* Dev: if we'll apply the "no retry" default, running the full test suite will require more manual work to retry failed attempts (when irrelevant tests will fail). Since those attempts are around 5 minutes, context switching here is not a nice thing.

Points for the "no retry" default:

* Dev: no extra test-run specific knowledge is necessary to debug a test.

Totktonada (approved):
LGTM.

lib/options.py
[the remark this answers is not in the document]
ylobankov: Fixed.

PR #337, opened 2022-04-07
Prettify `test-run.py --help` output

Totktonada:


Totktonada:
NB: Don't merge before PR #326.

Totktonada:
Rebased after PR #326.

lib/options.py
```diff
@@ -201,135 +286,190 @@ def __init__(self):
...
                 type=int,
-                help="""The number of test run retries after a failure.
-                Default: ${TEST_RUN_RETRIES} or 0. It is also the default value
-                for 'fragile' tests unless the `retries` option is set in the
-                suite.ini config file.""")
+                help=format_help(
+                    """
+                    The number of test run retries after a failure.
+
+                    Default: ${TEST_RUN_RETRIES} or 0. It is also the default
```
ylobankov, line 308: Let's move `Default: ${TEST_RUN_RETRIES} or 0.` to the end like it's done everywhere.

lib/options.py
```diff
@@ -201,135 +286,190 @@ def __init__(self):
...
                 "--server-start-timeout",
                 dest="server_start_timeout",
                 default=env_int('SERVER_START_TIMEOUT', 90),
                 type=int,
-                help="""Stop the server process if the server starts longer than
-                this amount of seconds. Default: 90 [seconds].""")
+                help=format_help(
+                    """
+                    Stop the server process if the server starts longer than
+                    this amount of seconds. Default: 90 [seconds].
```
ylobankov, line 335: `Default: 90 [seconds].` should be on the new line.

lib/options.py
```diff
@@ -201,135 +286,190 @@ def __init__(self):
...
+                    Useful for, say, running built-in luatest executable.
 
-                Usage: just source it into a current environment:
+                    Usage: just source it into a current environment:
 
-                ```
-                . <(./test/test-run.py --env)
-                ```
-                """)
+                    ```
```
ylobankov, line 469: Why not just `$ . <(./test/test-run.py --env)`?

lib/options.py
Totktonada: Done. Resolving the discussion as obviously fixed.

lib/options.py
Totktonada: Done. Resolving the discussion as obviously fixed.

lib/options.py
Totktonada: It feels for me that I should somehow mark a code block. It is not rendered anyhow specific in the `/path/to/test-run.py --help` output, but Markdown is so widespread that I found it acceptable to use it in the plain text for this purpose.

It has not much meaning here, more a matter of taste. I dropped the lines with backticks.

lib/options.py
```diff
@@ -65,134 +75,209 @@ def __init__(self):
...
         parser.add_argument(
                 "--builddir",
                 dest="builddir",
                 default="..",
-                help="""Path to project build directory. Default: ".." """)
+                help=format_help(
+                    """
+                    Path to project build directory.
+
+                    Default: ".."
```
ylobankov, line 239: 1. The dot is missing at the end of the sentence.
2. IMHO, `Default: ../ .` looks better.

But all this is up to you.

ylobankov (approved):

lib/options.py
Totktonada: I would prefer the following:

```diff
diff --git a/lib/options.py b/lib/options.py
index 834ba91..72dd77b 100644
--- a/lib/options.py
+++ b/lib/options.py
@@ -236,7 +236,11 @@ class Options(object):
                     """
                     Path to project build directory.
 
-                    Default: ".."
+                    Beware: A relative path is resolved against the directory,
+                    where all tests reside. **NOT** a current directory of a
+                    parent shell.
+
+                    Default: .. (parent directory).
                     """))
 
         parser.add_argument(
@@ -257,6 +261,10 @@ class Options(object):
                     """
                     Path to data directory.
 
+                    Beware: A relative path is resolved against the directory,
+                    where all tests reside. **NOT** a current directory of a
+                    parent shell.
+
                     Default: var.
                     """))
```

Applied this change and force-pushed.

ylobankov: I am ok with it.

PR #338, opened 2022-04-27
Change default value for vardir option 

ylobankov:
Now the default value for the `vardir` option is `/tmp/t`. This change
should help to solve the issue with the max length of the socket paths
(107 chars) in most cases.

Also, the default value for the `vardir` option is taken from the `VARDIR`
env variable when it is defined.

### DIFF
--- a/lib/options.py
+++ b/lib/options.py
@@ -256,7 +256,7 @@ class Options(object):
         parser.add_argument(
                 "--vardir",
                 dest="vardir",
-                default="var",
+                default='/tmp/t',
                 help=format_help(
                     """
                     Path to data directory.
@@ -265,7 +265,7 @@ class Options(object):
                     where all tests reside. **NOT** a current directory of a
                     parent shell.
 
-                    Default: var.
+                    Default: /tmp/t.
                     """))
 
         parser.add_argument(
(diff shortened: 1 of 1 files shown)


lib/options.py
```diff
@@ -256,7 +256,7 @@ def __init__(self):
         parser.add_argument(
                 "--vardir",
                 dest="vardir",
-                default="var",
+                default='/tmp/t',
```
Totktonada: Let's also change default in the help message.

lib/options.py
```diff
@@ -256,7 +256,7 @@ def __init__(self):
         parser.add_argument(
                 "--vardir",
                 dest="vardir",
-                default="var",
+                default=os.environ.get('VARDIR') or '/tmp/t',
```
Totktonada, line 259: Let's reflect this logic in the help message.

Totktonada (approved):
I'm ok.

lib/options.py
ylobankov: Fixed.

lib/options.py
ylobankov: Fixed.

PR #348, opened 2022-09-01
Free port auto resolving for TarantoolServer and AppServer

### LINKED ISSUE #141: Fix 'Address already in use' error with TCP sockets

ylobankov:
This change replaces manual choosing an iproto port for TarantoolServer
by the auto resolving mechanism. In two words, test-run always provides
'127.0.0.1:0' as a value for LISTEN env variable that is used in a lua
file to start a tarantool instance. In this way, the port will be picked
automatically, and we are getting the real value of it via the admin
console by executing `box.info.listen` that is available for tarantool
version >= 2.4.1, and special lua script intended for tarantool version
< 2.4.1.

The same for AppServer, test-run always provides '127.0.0.1:0' as a
value for LISTEN env variable that is used in a lua test script. In
this way, the iproto port will be picked automatically, but if the test
needs the real value of the port, it has to execute `box.info.listen`
for tarantool version >= 2.4.1, or other lua code for tarantool version
< 2.4.1.

Fixes #141

### DIFF
--- a/dispatcher.py
+++ b/dispatcher.py
@@ -46,35 +46,6 @@ from listeners import OutputWatcher
 from listeners import StatisticsWatcher
 
 
-class TcpPortDispatcher:
-    """ Helper class holds available and occupied TCP port ranges. This ranges
-    intended to distributes between workers.
-    """
-    def __init__(self, range_count):
-        lowest_port = 3000
-        highest_port = 59999
-        port_count = highest_port - lowest_port + 1
-        range_size = port_count // range_count
-
-        self.available_ranges = set()
-        for i in range(range_count):
-            start_port = lowest_port + i * range_size
-            end_port = start_port + range_size - 1
-            tcp_port_range = (start_port, end_port)
-            self.available_ranges.add(tcp_port_range)
-
-        self.acquired_ranges = dict()
-
-    def acquire_range(self, _id):
-        tcp_port_range = self.available_ranges.pop()
-        self.acquired_ranges[_id] = tcp_port_range
-        return tcp_port_range
-
-    def release_range(self, _id):
-        tcp_port_range = self.acquired_ranges.pop(_id)
-        self.available_ranges.add(tcp_port_range)
-
-
 class Dispatcher:
     """Run specified count of worker processes ('max_workers_cnt' arg), pass
     task IDs (via 'task_queue'), receive results and output (via
@@ -136,8 +107,6 @@ class Dispatcher:
         self.worker_id_to_pid = dict()
 
         self.randomize = randomize
-        self.tcp_port_dispatcher = TcpPortDispatcher(
-            range_count=max_workers_cnt)
 
     def terminate_all_workers(self):
         for process in self.processes:
@@ -235,10 +204,7 @@ class Dispatcher:
         # find_nonempty_task_queue_disp()
         if self.workers_cnt >= self.max_workers_cnt:
             return False
-        tcp_port_range = self.tcp_port_dispatcher.acquire_range(
-            self.worker_next_id)
-        process = task_queue_disp.add_worker(self.worker_next_id,
-                                             tcp_port_range)
+        process = task_queue_disp.add_worker(self.worker_next_id)
         self.processes.append(process)
         self.pids.append(process.pid)
         self.pid_to_worker_id[process.pid] = self.worker_next_id
@@ -255,7 +221,6 @@ class Dispatcher:
         task_queue_disp = self.get_task_queue_disp(worker_id)
         task_queue_disp.del_worker(worker_id)
         self.workers_cnt -= 1
-        self.tcp_port_dispatcher.release_range(worker_id)
 
         self.pids.remove(pid)
         del self.worker_id_to_pid[worker_id]
@@ -412,24 +377,22 @@ class TaskQueueDispatcher:
         self.done = False
         self.done_task_ids = set()
 
-    def _run_worker(self, worker_id, tcp_port_range):
+    def _run_worker(self, worker_id):
         """Entry function for worker processes."""
         os.environ['TEST_RUN_WORKER_ID'] = str(worker_id)
-        os.environ['TEST_RUN_TCP_PORT_START'] = str(tcp_port_range[0])
-        os.environ['TEST_RUN_TCP_PORT_END'] = str(tcp_port_range[1])
         color_stdout.queue = self.result_queue
         worker = self.gen_worker(worker_id)
         sampler.set_queue(self.result_queue, worker_id, worker.name)
         worker.run_all(self.task_queue, self.result_queue)
 
-    def add_worker(self, worker_id, tcp_port_range):
+    def add_worker(self, worker_id):
         # Note: each of our workers should consume only one None, but for the
         # case of abnormal circumstances we listen for processes termination
         # (method 'check_for_dead_processes') and for time w/o output from
         # workers (class 'HangWatcher').
         self.task_queue.put(None)  # 'stop worker' marker
 
-        entry = functools.partial(self._run_worker, worker_id, tcp_port_range)
+        entry = functools.partial(self._run_worker, worker_id)
 
         self.worker_ids.add(worker_id)
         process = multiprocessing.Process(target=entry)
--- a/lib/app_server.py
+++ b/lib/app_server.py
@@ -19,7 +19,6 @@ from lib.server import DEFAULT_SNAPSHOT_NAME
 from lib.tarantool_server import Test
 from lib.tarantool_server import TarantoolServer
 from lib.tarantool_server import TarantoolStartError
-from lib.utils import find_port
 from lib.utils import format_process
 from lib.utils import signame
 from lib.utils import warn_unix_socket
@@ -33,7 +32,7 @@ def timeout_handler(server_process, test_timeout):
 
 
 def run_server(execs, cwd, server, logfile, retval, test_id):
-    os.putenv("LISTEN", server.iproto)
+    os.putenv("LISTEN", server.listen_uri)
     server.process = Popen(execs, stdout=PIPE, stderr=PIPE, cwd=cwd)
     sampler.register_process(server.process.pid, test_id, server.name)
     test_timeout = Options().args.test_timeout
@@ -113,6 +112,7 @@ class AppServer(Server):
         self.lua_libs = ini['lua_libs']
         self.name = 'app_server'
         self.process = None
+        self.localhost = '127.0.0.1'
         self.use_unix_sockets_iproto = ini['use_unix_sockets_iproto']
 
     @property
@@ -156,9 +156,9 @@ class AppServer(Server):
         if self.use_unix_sockets_iproto:
             path = os.path.join(self.vardir, self.name + ".i")
             warn_unix_socket(path)
-            self.iproto = path
+            self.listen_uri = path
         else:
-            self.iproto = str(find_port())
+            self.listen_uri = self.localhost + ':0'
         shutil.copy(os.path.join(self.TEST_RUN_DIR, 'test_run.lua'),
                     self.vardir)
 
--- a/lib/tarantool_server.py
+++ b/lib/tarantool_server.py
@@ -10,6 +10,7 @@ import shutil
 import signal
 import subprocess
 import sys
+import textwrap
 import time
 import yaml
 
@@ -655,12 +656,15 @@ class TarantoolServer(Server):
             'lua_libs': [],
             'valgrind': False,
             'vardir': None,
-            'use_unix_sockets': False,
             'use_unix_sockets_iproto': False,
             'tarantool_port': None,
             'strace': False
         }
         ini.update(_ini)
+        if ini.get('use_unix_sockets') is not None:
+            qa_notice(textwrap.fill('The "use_unix_sockets" option is defined '
+                                    'in suite.ini, but it was dropped in favor '
+                                    'of permanent using Unix sockets'))
         Server.__init__(self, ini, test_suite)
         self.testdir = os.path.abspath(os.curdir)
         self.sourcedir = os.path.abspath(os.path.join(os.path.basename(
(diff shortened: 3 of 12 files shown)


lib/tarantool_server.py
```diff
@@ -906,8 +907,9 @@ def start(self, silent=True, wait=True, wait_load=True, rais=True, args=[],
         self.crash_detector.info = "Crash detector: %s" % self.process
         self.crash_detector.start()
 
+        deadline = time.time() + Options().args.server_start_timeout
```
Totktonada, line 910: Unnecessary change?

Totktonada: Ah, I see, `self.get_iproto_port(deadline)`. Sorry for noise.

lib/tarantool_server.py
```diff
@@ -872,6 +868,11 @@ def start(self, silent=True, wait=True, wait_load=True, rais=True, args=[],
         color_log(prefix_each_line(' | ', self.version()) + '\n',
                   schema='version')
 
+        # When the server is restarted, we need to reset port.
+        if not self.use_unix_sockets_iproto:
+            if self.iproto.port > 0:
+                self._iproto = 0
```
Totktonada, line 874: TBH, this jugglish seems confusing for me:

```python
iproto = '127.0.0.1:0'
env['IPROTO'] = iproto
popen('tarantool')
iproto = get_iproto_port()  # '127.0.0.1:xxxx'
```

At different point the iproto variable means either an address to listen on (always with zero port) or an address to connect to (advertize address). I think the code would be simpler if we would separate those objects.

In fact, we don't need to store '127.0.0.1:0' anyhow, just place it to the env variable:

```python
env['IPROTO'] = '127.0.0.1:0'
open('tarantool')
iproto = get_iproto_address()
```

I don' insist, feel free to ignore.

lib/tarantool_server.py
```diff
@@ -1270,3 +1276,79 @@ def wait_lsn(self, node_id, lsn):
...
+            # At first, it catches both server (listening) and client (sending)
+            # sockets (client sockets usually occur when starting an instance as
+            # a replica). Then it finds the server socket among caught sockets.
+            script = """
+                ffi = require('ffi')
+                socket = require('socket')
+
+                val = ffi.new('int[1]')
+                len = ffi.new('size_t[1]', ffi.sizeof('int'))
+                while not {deadline} or os.time() < {deadline} do
```
Totktonada, line 1305: When `deadline` is None, it'll be `while not None <...>` in Lua. It means accessing to a `None` variable. If it is undefined (it is likely so), this code will fail in the strict mode.

lib/tarantool_server.py
```diff
@@ -1270,3 +1276,79 @@ def wait_lsn(self, node_id, lsn):
...
+                    for fd = 0, 65535 do
+                        local addrinfo = socket.internal.name(fd)
+                        local is_matched = addrinfo ~= nil and
+                            addrinfo.host == '{localhost}' and
+                            addrinfo.family == 'AF_INET' and
+                            addrinfo.type == 'SOCK_STREAM' and
+                            addrinfo.protocol == 'tcp' and
+                            type(addrinfo.port) == 'number'
+                        if is_matched then
+                            local key = socket.internal.SOL_SOCKET
```
Totktonada, line 1315: Nit: It is named `level` in `man getsockopt`.

lib/tarantool_server.py
```diff
@@ -1270,3 +1276,79 @@ def wait_lsn(self, node_id, lsn):
...
+                                socket.internal.SO_OPT[key].SO_REUSEADDR.iname,
+                                val, len)
+                            if val[0] > 0 then
+                                return {{
+                                    host = addrinfo.host, port = addrinfo.port
+                                }}
+                            end
+                        end
+                    end
+                    os.execute("sleep " .. {sleep_time})
```
Totktonada, line 1326: Why not just `fiber.sleep({sleep_time})`? Moreover, if you stopped all the fibers, unlikely any of them will change anything. No reason to spin in the loop so.

lib/tarantool_server.py
```diff
@@ -1270,3 +1276,79 @@ def wait_lsn(self, node_id, lsn):
...
+            return
+
+        deadline = deadline or 'nil'
+        sleep_time = 0.001
+        timeout = Options().args.server_start_timeout
+
+        version = yaml.safe_load(
+            self.admin("require('tarantool').version", silent=True)
+        )[0]
+        if version.split('.')[0] == '1':
```
Totktonada, line 1293: You can just execute this code if `box.info.listen` gives `nil`. No need explicit version check.

```lua
while <...> do
    local res = box.info.listen
    if res ~= nil then
        return <...>
    end
    -- <..tarantoool < 2.4.1 code..>
end
```


The explicit version check is worse than direct attempts to use a feature you need, at least because it'll work for versions before 2.4.0-103-g2c3acf2b0.

lib/tarantool_server.py
```diff
@@ -1270,3 +1276,79 @@ def wait_lsn(self, node_id, lsn):
...
+                            type(addrinfo.port) == 'number'
+                        if is_matched then
+                            local key = socket.internal.SOL_SOCKET
+                            ffi.C.getsockopt(fd, key,
+                                socket.internal.SO_OPT[key].SO_REUSEADDR.iname,
+                                val, len)
+                            if val[0] > 0 then
+                                return {{
+                                    host = addrinfo.host, port = addrinfo.port
+                                }}
```
Totktonada, line 1322: Don't you want to detect the situation, when several listening ports are matched? It would be better to give an explicit error instead of returning a random address.

lib/app_server.py
```diff
@@ -158,7 +158,7 @@ def deploy(self, vardir=None, silent=True, need_init=True):
             warn_unix_socket(path)
             self.iproto = path
         else:
-            self.iproto = str(find_port())
+            self.iproto = self.localhost + ':0'
```
Totktonada, line 161: AFAIR it is only needed for the `create server` command with the `return_listen_uri` option. Now it'll return an unsuitable result.

Feel free to drop this feature together with the `self.iproto` field if you don't see a real need, but, please, don't leave it broken.

dispatcher.py
```diff
@@ -46,35 +46,6 @@
 from listeners import StatisticsWatcher
 
 
-class TcpPortDispatcher:
```
Totktonada, line 49: How about removing of `utils.find_port()`?

Totktonada (approved):
Thank you for working on the ancient (but painful) problem!

I'm okay on the high level. I see a potential problem with the `create server` command, it should be handled somehow (see details in the comment above).

Also left several stylistic thoughts, but they're not crucial to fix.

I just looked at the code and lean on you in performing a thorough testing on tarantool and vshard test suites.

(I'm going to a vacation and can't participate in the next review iteration.)

lib/tarantool_server.py
ylobankov: Agree, there is no sense to define `iproto` attribute until a real address is known.

lib/tarantool_server.py
ylobankov: I have above `deadline = deadline or 'nil'`.

lib/tarantool_server.py
ylobankov: We have some tests where a server starts without waiting for the full load. I don't know exactly, but I think there may be a situation when `box.info.listen` will give `nil` (that's why we do it in the loop) and in this case we will execute 'tarantoool < 2.4.1' code that may produce wrong result.

dispatcher.py
ylobankov: `find_port()` is still used in inspector.py.

lib/app_server.py
ylobankov: No, it is not used for the `create server` command with the `return_listen_uri` option. The `create server` command uses `TarantoolServer` (please check [this](https://github.com/tarantool/test-run/blob/b59464f17f391ecffd985a2366777bc3cb354eca/lib/app_server.py#L54-L57)).

Actually, this is just an attribute that is used only [here](https://github.com/tarantool/test-run/blob/b59464f17f391ecffd985a2366777bc3cb354eca/lib/app_server.py#L36). I guess, the line `os.putenv("LISTEN", server.iproto)` exists only because some app tests have `box.cfg {listen = os.getenv('LISTEN')}` inside the test script. 

I can remove this attribute and move this `if ... else ...` block to the `run_server` function.

lib/tarantool_server.py
```diff
@@ -774,7 +770,7 @@ def install(self, silent=True):
             warn_unix_socket(path)
             self._iproto = path
         else:
-            self._iproto = find_port()
+            self._iproto = 0
```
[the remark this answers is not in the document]
ylobankov: We can remove this line.

lib/app_server.py
ylobankov: Or rename it into `self.listen` to not be confused.

lib/tarantool_server.py
ylobankov: Fixed.

lib/tarantool_server.py
```diff
@@ -1270,3 +1276,79 @@ def wait_lsn(self, node_id, lsn):
...
+            # This is a simple script to get the server socket of the instance.
+            # At first, it catches both server (listening) and client (sending)
+            # sockets (client sockets usually occur when starting an instance as
+            # a replica). Then it finds the server socket among caught sockets.
+            script = """
+                ffi = require('ffi')
+                socket = require('socket')
+
+                val = ffi.new('int[1]')
+                len = ffi.new('size_t[1]', ffi.sizeof('int'))
```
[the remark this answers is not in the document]
ylobankov: Fixed.

lib/tarantool_server.py
ylobankov: Fixed.

lib/tarantool_server.py
ylobankov: Fixed.

lib/app_server.py
ylobankov: Fixed.

lib/tarantool_server.py
Totktonada: Doesn't it mean that `wait=False wait_load=False` now will perform waiting anyway?

dispatcher.py
Totktonada: And there is the same race here, right? See, it is the simplest place: here we acquire a port, listen and use it later. Nothing prevent us from binding to zero, listen, get the real port and add it into the environment. Moreover, nothing actually prevent us from using Unix sockets here. A worker has one inspector, a worker has one working directory.

It would be interesting to reproduce the situation, when we get the port clash here. Maybe we observed it many times but didn't know the reason.

lib/tarantool_server.py
Totktonada: Doesn't it mean that we should either:

* if waiting is acquired, do it,
* otherwise run a greenlet to poll the server and set the `iproto` attribute (don't forget to finish it when the server is stopped)?

ylobankov: > Doesn't it mean that `wait=False wait_load=False` now will perform waiting anyway?

Looks like not. Note, the `wait=False wait_load=False` and `wait=True wait_load=False` cases are identical from the point of view of getting the iproto socket. Look at [this](https://github.com/tarantool/test-run/blob/b59464f17f391ecffd985a2366777bc3cb354eca/lib/tarantool_server.py#L1143-L1152). When `wait=True wait_load=False`, we test only the admin connection and return. When  `wait=True wait_load=True`, `box.info.listen` gives the socket even though `box.info.status` is `loading`.

Totktonada: You assume that `box.cfg()` is successful (or will be successful) and so binding a listening socket for iproto is successful. It may be not the case for a `wait_load=False` instance. AFAIR, we have test cases, where a supplementary server is expected to fail loading.

dispatcher.py
ylobankov: Ok, it was trivial to switch to free port auto resolving in `TarantoolInspector`. Fixed.

Totktonada: Nice! (It would be good to mark the new two commits as part of/related/fixes the issue.)

lib/tarantool_server.py
ylobankov: I realized that we shouldn't get the iproto port if `wait=False` because we can get the port only when the admin connection can be established and tested (`wait=True` guarantees it). 

Also, I think we shouldn't get the iproto port when `crash_expected=True`. When a server is going to be failed due to some error (for example, when `box.cfg()` is unsuccessful) and we expect this, we have to provide `crash_expected=True`, otherwise the test will fail.

ylobankov: I would like to get the iproto port only when `wait=True wait_load=True crash_expected=False`. In this case, we can use the code (without `while` loop) that you proposed in your first comment. It really simplifies the things, but if a server starts with `wait=False` or `wait_load=False`, the iproto socket will be missing for it. But it's not the case for tests where Unix sockets are used (almost all). Anyway, all tests passed when I did testing, and diff and app tests are deprecated in favor of luatest ones and not produced anymore.

PR #353, opened 2022-10-17
Fix determination of debug build

### LINKED ISSUE #352: Determination of debug build is wrong if compilation flags contain `-Wa,--debug-prefix-map`
**Steps to reproduce**
```console
cmake -DCMAKE_BUILD_TYPE=RelWithDebInfo -DCMAKE_C_FLAGS="-Wa,--debug-prefix-map=="
test/test-run.py box-tap/gh-6198-max-cnt-of-tuple-fields-insert-overflow-err-msg.test.lua
```

**Actual behaviour**

```console
[001] box-tap/gh-6198-max-cnt-of-tuple-fields-insert>                 [ fail ]
```

**Expected behaviour**

```console
[001] box-tap/gh-6198-max-cnt-of-tuple-fields-insert>                 [ disabled ]
```

CuriousGeorgiy:
Determination of debug build uses an incorrect regular expression: since it uses `re.I` flag (case insensitivity), it can match compiler flags contained in the version string, e.g. '--Wa,--debug-prefix-map==' — remove this flag and make the regular expression more specific.

Closes #352

### DIFF
--- a/lib/app_server.py
+++ b/lib/app_server.py
@@ -224,7 +224,7 @@ class AppServer(Server):
     def find_exe(cls, builddir):
         cls.builddir = builddir
         cls.binary = TarantoolServer.binary
-        cls.debug = bool(re.findall(r'-Debug', str(cls.version()),
+        cls.debug = bool(re.findall(r'^Target:.*-Debug$', str(cls.version()),
                                     re.I))
 
     @staticmethod
--- a/lib/luatest_server.py
+++ b/lib/luatest_server.py
@@ -90,7 +90,7 @@ class LuatestServer(Server):
     def find_exe(cls, builddir):
         cls.builddir = builddir
         cls.binary = TarantoolServer.binary
-        cls.debug = bool(re.findall(r'-Debug', str(cls.version()),
+        cls.debug = bool(re.findall(r'^Target:.*-Debug$', str(cls.version()),
                                     re.I))
 
     @classmethod
--- a/lib/tarantool_server.py
+++ b/lib/tarantool_server.py
@@ -727,8 +727,8 @@ class TarantoolServer(Server):
                         ctl_dir + '/?.lua;' + \
                         ctl_dir + '/?/init.lua;' + \
                         os.environ.get("LUA_PATH", ";;")
-                cls.debug = bool(re.findall(r'-Debug', str(cls.version()),
-                                 re.I))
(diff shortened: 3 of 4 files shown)


lib/app_server.py
```diff
@@ -224,8 +224,8 @@ def kill():
     def find_exe(cls, builddir):
         cls.builddir = builddir
         cls.binary = TarantoolServer.binary
-        cls.debug = bool(re.findall(r'-Debug', str(cls.version()),
-                                    re.I))
+        color_stdout("LOGGING: " + cls.version())
```
Totktonada, line 227: Forgotten debug print (and the same in `tarantool_server.py`)?

lib/app_server.py
```diff
@@ -224,8 +224,7 @@ def kill():
     def find_exe(cls, builddir):
         cls.builddir = builddir
         cls.binary = TarantoolServer.binary
-        cls.debug = bool(re.findall(r'-Debug', str(cls.version()),
-                                    re.I))
+        cls.debug = bool(re.findall(r'^Target:.*-Debug$', str(cls.version())))
```
Totktonada, line 227: I search the web around, whether `CMAKE_BUILD_TYPE` should be interpreted case sensitive or case insensitive and found that the latter seems to be considered as a kind of the right way. For example, LLVM:

https://github.com/llvm/llvm-project/blob/699396131f745ec8b42ae33cb0e4553a783b6674/llvm/CMakeLists.txt#L371-L376
https://lists.llvm.org/pipermail/llvm-commits/Week-of-Mon-20150810/292471.html

I guess we don't handle it correctly in tarantool, but we likely should (sooner or later). I think we should reflect it in https://github.com/tarantool/tarantool/issues/3686.

I propose to leave the interpretation as case insensitive in test-run to don't change it twice forth and back.

Totktonada:
Context: https://github.com/tarantool/tarantool/pull/7809

Totktonada (approved):
Disregarding of our dispute in https://github.com/tarantool/tarantool/pull/7809 this patch seems valuable, because it eliminates possible false positives in the debug build detection.

lib/app_server.py
CuriousGeorgiy: Okay, left the regular expression case-insensitive.

ylobankov (approved):
LGTM, thanks!

PR #356, opened 2022-11-01
Fix detection of debug builds

### LINKED ISSUE #355: release_disabled tests are disabled in the debug mode as well
Tests marked as `release_disabled` in `suite.ini` are skipped when testing Tarantool built with `-Debug` mode.

```console
$ ./src/tarantool -v
Tarantool 2.11.0-entrypoint-654-g0e181b35e
Target: Darwin-x86_64-Debug
Build options: cmake . -DCMAKE_INSTALL_PREFIX=/usr/local -DENABLE_BACKTRACE=TRUE
Compiler: /usr/local/opt/ccache/libexec/cc /usr/local/opt/ccache/libexec/c++
C_FLAGS: -Wno-unknown-pragmas -fexceptions -funwind-tables -fno-common -msse2 -Wformat -Wformat-security -Werror=format-security -fstack-protector-strong -fPIC  -std=c11 -Wall -Wextra -Wno-strict-aliasing -Wno-char-subscripts -Wno-gnu-alignof-expression -Werror
CXX_FLAGS: -Wno-unknown-pragmas -fexceptions -funwind-tables -fno-common -msse2 -Wformat -Wformat-security -Werror=format-security -fstack-protector-strong -fPIC  -std=c++11 -Wall -Wextra -Wno-strict-aliasing -Wno-char-subscripts -Wno-invalid-offsetof -Wno-gnu-alignof-expression -Werror

$ cd test
$ ./test-run.py  --builddir=../build catch.test.lua

...
[001] replication/catch.test.lua                      vinyl           [ disabled ]
[002] replication/catch.test.lua                      memtx           [ disabled ]
...
```

NickVolynkin:
The `tarantool -v` output is a single multiline string. Flag re.MULTILINE is required to match it with start of line (`^`) or end of line (`$`) characters. The `.*` in the middle of the pattern does not match start/end of lines, so this expression can not match something like this:

```
Target: Darwin-x86_64-Release
Build options: ... -Debug ...
```

The re.IGNORECASE is not required, because the build mode is always spelled in one particular way: `-Debug`.

Follow-up to #352
Fixes #355

Co-authored-by: Serge Petrenko <sergepetrenko@tarantool.org>

### DIFF
--- a/lib/tarantool_server.py
+++ b/lib/tarantool_server.py
@@ -728,19 +728,21 @@ class TarantoolServer(Server):
                         ctl_dir + '/?/init.lua;' + \
                         os.environ.get("LUA_PATH", ";;")
                 cls.debug = bool(re.findall(r'^Target:.*-Debug$', str(cls.version()),
-                                            re.I))
+                                            re.M))
                 return exe
         raise RuntimeError("Can't find server executable in " + path)
 
     @classmethod
     def print_exe(cls):
-        color_stdout('Tarantool server information\n', schema='info')
+        color_stdout('Tarantool server information:\n', schema='info')
         if cls.binary:
             color_stdout(' | Found executable at {}\n'.format(cls.binary))
         if cls.ctl_path:
             color_stdout(' | Found tarantoolctl at {}\n'.format(cls.ctl_path))
         color_stdout('\n' + prefix_each_line(' | ', cls.version()) + '\n',
                      schema='version')
+        color_stdout('Detected build mode: {}\n'.format(
+                     'Debug' if cls.debug else 'Release'), schema='info')
 
     def install(self, silent=True):
         if self._start_against_running:
--- a/lib/app_server.py
+++ b/lib/app_server.py
@@ -225,7 +225,7 @@ class AppServer(Server):
         cls.builddir = builddir
         cls.binary = TarantoolServer.binary
         cls.debug = bool(re.findall(r'^Target:.*-Debug$', str(cls.version()),
-                                    re.I))
+                                    re.M))
 
     @staticmethod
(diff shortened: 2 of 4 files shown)


sergepetrenko (approved):
Thanks for a quick fix! LGTM.

lib/tarantool_server.py
```diff
@@ -741,6 +741,8 @@ def print_exe(cls):
             color_stdout(' | Found tarantoolctl at {}\n'.format(cls.ctl_path))
         color_stdout('\n' + prefix_each_line(' | ', cls.version()) + '\n',
                      schema='version')
+        color_stdout('Detected build mode: {}\n'.format(
```
ylobankov, line 744: Could you please add `:` to L737 for consistency?

NickVolynkin: done

ylobankov (approved):
LGTM

Totktonada:
It was merged so fast that I had no change to point to https://github.com/tarantool/test-run/pull/353#discussion_r997015987. Anyway, that's minor.

Totktonada:
I didn't get it initially, but understood now.

> `^`
> (Caret.) Matches the start of the string, and **in [MULTILINE](https://docs.python.org/3/library/re.html#re.MULTILINE) mode** also matches immediately after each newline.
>
> `$`
> Matches the end of the string or just before the newline at the end of the string, and **in [MULTILINE](https://docs.python.org/3/library/re.html#re.MULTILINE) mode** also matches before a newline. <...>

https://docs.python.org/3/library/re.html

IOW, `^` and `$` do not match start and end of a line without the multiline mode.

The problem was introduced in PR #353.

PR #359, opened 2022-11-17
Fix README: remove outdated engine mentions

sergos:
We have no 'sophia' engine, since it was renamed to 'vinyl'.


ylobankov (approved):
LGTM. I will merge this change into `master` after https://github.com/tarantool/test-run/pull/362 is landed to not get failed tests.

PR #365, opened 2022-12-12
ci: refactoring ci workflow

### LINKED ISSUE #363: Consider doing refactoring for CI workflow
What can be done:

- Drop old Python versions (2.7, 3.5, 3.6).
- Update Tarantool versions, add 2.8 and 2.10, probably remove 2.6 and 2.7.
- Drop workarounds around `pip`.
- etc.

slavakirichenko:
* Drop old Python versions
* Replace Tarantool versions to 2.8 and 2.10
* Update version of actions/checkout to v3
* Add support of multiline answer on Tarantool                                                                            
  version 2.8 and one line support in 2.10 using                                                                           
  push_filter function in test at call.test.py. 

Resolves #363


.github/workflows/test.yml
```diff
@@ -13,11 +13,11 @@ jobs:
...
       matrix:
-        python-version: ['2.7', '3.5', '3.6', '3.7', '3.8', '3.9', '3.10']
-        tarantool-version: [2.6, 2.7]
+        python-version: ['3.7', '3.8', '3.9', '3.10']
+        tarantool-version: ['2.8', '2.10']
 
     steps:
-    - uses: actions/checkout@v2
+    - uses: actions/checkout@v3
     - name: update git submodules
```
ylobankov, line 21: I think we can use the corresponding options in actions/checkout to do a submodules update.

ylobankov (commented):
Please take a look at workflow runs and fix the following warning as well:

```
Node.js 12 actions are deprecated. For more information see: https://github.blog/changelog/2022-09-22-github-actions-all-actions-will-begin-running-on-node16-instead-of-node12/. Please update the following actions to use Node.js 16: tarantool/setup-tarantool@v1, actions/setup-python@v2
```

NickVolynkin (changes requested):
Looks good, but please fix this as well.

> Node.js 12 actions are deprecated. For more information see: https://github.blog/changelog/2022-09-22-github-actions-all-actions-will-begin-running-on-node16-instead-of-node12/. Please update the following actions to use Node.js 16: tarantool/setup-tarantool@v1, actions/setup-python@v2

.github/workflows/test.yml
slavakirichenko: fixed.

NickVolynkin (approved):

.github/workflows/test.yml
```diff
@@ -17,15 +17,16 @@ jobs:
         tarantool-version: [2.6, 2.7]
 
     steps:
-    - uses: actions/checkout@v2
-    - name: update git submodules
-      run: git submodule update --init --recursive
+    - uses: actions/checkout@v3
+      with:
```
ylobankov, line 21: Shouldn't it be a part of refactoring? Looks like 

```yaml
with:
  fetch-depth: 1
  submodules: recursive
```

should go to another commit.

slavakirichenko: fixed.

.github/workflows/test.yml
```diff
@@ -13,11 +13,14 @@ jobs:
...
+        python-version: ['3.7', '3.8', '3.9', '3.10']
+        tarantool-version: ['2.8', '2.10']
 
     steps:
     - uses: actions/checkout@v3
+      with:
+        fetch-depth: 1
+        submodules: recursive
     - name: update git submodules
       run: git submodule update --init --recursive
```
ylobankov, line 25: Looks like it is not needed anymore.

slavakirichenko: Fixed.

ylobankov (approved):
LGTM

PR #367, opened 2023-01-17
test_run.lua: add ignore_zero option to get_vclock

grafin:
Sometimes we don't care about zero component of the vclock, for example we want to ignore it when waiting for replica sync, using vclock (don't care about local spaces).

### DIFF
--- a/test_run.lua
+++ b/test_run.lua
@@ -79,8 +79,13 @@ local function wait_lsn(self, waiter, master)
     end
 end
 
-local function get_vclock(self, node)
-    return self:get_param(node, 'vclock')[1]
+local function get_vclock(self, node, opts)
+    local opts = opts or {}
+    local vclock = self:get_param(node, 'vclock')[1]
+    if opts.ignore_zero then
+        vclock[0] = nil
+    end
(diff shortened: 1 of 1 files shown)


test_run.lua
```diff
@@ -79,8 +79,13 @@ local function wait_lsn(self, waiter, master)
     end
 end
 
-local function get_vclock(self, node)
-    return self:get_param(node, 'vclock')[1]
+local function get_vclock(self, node, opts)
+    local opts = opts or {}
+    local vclock = self:get_param(node, 'vclock')[1]
+    if opts.ignore_zero then
```
ylobankov, line 85: Wouldn't `opts.ignore_local` be a better name?

grafin: I've used same name from core https://github.com/tarantool/tarantool/blob/a24714c55305280838576fa6dec9a9781763e893/src/lib/vclock/vclock.h#L285 I can rename it, if you insist.

ylobankov: Ok.

ylobankov (approved):
LGTM

PR #376, opened 2023-04-07
Properly ignore Unix sockets on artifact saving

### LINKED ISSUE #351: Artifact saving fails on a Unix domain socket file
## How to reproduce

Clone tarantool, build, patch a test, which creates a Unix domain socket, to make it fail:

```diff
diff --git a/test/app-tap/gh-2717-no-quit-sigint.test.lua b/test/app-tap/gh-2717-no-quit-sigint.test.lua
index 21bee27aa..9eba960cf 100755
--- a/test/app-tap/gh-2717-no-quit-sigint.test.lua
+++ b/test/app-tap/gh-2717-no-quit-sigint.test.lua
@@ -265,4 +265,5 @@ test:unlike(ph:info().status.state, popen.state.EXITED,
 ph:shutdown({stdin = true})
 ph:close()
 
+error('boom')
 os.exit(test:check() and 0 or 1)
```

Run and see the following error:

```python
  File "/home/alex/projects/tarantool-meta/tarantool/./test/test-run.py", line 344, in <module>
    status = main_parallel()
  File "/home/alex/projects/tarantool-meta/tarantool/./test/test-run.py", line 154, in main_parallel
    res = main_loop_parallel()
  File "/home/alex/projects/tarantool-meta/tarantool/./test/test-run.py", line 133, in main_loop_parallel
    dispatcher.artifacts.save_artifacts()
  File "/home/alex/projects/tarantool-meta/tarantool/test-run/listeners.py", line 219, in save_artifacts
    shutil.copytree(os.path.join(vardir, worker_name),
  File "/usr/lib/python3.10/shutil.py", line 558, in copytree
    return _copytree(entries=entries, src=src, dst=dst, symlinks=symlinks,
  File "/usr/lib/python3.10/shutil.py", line 512, in _copytree
    raise Error(errors)
shutil.Error: [('/tmp/t/001_app-tap/tarantool.soc', '/tmp/t/artifacts/001_app-tap/tarantool.soc', "[Errno 6] No such device or address: '/tmp/t/001_app-tap/tarantool.soc'")]
```

The important point is to choose a test, which creates a socket, which is not matched by test-run's exclusion mask:

https://github.com/tarantool/test-run/blob/e970cdb735e3c4014a2d172215ffad5470f26a26/listeners.py#L224-L228

The `gh-2717-no-quit-sigint.test.lua` test creates `tarantool.soc` socket file.

## Thoughts on a fix

The similar problem was fixed for Python 3 in pip using the `copy_function` argument of `shutil.copytree()`: https://github.com/pypa/pip/pull/6844. However we can solve it for Python 2 & 3 by converting the `ignore` argument to a function, which will look at file types.

* Python 2: https://docs.python.org/2/library/shutil.html#shutil.copytree
* Python 3: https://docs.python.org/3/library/shutil.html#shutil.copytree

I would drop the file name mask, it is semi-working.

ylobankov:
Resolves #351

### DIFF
--- a/listeners.py
+++ b/listeners.py
@@ -222,6 +222,14 @@ class ArtifactsWatcher(BaseWatcher):
         if not self.failed_workers:
             return
 
+        def copytree_ignore(path, filenames):
+            ignored_filenames = []
+            for filename in filenames:
+                filepath = os.path.join(path, filename)
+                if not (os.path.isfile(filepath) or os.path.isdir(filepath)):
+                    ignored_filenames.append(filename)
+            return ignored_filenames
+
         vardir = Options().args.vardir
         artifacts_dir = os.path.join(vardir, 'artifacts')
         artifacts_log_dir = os.path.join(artifacts_dir, 'log')
@@ -241,9 +249,7 @@ class ArtifactsWatcher(BaseWatcher):
                                      os.path.basename(reproduce_file_path)))
             shutil.copytree(os.path.join(vardir, worker_name),
                             os.path.join(artifacts_dir, worker_name),
-                            ignore=shutil.ignore_patterns(
-                                '*.i', '*.c',
-                                '*.sock', '*.control'))
+                            ignore=copytree_ignore)
         shutil.copytree(os.path.join(vardir, 'statistics'),
                         os.path.join(artifacts_dir, 'statistics'))
 


listeners.py
```diff
@@ -222,6 +223,16 @@ def save_artifacts(self):
         if not self.failed_workers:
             return
 
+        def is_socket(path):
+            return stat.S_ISSOCK(os.lstat(path).st_mode)
```
Totktonada, line 227: `os.lstat()` is used, so a symlink is not followed, so a symlink to a socket will be accepted to copy and we'll get the same error (I guess).

I think we can ignore all the special files. IOW, do the `os.path.isfile(x) or os.path.isdir(x)` check.

Totktonada (approved):
The idea looks correct for me (I didn't do the testing, only looked visually).

listeners.py
ylobankov: You are absolutely right. So let's use `os.stat()` instead that follows symbolic links. I have tested it and it works as expected.

ylobankov: Ok, if we check `os.path.isfile(x) or os.path.isdir(x)`, we will not need to check for a socket at all. Fixed.

PR #391, opened 2023-05-22
Drop setting OMP_NUM_THREADS

nshy:
It was workaround before https://github.com/tarantool/tarantool/issues/2431 was fixed (see comment [1] in the issue). Now we are going to drop sort using OpenMP so it is time to drop this workaround altogether.

[1] https://github.com/tarantool/tarantool/issues/2431#issuecomment-856007551

Part of https://github.com/tarantool/tarantool/issues/3389

### DIFF
--- a/test-run.py
+++ b/test-run.py
@@ -323,10 +323,6 @@ if __name__ == "__main__":
 
         __builtins__.open = open_as_utf8
 
-    # don't sure why, but it values 1 or 2 gives 1.5x speedup for parallel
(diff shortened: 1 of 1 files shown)


Totktonada (approved):

Totktonada:
This hunk of code is to speed up testing on tarantool versions prior to 1.7.4-260-g2e864b5a7c. Marmless, but has no sense now. I completely agree to remove it.

ylobankov (approved):

PR #395, opened 2023-05-29
lint: update flake8 to 5.*

Totktonada:
Locally flake8 reports me a style problem that is not detected in CI. This is distracting. Let's update flake8 in CI and fix the style problem.

flake8==6.\* doesn't support Python 3.7, but we have it in CI. Left flake8==5.* for a while.

### DIFF
--- a/requirements-test.txt
+++ b/requirements-test.txt
@@ -1,3 +1,3 @@
 coverage==5.*
-flake8==3.7.9
+flake8==5.*
 hypothesis==4.*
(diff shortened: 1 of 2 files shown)


requirements-test.txt
```diff
@@ -1,3 +1,3 @@
 coverage==5.*
-flake8==3.7.9
+flake8==6.*
```
ylobankov, line 2: Can we use `5.*` to avoid workaround in CI? The latest version of 5 series `5.0.4` catches the flake error.

PR #396, opened 2023-06-13
Fix bug when lua script name truncated by dot

### LINKED ISSUE #300: test-run truncates script name at `.`
Step to reproduce:
1. Create simple test files with the following content:
`box/box.truncate-at-dot.test.lua`:
```lua
env = require('test_run')
test_run = env.new()
test_run:cmd("create server test with script='box/box.truncate-at-dot.lua'")
test_run:cmd("start server test")
test_run:cmd("stop server test")
test_run:cmd("cleanup server test")
test_run:cmd("delete server test")    
```
`box/box.truncate-at-dot.lua`:
```lua
#!/usr/bin/env tarantool
box.cfg({
    listen = os.getenv('LISTEN'),
})
```
empty `box/box.truncate-at-dot.result`, only needed to run the test.

2. `./test-run.py box/box.truncate-at-dot.test.lua` fails with following output:
```
[001] Starting instance box.truncate-at-dot...
[001] Start failed: builtin/box/console.lua:865: failed to create server unix/:/home/evgeny/TARANTOOL/IPROTO/tarantool/test/var/001_box/box.socket-admin: Address already in use
[001] 
[001] [Instance "box"] Failed to start tarantool instance "box"
[001] [ fail ]
```

i-cod:
Creating a lua diff test with a name containing a dot causes the name to be parsed incorrectly. Then an attempt is made to create an tarantool instance with an incorrect name, resulting in an error. This change fixes the bug.

Fixes #300

### DIFF
--- a/lib/tarantool_server.py
+++ b/lib/tarantool_server.py
@@ -583,7 +583,7 @@ class TarantoolServer(Server):
                 delattr(self, '_script')
             return
         self._script = os.path.abspath(val)
-        self.name = os.path.basename(self._script).split('.')[0]
+        self.name = os.path.basename(self._script).rsplit(".", maxsplit=1)[0]
 
     @property
     def _admin(self):


lib/tarantool_server.py
```diff
@@ -583,7 +583,7 @@ def script(self, val):
                 delattr(self, '_script')
             return
         self._script = os.path.abspath(val)
-        self.name = os.path.basename(self._script).split('.')[0]
+        self.name = re.sub(r".lua", "", os.path.basename(self._script))
```
ylobankov, line 586: Imagine if the script is named `box.luafoobar.lua`. In this case, `self.name` will have an incorrect value.

i-cod: No, if the script is named `box.luafoobar.lua`, the `self.name` will have an expected value `box.luafoobar`

ylobankov (approved):

PR #397, opened 2023-06-29
Raise an error and log it if test timeouts are set incorrectly

### LINKED ISSUE #250: Give a warning when configured timeouts look strange
Unlikely `--test-timeout` that is above than `--no-output-timeout` has any sense.

After PR #302 we'll also have `--server-start-timeout`, which should be less than `--test-timeout`.

IOW, `--server-start-timeout` < `--test-timeout` < `--no-output-timeout`.

On the other hand, if one want to pass a large `--server-start-timeout`, we can issue a warning and automatically increase `--test-timeout` (`--server-start-timeout` + 10 seconds) and `--no-output-timeout` (`--test-timeout` + 10 seconds) for convenience of interactive work.

And the same in the opposite direction: if lower `--no-output-timeout` is set, other ones should also be lowered.

(To be done after PR #242 and PR #244.)

<details>
<summary>Irrelevant points (ones where I changed my mind)</summary>

We had the following sentence here:

> Unlikely --replication-sync-timeout that is above than --test-timeout have any sense.

`replication_sync_timeout` value it NOT a lower bound of timeout for going into the `'running'` status (`box.info.status`) — it is irrelevant here at all. An instance can set `box.info.status` to `'running'` before syncing with all upstreams. If the sync will fail, the instance will go into `'orphan'` from `'running'`.

See https://github.com/tarantool/tarantool/pull/6676#issuecomment-1029107386 and below.

We also had the following paragraph:

> On the other hand, if one want to pass a large --replication-sync-timeout, we can issue a warning and automatically increase --test-timeout (--replication-sync-timeout + 10 seconds) and --no-output-timeout (--test-timeout + 10 seconds) for convenience of interactive work.

I rewrote it for `--server-start-timeout`, `--test-timeout` and `--no-output-timeout`.
</details>

i-cod:
Added logic that abort test run and logs an error message if
condition is not met: no-output-timeout is at least 10 seconds
longer than test-timeout and test-timeout is at least 10 seconds
longer than server-start-timeout.

Closes #250

### DIFF
--- a/lib/options.py
+++ b/lib/options.py
@@ -511,7 +511,19 @@ class Options(object):
 
         self.check()
 
+        self.check_timeouts()
+
         Options._initialized = True
+
+    def check_timeouts(self) -> None:
+        default_time_offset = 10
+        if (self.args.no_output_timeout - self.args.test_timeout) < default_time_offset or \
+                (self.args.test_timeout - self.args.server_start_timeout) < default_time_offset:
+            color_stdout("Some timeouts are set incorrectly.\n"
+                         "Change the value(s) so that --no-output-timeout is at least 10 seconds "
+                         "longer than --test-timeout\nand --test-timeout is at least 10 seconds "
+                         "longer than --server-start-timeout\n", schema='error')
+            exit(1)
 
     def check(self):
         """Check the arguments for correctness."""


lib/options.py
```diff
@@ -511,8 +511,36 @@ def __init__(self):
...
+        self.correct_timeout_offset(parser)
+
         Options._initialized = True
 
+    def correct_timeout_offset(self, parser: argparse.ArgumentParser) -> None:
+        """Check and correct timeout offset if necessary."""
+        default_time_offset = 10
+        if self.args.server_start_timeout > parser.get_default('server_start_timeout'):
+            if (self.args.test_timeout - self.args.server_start_timeout) < default_time_offset:
+                color_stdout("Warning: test-timeout and no-output-timeout has been changed. \n",
```
ochaplashkin, line 523: I think this message does **not** answer the question "why".
The values changed automatically, why? What should the user do to get rid of this warning?

ochaplashkin:
Please add the issue number to the commit message.

lib/options.py
```diff
@@ -511,8 +511,36 @@ def __init__(self):
 
         self.check()
 
+        self.correct_timeout_offset(parser)
+
         Options._initialized = True
 
+    def correct_timeout_offset(self, parser: argparse.ArgumentParser) -> None:
+        """Check and correct timeout offset if necessary."""
+        default_time_offset = 10
```
ochaplashkin, line 520: You need to add a description of this offset (and what it is) to both the **commit message** and the **PR description**.

lib/options.py
i-cod: Fixed

lib/options.py
```diff
@@ -511,8 +511,20 @@ def __init__(self):
...
 
+        self.check_timeouts()
+
         Options._initialized = True
 
+    def check_timeouts(self) -> None:
+        default_time_offset = 10
+        if (self.args.no_output_timeout - self.args.test_timeout) < default_time_offset or \
+                (self.args.test_timeout - self.args.server_start_timeout) < default_time_offset:
+            color_stdout("Error: timeouts set incorrectly. Change the value so that "
```
ylobankov, line 522: What about such formatting?

```python
            color_stdout("Some timeouts are set incorrectly.\n"
                         "Change the value(s) so that --no-output-timeout is at least 10 seconds "
                         "longer than --test-timeout\nand --test-timeout is at least 10 seconds "
                         "longer than --server-start-timeout\n", schema='error')
```

lib/options.py
```diff
@@ -511,8 +511,20 @@ def __init__(self):
...
+                         "--no-output-timeout is at least 10 seconds longer than --test-timeout "
+                         "and --test-timeout is at least 10 seconds longer "
+                         "than --server-start-timeout. \n", schema='error')
+            raise ValueError
```
ylobankov, line 526: Please do `exit(-1)` here like it is done in the `check` method.

Totktonada: AFAIR, -1 is a mistake and it should be 1 (to set exit code 1). I would kindly ask to re-check it.

lib/options.py
i-cod: Fixed

lib/options.py
i-cod: Fixed

ochaplashkin: [List of common exit codes for GNU/Linux](https://www.cyberciti.biz/faq/linux-bash-exit-status-set-exit-statusin-bash/):
```
0   Success
1   Operation not permitted
2   No such file or directory
3   No such process
4   Interrupted sys
5   Input/output error
6   No such device or address
7   Argument list too long
8   Exec format error
9   Bad file descriptor
10  No child processes
11  Resource temporarily unavailable
12  Cannot allocate memory
13  Permission denied
14  Bad address
15  Block device required
16  Device or resource busy
17  File exists
18  Invalid cross-device link
19  No such device
20  Not a directory
21  Is a directory
22  Invalid argument
23  Too many open files in system
24  Too many open files
25  Inappropriate ioctl for device
26  Text file busy
27  File too large
28  No space left on device
29  Illegal seek
30  Read-only file system
31  Too many links
32  Broken pipe
33  Numerical argument out of domain
34  Numerical result out of range
35  Resource deadlock avoided
36  File name too long
37  No locks available
38  Function not implemented
39  Directory not empty
40  Too many levels of symbolic links
42  No message of desired type
43  Identifier removed
44  Channel number out of range
45  Level 2 not synchronized
46  Level 3 halted
47  Level 3 reset
48  Link number out of range
49  Protocol driver not attached
50  No CSI structure available
51  Level 2 halted
52  Invalid exchange
53  Invalid request descriptor
54  Exchange full
55  No anode
56  Invalid request code
57  Invalid slot
59  Bad font file format
60  Device not a stream
61  No data available
62  Timer expired
63  Out of streams resources
64  Machine is not on the network
65  Package not installed
66  Object is remote
67  Link has been severed
68  Advertise error
69  Srmount error
70  Communication error on send
71  Protocol error
72  Multihop attempted
73  RFS specific error
74  Bad message
75  Value too large for defined data type
76  Name not unique on network
77  File descriptor in bad state
78  Remote address changed
79  Can not access a needed shared library
80  Accessing a corrupted shared library
81  .lib section in a.out corrupted
82  Attempting to link in too many shared libraries
83  Cannot exec a shared library directly
84  Invalid or incomplete multibyte or wide character
85  Interrupted system call should be restarted
86  Streams pipe error
87  Too many users
88  Socket operation on non-socket
89  Destination address required
90  Message too long
91  Protocol wrong type for socket
92  Protocol not available
93  Protocol not supported
94  Socket type not supported
95  Operation not supported
96  Protocol family not supported
97  Address family not supported by protocol
98  Address already in use
99  Cannot assign requested address
100 Network is down
101 Network is unreachable
102 Network dropped connection on reset
103 Software caused connection abort
104 Connection reset by peer
105 No buffer space available
106 Transport endpoint is already connected
107 Transport endpoint is not connected
108 Cannot send after transport endpoint shutdown
109 Too many references
110 Connection timed out
111 Connection refused
112 Host is down
113 No route to host
114 Operation already in progress
115 Operation now in progress
116 Stale file handle
117 Structure needs cleaning
118 Not a XENIX named type file
119 No XENIX semaphores available
120 Is a named type file
121 Remote I/O error
122 Disk quota exceeded
123 No medium found
125 Operation canceled
126 Required key not available
127 Key has expired
128 Key has been revoked
129 Key was rejected by service
130 Owner died
131 State not recoverable
132 Operation not possible due to RF-kill
133 Memory page has hardware error
```

i-cod: Exit code changed to 1

ochaplashkin (approved):
LGTM :rocket:

Issue #249, opened 2020-12-06
Fail testing of tarantool when it is not built


Totktonada:
It is convenient that test-run uses a tarantool / a tarantoolctl executables found in PATH, when we want to test a module (say, vshard). However it is undesirable behaviour, when we test tarantool itself: it is not always obvious what is going on when I forgot to build tarantool prior to run tests.

We already detect tarantool source directory:

https://github.com/tarantool/test-run/blob/e84355283f117e91964fe84eac70f26bb32d7f06/lib/__init__.py#L14-L22

So it should be easy to forbid using of an external tarantool in the case.

PR #401, opened 2023-07-19
Migrate tarantoolctl from the tarantool repository

### LINKED ISSUE #400: Adapt an using `tarantoolctl` for static build and tarantool version 3.0+
After completing the work on creating a `static build`, we need to check it via lua tests.

The main problems, why do we need to change test-run:
- different location of binaries and directory tree for static build;
- `tarantoolctl` does **NOT** get into new packages;
- `tarantoolctl` will be replaced by tt (sometime in the future).

ochaplashkin:
## What's the problem?

In the near future we plan to test Tarantool installed from a DEB/RPM package. Since 3.0.0-alpha1 release [^1] Tarantool packages don't have the tarantoolctl utility inside [^2][^3].

[^1]: https://github.com/tarantool/tarantool/releases/tag/3.0.0-alpha1
[^2]: https://github.com/tarantool/tarantool/pull/8771
[^3]: https://github.com/tarantool/tarantool/pull/8866

## What's the solution?

It is necessary to "save" the tarantoolctl utility in order to use it regardless of whether it is part of future Tarantool packages or not. This can be done by simply copying the file to the root of the project directory.

### Debian 11

```
$ docker run -v /tarantool:/tarantool -it debian:11 /bin/bash
root@14acb818b457:/# apt-get update && apt-get install python3-pip -y
root@14acb818b457:/# python3 --version
Python 3.9.2
root@14acb818b457:/tarantool/test-run# pip install -r requirements.txt
root@14acb818b457:/tarantool/static-build/build# dpkg -i tarantool_3.0.0~alpha1-1_amd64.deb
root@14acb818b457:/tarantool/static-build/build# tarantool --version
Tarantool 3.0.0-alpha1-0-gef6335e
...
root@14acb818b457:/tarantool/test# ./test-run.py --force --builddir ../static-build/tarantool-prefix/src/tarantool-build --executable /usr/bin/tarantool
...
Tarantool server information:
 | Found executable at /usr/bin/tarantool
 | Found tarantoolctl at /tarantool/test-run/tarantoolctl

 | Tarantool 3.0.0-alpha1-0-gef6335e
 | Target: Linux-x86_64-RelWithDebInfo
 | Build options: cmake . -DCMAKE_INSTALL_PREFIX=/tarantool/static-build/tarantool-prefix -DENABLE_BACKTRACE=TRUE
 | Compiler: GNU-8.3.1
 | C_FLAGS: -fexceptions -funwind-tables -fasynchronous-unwind-tables -static-libstdc++ -fno-common -msse2 -Wformat -Wformat-security -Werror=format-security -fstack-protector-strong -fPIC -fmacro-prefix-map=/tarantool=. -std=c11 -Wall -Wextra -Wno-gnu-alignof-expression -fno-gnu89-inline -Wno-cast-function-type
 | CXX_FLAGS: -fexceptions -funwind-tables -fasynchronous-unwind-tables -static-libstdc++ -fno-common -msse2 -Wformat -Wformat-security -Werror=format-security -fstack-protector-strong -fPIC -fmacro-prefix-map=/tarantool=. -std=c++11 -Wall -Wextra -Wno-invalid-offsetof -Wno-gnu-alignof-expression -Wno-cast-function-type

Detected build mode: Release
...
...
Statistics:
* pass: 1613
* disabled: 167
* skip: 21
```
### Fedora 36

```
$ docker run -v /tarantool:/tarantool -it fedora:36 /bin/bash
[root@e118755ec5e2 /]# yum update -y
[root@e118755ec5e2 /]# yum install python3-pip
[root@e118755ec5e2 /]# python3 --version
Python 3.10.11
[root@e118755ec5e2 build]# rpm -i tarantool-3.0.0~alpha1-1.x86_64.rpm
[root@e118755ec5e2 build]# tarantool --version
Tarantool 3.0.0-alpha1-0-gef6335e
...
[root@e118755ec5e2 test-run]# pip install -r requirements.txt
[root@e118755ec5e2 test]# ./test-run.py --force --builddir ../static-build/tarantool-prefix/src/tarantool-build --executable /usr/bin/tarantool
...
Tarantool server information:
 | Found executable at /usr/bin/tarantool
 | Found tarantoolctl at /tarantool/test-run/tarantoolctl

 | Tarantool 3.0.0-alpha1-0-gef6335e
 | Target: Linux-x86_64-RelWithDebInfo
 | Build options: cmake . -DCMAKE_INSTALL_PREFIX=/tarantool/static-build/tarantool-prefix -DENABLE_BACKTRACE=TRUE
 | Compiler: GNU-8.3.1
 | C_FLAGS: -fexceptions -funwind-tables -fasynchronous-unwind-tables -static-libstdc++ -fno-common -msse2 -Wformat -Wformat-security -Werror=format-security -fstack-protector-strong -fPIC -fmacro-prefix-map=/tarantool=. -std=c11 -Wall -Wextra -Wno-gnu-alignof-expression -fno-gnu89-inline -Wno-cast-function-type
 | CXX_FLAGS: -fexceptions -funwind-tables -fasynchronous-unwind-tables -static-libstdc++ -fno-common -msse2 -Wformat -Wformat-security -Werror=format-security -fstack-protector-strong -fPIC -fmacro-prefix-map=/tarantool=. -std=c++11 -Wall -Wextra -Wno-invalid-offsetof -Wno-gnu-alignof-expression -Wno-cast-function-type

Detected build mode: Release
...
...
Statistics:
* pass: 1612 (flaky: 1)
* disabled: 167
* skip: 21
```

Close #400

### DIFF
--- a/lib/options.py
+++ b/lib/options.py
@@ -493,6 +493,17 @@ class Options(object):
                     $ . <(./test/test-run.py --env)
                     """))
 
+        parser.add_argument(
+                '--executable',
+                dest='executable',
+                default=None,
+                help=format_help(
+                    """
+                    Set a custom path to the Tarantool executable.
+
+                    Useful when Tarantool binary is not in $BUILDDIR and $PATH.
+                    """))
+
         # XXX: We can use parser.parse_intermixed_args() on
         # Python 3.7 to understand commands like
         # ./test-run.py foo --exclude bar baz
--- a/tarantoolctl
+++ b/tarantoolctl
@@ -1,0 +1,1353 @@
+#!/usr/bin/env tarantool
+
+local io = require('io')
+local os = require('os')
+local ffi = require('ffi')
+local fio = require('fio')
+local fun = require('fun')
+local log = require('log')
+local uri = require('uri')
+local json = require('json')
+local xlog = require('xlog')
+local yaml  = require('yaml')
+local errno = require('errno')
+local fiber = require('fiber')
+local netbox = require('net.box')
+local socket = require('socket')
+local console = require('console')
+local argparse = require('internal.argparse').parse
+
+ffi.cdef[[
+int kill(int pid, int sig);
+int isatty(int fd);
+int getppid(void);
+]]
+
+local TIMEOUT_INFINITY = 100 * 365 * 86400
+
+-- path of tarantoolctl binary
+local command_path = arg[0]
+-- name of tarantoolctl binary
+local self_name = fio.basename(arg[0])
+-- command that we're executing
+local command_name = arg[1]
+-- true if we're running in user's HOME directory
+local usermode = false
+-- true if tarantoolctl is a symlink and name != tarantoolctl
+local linkmode = false
+-- a file with system-wide settings
+local default_file
+-- current instance settings
+local instance_name
+local instance_path
+local console_sock
+local group_name
+-- overrides for defaults files
+local instance_dir
+local default_cfg
+local positional_arguments
+local keyword_arguments
+local lua_arguments = arg
+local language
+
+-- function for printing usage reference
+local usage
+
+-- shift argv to remove 'tarantoolctl' from arg[0]
+local function shift_argv(arg, argno, argcount)
+    for i = argno, 128 do
+        arg[i] = arg[i + argcount]
+        if arg[i] == nil then
+            break
+        end
+    end
+end
+
+local function check_user_level()
+    local uid = os.getenv('UID')
+    if uid == 0 or os.getenv("NOTIFY_SOCKET") then
+        return nil
+    end
+    -- local dir configuration
+    local pwd = os.getenv('PWD')
+    local udir = pwd and pwd .. '/.tarantoolctl'
+    udir = udir and fio.stat(udir) and udir or nil
+    -- or home dir configuration
+    local homedir = os.getenv('HOME')
+    udir = udir or homedir and homedir .. '/.config/tarantool/tarantool'
+    udir = udir and fio.stat(udir) and udir or nil
+    -- if one of previous is not nil
+    if udir ~= nil then
+        usermode = true
+        return udir
+    end
+
+    return nil
+end
+
+--
+-- Find if we're running under a user, and this user has a default file in their
+-- home directory. If present, use it. Otherwise assume a system-wide default.
+-- If it's missing, it's OK as well.
+--
+local function find_default_file()
+    -- try to find local dir or user config
+    local user_level = check_user_level()
+    if user_level ~= nil then
+        return user_level
+    end
+
+    -- no user-level defaults, use system-wide ones
+    local cfg = '@CMAKE_INSTALL_FULL_SYSCONFDIR@/@SYSCONFIG_DEFAULT@/tarantool'
+    if fio.stat(cfg) then
+        return cfg
+    end
+    -- It's OK if there is no default file.
+    -- load_default_file() will assume some defaults
+    return nil
+end
+
+local function check_file(path)
+    local rv, err = loadfile(path)
+    if rv == nil then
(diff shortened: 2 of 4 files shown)


tarantoolctl
```diff
@@ -0,0 +1,1353 @@
+#!/usr/bin/env tarantool
```
Totktonada, line 1: I guess that only start/stop/status is needed for test-run, not all the 1353 lines of the code. It should be easy to strip unneeded parts.

Totktonada, line 1: > tarantool will be build as a static build and in the future the tarantoolctl utility will be removed

The test-run change is completely unrelated to how tarantool is linked with libraries. It would confuse an external reader, which doesn't know that tarantoolctl deprecation/removal is in the same track as change of the linking type.

It is even more related to the second commit, which calls some layout changes 'static build' not only in the commit message, but also in the option name.

lib/options.py
```diff
@@ -228,6 +228,22 @@ def __init__(self):
...
+                help=format_help(
+                    """
+                    Mark tarantool build as static:
+
+                    - tarantoolctl utility will be used from the $TEST_RUN_DIR;
+                    - tarantool binary will be search in the '/static-build` directory
+                    or you can set the custom path via the `STATIC_BINARY` env variable.
+
+                    Default: false.
+                    """))
```
Totktonada, line 245: test-run already has a lot of code to choose tarantool executable automatically or manually.

First, test-run supports a usual development flow, when tests are run from a build directory. If it is the same as a source directory (so called in-source build, `cmake . && make -j`), everything works without any manual path tuning. There is `--builddir` for an out-of-source build (`cmake /path/to/source/dir && make -j`), it is integrated into `make test` in tarantool. So, it works pretty good too.

Next, if we need to run the testing against an externally installed tarantool, test-run search it in `PATH` (if the locally built one is not found). Sometimes it is confusing in context of running of the tarantool's test suite, see #249. Sometimes, convenient. But anyway: there is a way to run the testing against arbitrary tarantool executable. And it uses pretty standard Unix approach with the `PATH` variable.

So, I'm in doubt that we need another way to pass the executable path.

And I don't undestand, to be honest, from where `/static-build` default comes. It looks as quite unusual root directory.

I suspect that test-run has all the needed executable search functionality and maybe even more than needed.

(And, again, it is not related anyhow to how tarantool links needed libraries.)

tarantoolctl
Totktonada, line 1: TBH, I don't see where test-run is changed to use this `tarantoolctl` executable.

ochaplashkin: My experience in administration tells me that when we migrate something (`state1 -> state2`), it is necessary to save states and not change something. I have created a ticket and we will be able to return to this issue a little later.

ochaplashkin: Yes, thanks! Fixed it.
Indeed, many people do not care how the build tarantool and this is not the root cause of our changes.

lib/options.py
ochaplashkin: Thanks for the comment! Indeed, my bad.

The problem was that I thought of the static assembly as a **"custom"** approach with **"unique parameters"**. It turned out that this was not the case.

We want to add an additional parameter `--executable` for the binary path (it seems to be convenient).

tarantoolctl
ylobankov: Agree with @ochaplashkin.

Totktonada: OK, no problem.

Totktonada: I see the related changes in `find_exe()` now.

ylobankov (approved):

Totktonada (approved):
Glanced briefly, I see nothing critical.

`find_exe()` definitely needs rewriting, it is hard to understand. OK to do it separately, maybe in #249.

PR #402, opened 2023-07-22
Fix incorrect character in log file name

### LINKED ISSUE #90: Create tarball with failing test artefacts to use in CI
We can archive the entire `var` directory content.

There are two options:

* create the tarball outside of the `var` directory under the option like `--artefact`;
* create the tarball inside the `var` directory always (the directory removed on each run).

i-cod:
Running jobs in tarantool-ci fails at the `action/upload-artifact` runtime if there are broken tests. The reason for the problem is that the path of the test artifacts file contains the `:` character when test has been parametrized.

Its **non-supported** character in the upload action. The action doesn't allow the following characters in artifact path: ",:,<,>,|,*,?. [^1]

To fix the problem, the `:` character has been replaced with `.`.

[^1]: https://github.com/actions/toolkit/blob/master/packages/artifact/docs/additional-information.md#non-supported-characters

Fixes https://github.com/tarantool/tarantool-qa/issues/90

### DIFF
--- a/lib/app_server.py
+++ b/lib/app_server.py
@@ -116,9 +116,9 @@ class AppServer(Server):
     def logfile(self):
         # remove suite name using basename
         test_name = os.path.basename(self.current_test.name)
-        # add :conf_name if any
+        # add .conf_name if any
         if self.current_test.conf_name is not None:
-            test_name += ':' + self.current_test.conf_name
+            test_name += '.' + self.current_test.conf_name
         # add '.tarantool.log'
         file_name = test_name + '.tarantool.log'
         # put into vardir


lib/app_server.py
```diff
@@ -118,7 +118,7 @@ def logfile(self):
         test_name = os.path.basename(self.current_test.name)
         # add :conf_name if any
```
Totktonada, line 119: This comment is obsoleted by the change.

listeners.py
```diff
@@ -70,7 +70,7 @@ def get_long_mark(self, task):
         return '(long)' if task in self.long_tasks else ''
 
     def prettify_task_name(self, task_id):
-        return task_id[0] + ((':' + task_id[1]) if task_id[1] else '')
+        return task_id[0] + (('-' + task_id[1]) if task_id[1] else '')
```
Totktonada, line 73: I don't think that we should change `test:conf` format shown on the screen.

lib/app_server.py
ochaplashkin: Thanks! I didn't notice this line at all.
I decided to just update it by adding a new character to the comment.

listeners.py
ochaplashkin: Yep, indeed, it will be easier to read this way.

Removed these changes.

lib/app_server.py
```diff
@@ -116,9 +116,9 @@ def __init__(self, _ini=None, test_suite=None):
     def logfile(self):
         # remove suite name using basename
         test_name = os.path.basename(self.current_test.name)
-        # add :conf_name if any
+        # add `-conf_name` if any
         if self.current_test.conf_name is not None:
-            test_name += ':' + self.current_test.conf_name
+            test_name += '-' + self.current_test.conf_name
```
Totktonada, line 121: BTW, can we use `.` (dot) here? It is usually not used in test names.

Totktonada (approved):
At whole I have no objections.

lib/app_server.py
ochaplashkin: @ylobankov what do you think about it?

ylobankov: I vote for `.`.

ylobankov (approved):

lib/app_server.py
ochaplashkin: Ok, fixed it :rocket:

PR #418, opened 2024-01-23
get_iproto_port: remove duplicates

0x501D:
In some tests, instances have duplicate file descriptors on one socket, which leads to abnormal termination of the test.

This patch removes duplicates to work around this problem.


ylobankov (approved):

Totktonada (approved):
No objections from my side.

PR #432, opened 2024-05-27
Adjust test result report width to terminal size

Totktonada:
Now the test configuration name is not trimmed if the terminal width allows it.

The idea can be illustrated this way:

```
+----------------------------------------------------------------------+
| Wide terminal                                                        |
+----------------------------------------------------------------------+
| suite/mytest_test.lua my_configuration                  [ pass ]     |
+----------------------------------------------------------------------+

+-------------------------------------------------+
| Narrow terminal                                 |
+-------------------------------------------------+
| suite/mytest_test.lua my_configur> [ pass ]     |
+-------------------------------------------------+
```

I want to add support for ability to run test cases as separate parallel tasks for luatest based tests with reporting of the test case name as a test configuration to the terminal. The test case names sometimes are quite long and it is convenient to see as much symbols from it as possible.

At the same time, the configuration name is stripped if the terminal has a low width. This way the output look good as on a wide as well as on a narrow terminal.

The feature works on Python 3.3+.

### DIFF
--- a/test-run.py
+++ b/test-run.py
@@ -56,6 +56,8 @@ import time
 from lib import Options
 from lib import saved_env
 from lib.colorer import color_stdout
+from lib.colorer import separator
+from lib.colorer import test_line
 from lib.utils import find_tags
 from lib.utils import shlex_quote
 from lib.error import TestRunInitError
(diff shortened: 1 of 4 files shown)


ylobankov (approved):

ochaplashkin (approved):
LGTM :rocket:

test-run.py
```diff
@@ -114,18 +116,18 @@ def main_loop_parallel():
...
-    color_stdout("\n", '=' * 86, "\n", schema='separator')
-    color_stdout("WORKR".ljust(6),     schema='t_name')
-    color_stdout("TEST".ljust(48),     schema='t_name')
-    color_stdout("PARAMS".ljust(16),   schema='test_var')
-    color_stdout("RESULT\n",           schema='test_pass')
-    color_stdout('-' * 81, "\n",       schema='separator')
+    color_stdout('\n')
+    separator('=')
+    color_stdout('WORKR ', schema='t_name')
+    test_line('TEST', 'RARAMS')
```
Gumix, line 122: RARAMS

test-run.py
```diff
@@ -167,11 +169,11 @@ def main_loop_consistent(failed_test_ids):
...
     for name, task_group in task_groups:
         # print information about current test suite
-        color_stdout("\n", '=' * 80, "\n", schema='separator')
-        color_stdout("TEST".ljust(48),     schema='t_name')
-        color_stdout("PARAMS".ljust(16),   schema='test_var')
-        color_stdout("RESULT\n",           schema='test_pass')
-        color_stdout('-' * 75, "\n",       schema='separator')
+        color_stdout('\n')
+        separator('=')
+        test_line('TEST', 'RARAMS')
```
Gumix, line 174: RARAMS

test-run.py
Totktonada: Fixes in PR #448.

test-run.py
Totktonada: Fixes in PR #448.

PR #433, opened 2024-05-27
luatest: allow to run test cases in parallel

Totktonada:
Requires https://github.com/tarantool/luatest/pull/378.

Feels better with #432.

----

In order to use the feature, add the following comment to a beginning of the test file:

```lua
-- tags: parallel
```

### DIFF
--- a/lib/luatest_server.py
+++ b/lib/luatest_server.py
@@ -3,6 +3,7 @@ import os
 import re
 import sys
 
+from subprocess import PIPE
 from subprocess import Popen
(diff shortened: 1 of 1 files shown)


ochaplashkin (approved):

Totktonada:
> Requires [tarantool/luatest#378](https://github.com/tarantool/luatest/pull/378).

The bump of the submodule is in PR #436.

> Feels better with #432.

It already lands to test-run.

PR #434, opened 2024-05-29
dispatcher: lift pipe buffer size restriction

### LINKED ISSUE #287: Run many tasks on one worker
test-run puts tasks into queues at start, but size of those queues are not infinite (AFAIR, it is restricted by size of a pipe buffer).

How to reproduce:

```sh
./test/test-run.py -j 1 $(yes box-tap/key_def.test.lua | head -n 10000)
```

Totktonada:
A task queue dispatcher puts all the tasks to the task queue at startup. Then workers are started and are taking the tasks from it.

If there are many tasks in a task group (which roughly corresponds to a test suite), we can reach the pipe buffer size on putting into the queue, because `multiprocessing.SimpleQueue` uses a pipe under the hood.

The solution is to use `multiprocessing.Queue`, which has an intermediate buffer before the underlying pipe and writes to the pipe in a background thread, without blocking a thread that calls `<queue>.put()`.

The `Queue` API is a superset of the `SimpleQueue` API, so we can just replace the implementation.

Let's also use `Queue` for the worker's output queue to be on the safe side and for consistency.

Fixes #287

### DIFF
--- a/dispatcher.py
+++ b/dispatcher.py
@@ -8,30 +8,23 @@ import yaml
 
 import multiprocessing
 
-# SimpleQueue is available from multiprocessing.queues on
-# all Python versions known at the moment of writting the code
-# (up to 3.9).
-#
-# It was additionally exposed directly from the multiprocessing
-# module since Python 3.3 ([1]).
+# Queue is available from multiprocessing.queues on all Python
+# versions known at the moment of writting the code (up to 3.12).
 #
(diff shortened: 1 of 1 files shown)


ylobankov (approved):
Nice!

ochaplashkin (approved):
LGTM :rocket:

PR #435, opened 2024-06-08
Calculate parallel jobs based on available CPUs

Totktonada:
We're going to enable new ARM64 runners in CI, where test-run is invoked in a Docker container. At the same time, the runner is an LXD container created by the service provider. In this circumstances, the Docker container sees all the 128 online CPUs, but the runner may have only some of them available (depending on the pricing plan).

Python 3.3+ has a function to determine available CPUs, so we can reduce the parallelism to this value. We fall back to the online CPUs count on Python < 3.3.

After this change, test-run follows a CPU affinity mask set by `taskset` (and, I guess, by `numactl`).

The change is similar to replacing `nproc --all` to `nproc`.

The change only affects the default behavior, which can be overwritten by passing the `--jobs` (or `-j`) CLI option or using the `TEST_RUN_JOBS` environment variable.

Reported in https://github.com/tarantool/tarantool/pull/10102

### DIFF
--- a/lib/utils.py
+++ b/lib/utils.py
@@ -8,6 +8,7 @@ import difflib
 import time
 import json
 import subprocess
+import multiprocessing
 from lib.colorer import color_stdout
 
 try:
@@ -30,6 +31,12 @@ try:
 except ImportError:
     # Python 2.7.
     get_terminal_size = None
+
+try:
+    # Python 3.3+
+    from os import sched_getaffinity
+except ImportError:
+    sched_getaffinity = None
 
 UNIX_SOCKET_LEN_LIMIT = 107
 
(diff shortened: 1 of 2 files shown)


ylobankov (approved):

lib/utils.py
```diff
@@ -384,3 +391,22 @@ def terminal_columns():
...
+
+
+def cpu_count():
+    """
+    Return available CPU count available for the current process.
+
+    The result is the same as one from the `nproc` command.
+
+    It may be smaller than all the online CPUs count. For example,
+    an LDX container may have limited available CPUs or it may be
```
ylobankov, line 403: LDX -> LXD

Totktonada: Eagle eye! Fixed.

PR #446, opened 2024-08-27
Bump luatest to 1.0.1-20-ga978383

ylobankov:
New version includes the following commits related to luatest functionality:

- test: don't fail if luatest is in system paths [1]
- Introduce preloaded hooks [2]
- Migrate treegen module and adapt it [3]
- Adapt server to use declarative config [4]

[1] tarantool/luatest@9a07bc1
[2] tarantool/luatest@13ca5a6
[3] tarantool/luatest@82e8777
[4] tarantool/luatest@1ee220c


Totktonada (approved):
I've no objections.

PR #309, opened 2021-06-10
Add default server restart to each test run

avtikhon:
Patchset of 3 patches:

1. Found that the root cause of the issues happened with vinyl tests were
  backside effects of the not correct test 'vinyl/gh.test.lua' which
  leaved Tarantool worker process in inconsistent state. After it any
  other next test on the same Tarantool worker process could fail on
  running testings with snapshots calls, like tarantool/tarantool-qa#126:
      ```
        error: Snapshot is already in progress
      ```
      Either restarting Tarantool worker process could fail on stopping it,
      like #261 and tarantool/tarantool#5141:
      
        E> failed to process vylog record: delete_slice{slice_id=115, }
        E> ER_INVALID_VYLOG_FILE: Invalid VYLOG file: Slice 115 deleted but not registered
      
      Decided to avoid of such situations for now and in the future all tests
      must be run after Tarantool worker default server process was restarted.
      This patch moves this server restart call from check if the test failed
      to the common part of the tests run loop.
      
      Part of tarantool/tarantool-qa#97
      Fixes #260
      Fixes #261
      Needed for tarantool/tarantool#5089
      Closes tarantool/tarantool-qa#126

2. After the previous commit in the current patchset we don't use the same
tarantool server process from test to test and we don't need to use any
scripts like 'pretest_clean.lua' to cleanup it between tests. So this
patch removes this script.

3. Found that all currently used suite.ini files use this option enabled.
Also there is no need to have any suite of tests where snap either
log files whould be used by the next test from the previous, due to
such test schema can be easily implemented within single test file.

PR #449, opened 2024-10-25
Add default server restart before each test in consistent mode

Astronomax:
Needed for tarantool/tarantool#6965
Closes tarantool/tarantool#6965


Totktonada (approved):
It seems a straightforward follow-up for #309. I'm OK, but, please, verify this change on tarantool's and vshard's test suites. The consistent mode is not touched for a long time and I can't be sure that I remember all the details.

PR #454, opened 2025-04-11
Remove luatest and checks submodules

### LINKED ISSUE #453: Remove luatest submodule
Having luatest as a submodule is inconvenient because:
1. To update luatest in some project using it via test-run, we have to update it in test-run first.
2. From point 1 it follows that all projects using test-run are bound to use the same version of luatest.

Let's drop the luatest submodule from test-run and require projects using test-run install luatest as a prerequisite, either as a submodule or from luarocks.

locker:
The `test-run` project is rarely updated these days while `luatest` is rapidly developing. Let's remove the `luatest` submodule from `test-run` to ease luatest bumps in projects using `test-run`. Now, `luatest` is supposed to be added as a submodule to all projects using `test-run`.

By default, `test-run` looks for `luatest` in `../third_party/luatest`, but this can be overridden with the new `--luatestdir option`.

Note that we also remove the `checks` submodule, which was added as a dependency of `luatest`. It was built in Tarantool in 2.11.0 so there's no need in it anymore.

Closes #453

### DIFF
--- a/lib/options.py
+++ b/lib/options.py
@@ -229,6 +229,21 @@ class Options(object):
                     """))
 
         parser.add_argument(
+                "--luatestdir",
+                dest="luatestdir",
+                default="../third_party/luatest",
+                help=format_help(
+                    """
+                    Path to luatest directory.
+
+                    Beware: A relative path is resolved against the directory,
+                    where all tests reside. **NOT** a current directory of a
+                    parent shell.
+
+                    Default: ../third_party/luatest
+                    """))
+
+        parser.add_argument(
                 "--builddir",
                 dest="builddir",
                 default="..",
--- a/.gitmodules
+++ b/.gitmodules
@@ -4,9 +4,3 @@
 [submodule "lib/tarantool-python"]
 	path = lib/tarantool-python
 	url = https://github.com/tarantool/tarantool-python.git
-[submodule "lib/checks"]
-	path = lib/checks
-	url = https://github.com/tarantool/checks.git
-[submodule "lib/luatest"]
-	path = lib/luatest
-	url = https://github.com/tarantool/luatest.git
--- a/.luacheckrc
+++ b/.luacheckrc
@@ -27,6 +27,4 @@ include_files = {
 exclude_files = {
     "lib/tarantool-python",
     "test/test-tarantool/*.test.lua",
-    "lib/luatest/**",
-    "lib/checks/**",
 }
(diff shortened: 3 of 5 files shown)


locker:
We can't remove the submodules because they are needed for integration tests. I think I'll leave them be and instead allow to specify a custom luatest directory.

Totktonada:
We still can do it, just add `luatest` installation into the CI script.

lib/options.py
```diff
@@ -228,6 +228,21 @@ def __init__(self):
                     Default: false.
                     """))
 
+        parser.add_argument(
+                "--luatestdir",
+                dest="luatestdir",
+                default="../third_party/luatest",
```
Totktonada, line 234: Can we search it using `LUA_PATH` by default?

I mean, we don't need the `bin/luatest` wrapper (it just makes things more complicated for us in the case), we can just do `tarantool -e 'require('luatest.cli_entrypoint')()'`.

Totktonada: I think, we don't ever need this `--luatestdir` argument or env variable. Just correctly configured lua paths.

locker: The thing is I don't want to setup `LUA_PATH` when running tests in Tarantool with `test-run.py`. I just want everything to work by default, without any extra parameters or environment variables. Moreover, I'd like `test-run.py --env` to setup `LUA_PATH` so that I can run tests with `luatest`.

locker:
> We still can do it, just add `luatest` installation into the CI script.

Yep, that'd probably be better than leaving the `luatest` submodule in `test-run` just for the sake of those tests. However, to achieve that I'd have to teach `test-run` to work with `luatest` installed from rocks, which seems tricky. I'll try that.

TBH I find these tests useless because we check that `test-run` works by updating the submodule in the Tarantool project, which is much more extensive and reliable. Maintaining these tests is a burden. Actually, they don't even work with Tarantool 3.0. Given that `test-run` is a legacy, which should eventually be replaced with pure `luatest` or `ctest`, I'd prefer to simply drop those tests. BTW these tests were added in 2021 (76cefbeb0cdb5283fb7555d576b7781bb2197c5c), when `test-run` in fact stopped developing.

lib/options.py
locker: In other words, yes, we could require developers to setup `luatest` by themselves for running Tarantool tests, but (a) this would be inconvenient and (b) could result in non-obvious errors caused by using an old version of `luatest`. Better checkout `luatest` as a submodule in Tarantool and make `test-run` use it automatically IMO.

locker: Done in https://github.com/tarantool/test-run/pull/455.

locker:
Superseded by:
- #455.

PR #455, opened 2025-05-27
Remove luatest submodule (v2)

### LINKED ISSUE #453: Remove luatest submodule

locker:
The test-run project is rarely updated these days while luatest is rapidly developing. Let's remove the luatest submodule from test-run to ease luatest bumps in projects using test-run. Now, `luatest` is supposed to be installed from rocks.

Closes #453

The first version of this patch used a different approach. It required the superproject to pull luatest as a submodule[^1]. We've decided to reject that approach.

[^1]: https://github.com/tarantool/test-run/pull/454

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,3 +1,5 @@
+export PATH := .rocks/bin:$(PATH)
+
 MAKEFILE_PATH := $(abspath $(lastword $(MAKEFILE_LIST)))
 PROJECT_DIR := $(patsubst %/,%,$(dir $(MAKEFILE_PATH)))
 TEST_RUN_EXTRA_PARAMS?=
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -8,13 +8,13 @@ jobs:
         github.event.pull_request.head.repo.full_name != github.repository ) &&
         ( github.repository == 'tarantool/test-run' )
 
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-latest
 
     strategy:
       fail-fast: false
       matrix:
-        python-version: ['3.7', '3.8', '3.9', '3.10', '3.11', '3.12']
-        tarantool-version: ['2.8', '2.10', '2.11']
+        python-version: ['3.8', '3.9', '3.10', '3.11', '3.12']
+        tarantool-version: ['2.10', '2.11']
 
     steps:
     - uses: actions/checkout@v3
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -31,11 +31,15 @@ jobs:
         python-version: ${{ matrix.python-version }}
     - name: display python version
       run: python -c "import sys; print(sys.version)"
+    - name: setup tt
+      run: |
+        curl -L https://tarantool.io/release/2/installer.sh | sudo bash
+        sudo apt install -y tt
+        tt version
     - name: setup dependencies
       run: |
-       sudo apt update -y
-       sudo apt-get -y install lua5.1 luarocks
-       sudo luarocks install luacheck
+        tt rocks install luatest
+        tt rocks install luacheck
     - name: setup python dependencies
       run: |
        pip install -r requirements.txt
--- a/.gitmodules
+++ b/.gitmodules
@@ -4,9 +4,3 @@
 [submodule "lib/tarantool-python"]
 	path = lib/tarantool-python
 	url = https://github.com/tarantool/tarantool-python.git
-[submodule "lib/checks"]
-	path = lib/checks
-	url = https://github.com/tarantool/checks.git
-[submodule "lib/luatest"]
-	path = lib/luatest
-	url = https://github.com/tarantool/luatest.git
--- a/.luacheckrc
+++ b/.luacheckrc
@@ -27,6 +27,5 @@ include_files = {
 exclude_files = {
     "lib/tarantool-python",
     "test/test-tarantool/*.test.lua",
-    "lib/luatest/**",
-    "lib/checks/**",
+    ".rocks/**/*.lua",
 }
--- a/bin/luatest
+++ b/bin/luatest
@@ -1,0 +1,20 @@
+#!/usr/bin/env tarantool
+
+--
+-- Add the luatest module to LUA_PATH so that it can be used in processes
+-- spawned by tests.
+--
+local fio = require('fio')
+local path = package.search('luatest')
+if path == nil then
+    error('luatest not found')
+end
+path = fio.dirname(path) -- strip init.lua
+path = fio.dirname(path) -- strip luatest
+os.setenv('LUA_PATH',
+          path .. '/?.lua;' .. path .. '/?/init.lua;' ..
+          (os.getenv('LUA_PATH') or ';'))
+
+print(('Tarantool version is %s'):format(require('tarantool').version))
+
+require('luatest.cli_entrypoint')()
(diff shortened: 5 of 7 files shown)


locker:
There's a problem with this update: it breaks auto-requiring of the luatest module from a Tarantool test server. One way to fix that issue is using the original search root, like this:

```diff
diff --git a/luatest/server.lua b/luatest/server.lua
index 96b1e40e114a..c303952403cc 100644
--- a/luatest/server.lua
+++ b/luatest/server.lua
@@ -347,6 +347,7 @@ function Server:build_env()
         TARANTOOL_HTTP_PORT = self.http_port,
         TARANTOOL_LISTEN = self.net_box_port or self.net_box_uri,
         TARANTOOL_ALIAS = self.alias,
+        TARANTOOL_SEARCHROOT = package.searchroot()
     }
     if self.box_cfg ~= nil then
         res.TARANTOOL_BOX_CFG = json.encode(self.box_cfg)
@@ -834,7 +835,10 @@ function Server:exec(fn, args, options)
         for i = 1, debug.getinfo(fn, 'u').nups do
             local name, _ = debug.getupvalue(fn, i)
             if passthrough_ups[name] then
+                local old_root = package.searchroot()
+                package.setsearchroot(os.getenv('TARANTOOL_SEARCHROOT'))
                 debug.setupvalue(fn, i, require(passthrough_ups[name]))
+                package.setsearchroot(old_root)
             end
         end
         local result = {xpcall(function()
```

Is there a better way?

Totktonada:
> Is there a better way?

I would prefer to pass `LUA_PATH` into the child environment instead of flipping `searchroot` back and forth. We can split `package.path`, make all the paths absolute, and pass it to the child's environment.

In fact, it solves the problem that is more general than just `luatest`'s autorequire. Even two problems:

* The child process has its own current working directory that means that all the *default* *relative* search paths that successfully works with `.` and `.rocks` in the parent process doesn't work in the child process.
* All the non-standard `package.path` additions from the test that are working in the parent process do not passed to the child process.

It seems quite convenient to pass all these paths to the child process from the parent one.

Makefile
```diff
@@ -12,7 +12,7 @@ flake8:
 	$(PYTHON) -m flake8 *.py lib/*.py
 
 luacheck:
-	luacheck --config .luacheckrc .
+	.rocks/bin/luacheck --config .luacheckrc .
```
Totktonada: Can we add `.rocks/bin` into the `PATH` instead, so if `luacheck` is installed into the system, we just call the system's `luacheck`?

I prefer this setup for some of the Lua modules (to let the package manager track files for me), including `luacheck`.

Totktonada (approved):
Thank you!

I've no objections here. It is interesting to try the proposed development flow in practice.

Makefile
locker: Good point! Done.

<details><summary>diff</summary>

```diff
diff --git a/Makefile b/Makefile
index c353305f90b5..fbc424a3910c 100644
--- a/Makefile
+++ b/Makefile
@@ -1,3 +1,5 @@
+export PATH := .rocks/bin:$(PATH)
+
 MAKEFILE_PATH := $(abspath $(lastword $(MAKEFILE_LIST)))
 PROJECT_DIR := $(patsubst %/,%,$(dir $(MAKEFILE_PATH)))
 TEST_RUN_EXTRA_PARAMS?=
@@ -12,7 +14,7 @@ flake8:
 	$(PYTHON) -m flake8 *.py lib/*.py
 
 luacheck:
-	.rocks/bin/luacheck --config .luacheckrc .
+	luacheck --config .luacheckrc .
 
 test_integration:
 	PYTHONPATH=$(PROJECT_DIR) $(PYTHON) test/test-run.py --force --exclude unittest $(TEST_RUN_EXTRA_PARAMS)
```

</details>

locker:
The update with the luatest fix from https://github.com/tarantool/test-run/pull/455#issuecomment-2913190728 still results in 18 Tarantool tests failing. 5 of them are caused by the recent logging update, see https://github.com/tarantool/luatest/commit/2f16ac767f75901d233b4cca6d9fe2ead9757609. Other tests fail because they explicitly require the luatest module in scripts. Here's the list:

```
test name								failure reason
---------
box-luatest/box_cfg_env_test.lua					require('luatest')
metrics-luatest/tarantool_spaces_test.lua				require('luatest') in test utils
box-luatest/downgrade_test.lua						require('luatest') in helper
box-luatest/gh_5809_bitset_index_len_crash_test.lua			require('luatest'); may be removed
box-luatest/gh_9309_errors_in_triggers_test.lua				grep_log
replication-luatest/gh_10088_apply_deletion_of_replica_from_cluster_on_deleted_replica_test.lua		log_file
box-luatest/gh_10032_update_upsert_splice_test.lua			require('luatest') in helper
box-luatest/gh_7904_export_box_schema_version_to_public_api_test.lua	log_file
app-luatest/gh_7288_console_flavours_vs_txn_test.lua			require('luatest') in helper
app-luatest/trigger_module_test.lua					grep_log
box-luatest/box_schema_before_box_cfg_test.lua				require('luatest')
app-luatest/require_first_test.lua					require('luatest')
box-luatest/gh_7231_memtx_hash_idx_gt_iter_deprecation_test.lua		log_file
replication-luatest/gh_9723_sync_system_spaces_test.lua			require('luatest')
box-luatest/gh_8192_feedback_daemon_metrics_test.lua			require('luatest'); may be removed
config-luatest/vars_test.lua						require('luatest')
config-luatest/config_test.lua						require('luatest')
config-luatest/box_cfg_env_test.lua					require('luatest') in helper
```

locker:
> Is there a better way?

@Totktonada suggested an approach that doesn't require patching luatest:

```diff
diff --git a/bin/luatest b/bin/luatest
index 8bfb9b5626cd..4afd068e5236 100755
--- a/bin/luatest
+++ b/bin/luatest
@@ -1,5 +1,17 @@
 #!/usr/bin/env tarantool
 
+--
+-- Add the luatest module to LUA_PATH so that it can be used in processes
+-- spawned by tests.
+--
+local fio = require('fio')
+local path = package.search('luatest')
+path = fio.dirname(path) -- strip init.lua
+path = fio.dirname(path) -- strip luatest
+os.setenv('LUA_PATH',
+          path .. '/?.lua;' .. path .. '/?/init.lua;' ..
+          (os.getenv('LUA_PATH') or ';'))
+
 print(('Tarantool version is %s'):format(require('tarantool').version))
 
 require('luatest.cli_entrypoint')()
```

It adds the path to the luatest module to `LUA_PATH` from `bin/luatest` (which has to be created by test-run anyways because finding a binary installed from rocks is complicated). @Totktonada PTAL if that's what you meant.

**Update:** Also added an explicit error in case the luatest module isn't found:

```diff
diff --git a/bin/luatest b/bin/luatest
index 4afd068e5236..e51b98b75178 100755
--- a/bin/luatest
+++ b/bin/luatest
@@ -6,6 +6,9 @@
 --
 local fio = require('fio')
 local path = package.search('luatest')
+if path == nil then
+    error('luatest not found')
+end
 path = fio.dirname(path) -- strip init.lua
 path = fio.dirname(path) -- strip luatest
 os.setenv('LUA_PATH',
```

PR #456, opened 2025-07-07
Add repeat option to run untill tests fail

LevKats:
Add the `--repeat` command line option so each test should succeed `--repeat` times to pass. One may think of the option as a `--retries` analog, but instead of assuring that each test passes at least once, we require each test to pass every time. Combining both non-trivial `--retires` and `--repeat` options values require each test to pass at least once in `--retires` runs `--repeat` times in a row.

Needed for tarantool/tarantool#6646

### DIFF
--- a/lib/options.py
+++ b/lib/options.py
@@ -320,6 +320,22 @@ class Options(object):
                     """))
 
         parser.add_argument(
+                "--repeat",
+                dest='repeat',
+                default=env_int('TEST_RUN_REPEAT', 1),
+                type=int,
+                help=format_help(
+                    """
+                    The number of times each test will be repeated.
+
+                    Combining both non-trivial --retries and --repeat options
+                    values require each test to pass at least once in
+                    --retries runs --repeat times in a row.
+
+                    Default: ${TEST_RUN_REPEAT} or 1.
+                    """))
+
+        parser.add_argument(
                 "-p", "--pattern",
                 dest='pattern',
                 nargs='+',
--- a/lib/worker.py
+++ b/lib/worker.py
@@ -96,14 +96,15 @@ def get_task_groups():
     """
     suites = find_suites()
     res = collections.OrderedDict()
+    repeat = Options().args.repeat
     for suite in suites:
         key = os.path.basename(suite.suite_path)
         gen_worker = functools.partial(Worker, suite)  # get _id as an arg
         # Split stable and fragile tasks to two groups. Run stable
         # tasks in parallel with other ones. Run fragile tasks one
         # by one when all parallel tasks will be finished.
-        stable_task_ids = [task.id for task in suite.stable_tests()]
-        fragile_task_ids = [task.id for task in suite.fragile_tests()]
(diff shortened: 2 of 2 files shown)


lib/worker.py
```diff
@@ -96,14 +96,15 @@ def get_task_groups():
...
     for suite in suites:
         key = os.path.basename(suite.suite_path)
         gen_worker = functools.partial(Worker, suite)  # get _id as an arg
         # Split stable and fragile tasks to two groups. Run stable
         # tasks in parallel with other ones. Run fragile tasks one
         # by one when all parallel tasks will be finished.
-        stable_task_ids = [task.id for task in suite.stable_tests()]
-        fragile_task_ids = [task.id for task in suite.fragile_tests()]
+        stable_task_ids = [task.id for task in suite.stable_tests()] * repeat
+        fragile_task_ids = [task.id for task in suite.fragile_tests()] * repeat
```
locker, line 107: I don't think it's the right place for the 'repeat' option, because AFAIU this function is supposed to find all available test groups, not generate the test execution plan. It looks like the tests names should be repeated somewhere in `find_tests`.

lib/options.py
```diff
@@ -319,6 +319,18 @@ def __init__(self):
...
                     """))
 
+        parser.add_argument(
+                "--repeat",
+                dest='repeat',
+                default=env_int('TEST_RUN_REPEAT', 1),
+                type=int,
+                help=format_help(
+                    """
+                    The number of test run retries needed for success.
```
locker: I'd rather say "Repeat each test the specified number of times". I'd also say here in the comment how this option interacts with 'retries'.

lib/worker.py
Totktonada: This variant makes sense too: this way `--repeat` relates to the task scheduling mechanism (which is quite generic) rather than test searching logic.

My initial proposal about `--retries` also targets the task scheduling machinery (https://github.com/tarantool/test-run/pull/217#issuecomment-691829969), but it it much more complicated, because we need results listening and dynamic scheduling of new tasks. Here we can just generate more tasks at start.

I guess I have no strong preference, where to handle it: in `find_tests`, here or somewhere in the dispatcher logic.

If we compare code, then `find_tests` should handle it in each test suite type. The dispatcher needs some logic to spawn several workers for 'one' task. And here we just generate several tasks from one. That looks like the simplest possible change.

Totktonada: Am I remember right: we can't just tune the `--retries` logic to handle `--repeat` as well, because the former works inside one task (that means inside one particular worker) and doesn't allow to run a test many times *in parallel*?

Totktonada (approved):

lib/worker.py
LevKats: Yes, this is the case. We want different "repetitions" to run in parallel.

lib/options.py
LevKats: Fixed.

locker (approved):

PR #458, opened 2026-04-06
Remove LUA_PATH provided to luatest

ImeevMA:
There is no need to specify `LUA_PATH` anymore, since starting with luatest 1.4.3 the Tarantool `package.searchroot` mechanism is supported.

### DIFF
--- a/bin/luatest
+++ b/bin/luatest
@@ -1,20 +1,5 @@
 #!/usr/bin/env tarantool
 
---
--- Add the luatest module to LUA_PATH so that it can be used in processes
--- spawned by tests.
---
-local fio = require('fio')
-local path = package.search('luatest')
-if path == nil then
-    error('luatest not found')
-end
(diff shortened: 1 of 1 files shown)


bin/luatest
```diff
@@ -1,20 +1,5 @@
...
--- spawned by tests.
---
-local fio = require('fio')
-local path = package.search('luatest')
-if path == nil then
-    error('luatest not found')
-end
-path = fio.dirname(path) -- strip init.lua
-path = fio.dirname(path) -- strip luatest
-os.setenv('LUA_PATH',
```
locker, line 14: We need to make sure that all Tarantool tests pass with this change. Please prepare a PR bumping test-run in the Tarantool repository.

ImeevMA: Done: https://github.com/tarantool/tarantool/pull/12606

locker (approved):

PR #460, opened 2026-05-21
test-run: show luatest stderr on startup failure

Godrik0:
If luatest fails to start (e.g., due to a Lua error in one of the loaded modules), test-run only shows the exit code without the actual error message. Capture stderr from the subprocess and include it in the error message so the cause of the failure is visible right away.

This may result in verbose error output in some cases. Let me know if the original behavior of discarding stderr was intentional - I found it very inconvenient to debug without seeing the actual error message.


Totktonada (approved):

ImeevMA (approved):
Thanks for the patch! It seems very useful!

